PR #26, opened 2026-02-14
Add baseline ESlint config

lpukatch:
- ESLint Integration: Added eslint.config.mjs and updated dependencies to establish consistent linting standards.


MotWakorb:
I'm gonna hold this PR up a bit - I'm modularizing the backend, so i want to be careful about merging it; you'll probably have some rework to do, unfortunately.

lpukatch:
I just re-scoped this to only add eslint and the baseline config. If that still seems like too much I'll hold off for now

MotWakorb:
Okay, check 0.13.0 in dev - latest dev.  Make sure you're compliant against that and I'll take a look at merging into dev.

PR #39, opened 2026-02-24
  feat: make HTTP and HTTPS ports fully configurable via environment variables

lpukatch:
This merge request introduces full configurability for the application's network ports, removing hardcoded defaults and providing better support for varied deployment environments.

ECM_PORT (default: 6100) for the primary HTTP interface.
ECM_HTTPS_PORT (default: 6143) for HTTPS traffic when TLS is enabled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,13 +62,15 @@ RUN mkdir -p /config /config/tls /config/uploads/logos \
 
 # Environment
 ENV CONFIG_DIR=/config
+ENV ECM_PORT=6100
+ENV ECM_HTTPS_PORT=6143
 
-# Expose ports (HTTP on 6100, HTTPS on 6143 when TLS enabled)
-EXPOSE 6100 6143
+# Expose ports (HTTP on ECM_PORT, HTTPS on ECM_HTTPS_PORT when TLS enabled)
+EXPOSE $ECM_PORT $ECM_HTTPS_PORT
 
 # Add healthcheck
 HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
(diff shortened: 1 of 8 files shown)


MotWakorb:
Thanks for the contribution! Here are some issues I noticed during review:

**1. `EXPOSE $ECM_PORT` is build-time only (Critical)**
Dockerfile `EXPOSE` is evaluated at build time, not runtime. Since `ENV ECM_PORT=6100` is set before `EXPOSE`, it will always bake in `6100` regardless of what the user sets at `docker run` time. The comment suggests it's dynamic but it isn't. EXPOSE is documentation-only so it doesn't break anything functionally, but it's misleading. Consider keeping `EXPOSE 6100 6143` with a comment noting the defaults, or removing EXPOSE entirely.

**2. Redundant `ECM_PORT=${ECM_PORT:-6100}` in entrypoint.sh**
The fallback is repeated 4 times across `check_network`, `check_tls_config`, `print_startup_info`, and the final exec block. Set it once at the top of the script instead.

**3. TLS settings.py default reads env at import time**
```python
https_port: int = int(os.environ.get("ECM_HTTPS_PORT", 6143))
```
This changes the Pydantic model default based on the env var at module import time. If a user has already saved a TLS config with an explicit `https_port` in their JSON, the saved value wins (correct). But if they haven't configured TLS yet, the env var silently controls the default. The interaction between env var default vs saved config vs UI-editable field should be documented so users understand the precedence.

**4. Missing: direct invocation path**
The uvicorn port is only parameterized in `entrypoint.sh`. If someone runs the app directly (`uvicorn main:app`) without the entrypoint script, they still get hardcoded ports. Consider reading `ECM_PORT` in `main.py` as well so both paths are consistent.

**5. Commit message typo**
"enviornmnet variables" → "environment variables"

PR #40, opened 2026-02-24
feat: make HTTP and HTTPS ports fully configurable via environment variables

### LINKED ISSUE #38: [Feature]: Allow for changing the port within the container via Environment variable

I run ECM through gluetun and I want to run dev alongside. Adding this would allow me to change it in this situation where port mapping happens through gluetun rather than the docker compose

_No response_

### Additional Context

_No response_

lpukatch:
This merge request introduces full configurability for the application's network ports, removing hardcoded defaults and providing better support for varied deployment environments.

ECM_PORT (default: 6100) for the primary HTTP interface.
ECM_HTTPS_PORT (default: 6143) for HTTPS traffic when TLS is enabled.

closes #38

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -1,5 +1,4 @@
 from pydantic import BaseModel
-from pydantic_settings import BaseSettings
 import json
 import os
 import logging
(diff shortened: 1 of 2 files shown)


MotWakorb:
## Code Review

Overall this is a clean, focused PR — the entrypoint, docker-compose, and TLS settings changes all look correct.

One issue with the Dockerfile:

```dockerfile
EXPOSE $ECM_PORT $ECM_HTTPS_PORT
```

`EXPOSE` is evaluated at **build time**, not runtime. Even though `ENV ECM_PORT=6100` is set above, if a user overrides `ECM_PORT` at runtime (e.g., `docker run -e ECM_PORT=8080`), the `EXPOSE` metadata will still say 6100. Since the intent is to make the ports configurable inside the container, `EXPOSE` should stay hardcoded to the defaults (`EXPOSE 6100 6143`) — it's just documentation/metadata anyway, and the actual port binding in `entrypoint.sh` correctly reads the env var at runtime. Alternatively, you could use `docker build --build-arg` with `ARG` instead of `ENV` for `EXPOSE`, but that adds complexity for no real benefit.

Please revert `EXPOSE` to the hardcoded defaults so it doesn't give the false impression that it responds to runtime config.

PR #45, opened 2026-02-25
EPG title and EPG description search added to Auto-Channel creation feature.

titooo7:
Please check carefully the code quality before approving the PR. I'm a geek not a software developer (I did it with gemini-cli)

Worst case scenario it might serve as an idea to get it implemented differently or in a more clean way :)

Example of Import rules I used: https://pastebin.com/raw/kPwJNE4c


titooo7:
I'll try to fix it and send a separate PR

Issue #4, opened 2026-01-02
Feature request


lavazza2017:
1. inside I have channells like ✦●✦ EN| ENGLAND CINEMA ✦●✦ which are to trash ... can I regex somehow for deletion automaticly ?
2. VIP| E ENTERTAINMENT RAW -> I channels name like this so prefix not catch "VIP| " same with PLAY+| CANAL SPORT ᴿᴬᵂ -> not catch "PLAY+|"
3. mby it's to much but mby you will have some ideas: I have a lot of miss match of channels like "STARS TV" and "STARS.TV" or "SUNDANCE" and "SUNDANCE TV" or "SPORT KLUB" and "SPORTKLUB"  -> any AI/logic to marge them too ??
4. similar to 2, I have "HBO 3 HD (NA)" , "HBO HD (NA)" which oryginal are "EN| HBO 3 HD (NA)"
5. if on left side I have 2 smilar channel I didn't find marge option

MotWakorb:
Use v0.12.0 dev for handling most of this.

Issue #3, opened 2026-01-02
when adding multiple or single group/s gives error


lavazza2017:
[message.txt](https://github.com/user-attachments/files/24410663/message.txt)
everything in topic

MotWakorb:
This should be resolved in the latest version.

Issue #2, opened 2026-01-02
Suffix not being recognized


Deekerman:
it doesnt recognize L1 as a suffix so it wont strip it ( UK | iFollow League 1's suffix is L1, league 2 is L2)

On V3.0

MotWakorb:
Custom prefix matching was added in 0.5.0.

Issue #1, opened 2026-01-02
Copy Paste doesnt work


Deekerman:
on V3.0 

i have both streams and channels loaded 
the copy paste icon only populates on the channels side
and clicking it doesnt work.

PR #48, opened 2026-03-01
Another attempt at adding EPG Search & Timezone features to Auto-Creation

titooo7:
Done with gemini-cli, please review carefully.

### DIFF
--- a/.claude/commands/logs.md
+++ b/.claude/commands/logs.md
@@ -1,0 +1,150 @@
+# Log Analysis
+
+Analyze ECM logs to diagnose issues, trace requests, and surface anomalies. Supports both live container logs and log files on disk.
+
+## Usage
+`logs [focus] [path]`
+
+- `focus` — optional filter: an error message, module tag, time range, or keyword like "slow", "errors", "startup"
+- `path` — optional path to a log file or directory containing log files (e.g., `~/ecm`)
+
+## Steps
+
+### 1. Determine Log Source
+
+**If a path is provided**, use log files from that location:
+
+```bash
+# If path is a directory, find log files in it
+find <path> -maxdepth 1 -name "*.log" -o -name "*.txt" | head -20
+ls -lhS <path>/*.log <path>/*.txt 2>/dev/null
+
+# For large files (>10MB), NEVER read the whole file. Use tail/grep:
+tail -500 <path>/output.log                          # Last 500 lines
+grep -c ' - ERROR - \| - CRITICAL - ' <path>/output.log  # Count errors
+grep ' - ERROR - \| - CRITICAL - ' <path>/output.log | tail -50  # Recent errors
+
+# For time-bounded analysis on large files:
+grep '^2026-03-09' <path>/output.log | tail -500     # Specific date
+sed -n '/^2026-03-09 14:00/,/^2026-03-09 15:00/p' <path>/output.log | tail -500  # Time range
+
+# For keyword/tag filtering on large files:
+grep '\[STREAM-PROBE\]' <path>/output.log | tail -100
+grep -B5 -A5 'Traceback' <path>/output.log | tail -200
+```
+
+**If no path is provided**, fetch from the live container:
+
+```bash
+# Default: last 500 lines
+docker compose logs --no-log-prefix --tail=500 ecm 2>&1
+
+# Time-bounded (if user specifies a range)
+docker compose logs --no-log-prefix --since=1h ecm 2>&1
+
+# Follow mode is NOT supported in this context — always use tail/since
+```
+
+### Known Log Files
+
+| Location | Description | Size |
+|-|-|-|
+| `~/ecm/output.log` | Full captured backend log (~200MB, multi-day) | Large — always use tail/grep |
+| `~/ecm/*.txt` | Captured log snippets (e.g., stream prober runs) | Small — can read directly |
+| Container stdout | Live logs from running ECM instance | Use `docker compose logs` |
+
+### 2. Parse Log Format
+
+ECM backend logs follow this format:
+```
+TIMESTAMP - MODULE_NAME - LEVEL - [TAG] Message
+```
+
+Key fields:
+| Field | Example |
+|-|-|
+| Timestamp | `2026-03-09 14:22:01,123` |
+| Module | `main`, `routers.channels`, `config` |
+| Level | `DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL` |
+| Tag | `[MAIN]`, `[CHANNELS]`, `[M3U]`, `[REQUEST]`, `[SLOW-REQUEST]`, `[RAPID-POLLING]`, `[CONFIG]`, `[VALIDATION-ERROR]` |
+
+Uvicorn also emits its own access/lifecycle logs (different format — no tag prefix).
+
+### 3. Triage by Focus
+
+**If focus is "errors" or unspecified and errors exist:**
+- Filter for `ERROR` and `CRITICAL` level lines
+- Group by module/tag
+- Show full tracebacks (look for `Traceback` blocks and `exc_info` output)
+- Check if errors are recurring or one-off
+
+**If focus is "slow":**
+- Find `[SLOW-REQUEST]` entries (requests >1000ms)
+- Extract endpoint, method, and duration
+- Look for patterns (same endpoint? same time window?)
+
+**If focus is "startup":**
+- Find the most recent `[MAIN] Starting ECM` or uvicorn startup message
+- Trace the initialization sequence: config load, DB connection, service starts
+- Flag any warnings or errors during startup
+
+**If focus is "rapid-polling" or "polling":**
+- Find `[RAPID-POLLING]` warnings (20+ hits in 10s)
+- Identify which endpoints and source IPs are flooding
+- Check if this correlates with slow requests or errors
+
+**If focus is a specific module tag (e.g., "m3u", "channels", "epg"):**
+- Filter logs to lines containing `[TAG]` (case-insensitive match)
+- Show chronological flow for that subsystem
+- Highlight any errors or warnings
+
+**If focus is a specific error message or keyword:**
+- Grep for the keyword in the captured logs
+- Show surrounding context (5 lines before/after)
+- Trace the request flow that produced it
+
+### 4. Analyze Patterns
+
+Look for:
+- **Error clusters**: Multiple errors in a short time window
+- **Cascading failures**: One error triggering others (e.g., DB timeout → multiple endpoint failures)
+- **Recurring warnings**: Same warning repeating — may indicate a configuration issue
+- **Memory/resource hints**: Messages about connection pools, file descriptors, or timeouts
+- **Request timing trends**: Are slow requests concentrated on certain endpoints or time periods?
+
+### 5. Report
+
+Present findings as a structured summary:
+
+**Log window**: Time range of analyzed logs
+**Total lines**: Count of log lines analyzed
+**Breakdown by level**: Count of DEBUG / INFO / WARNING / ERROR / CRITICAL
+
+Then for each finding:
+- **What**: The issue or pattern observed
+- **Where**: Module, endpoint, or tag
+- **When**: Timestamp(s) or frequency
+- **Impact**: What this likely means for the running system
+- **Suggestion**: Next step to investigate or fix
+
+### 6. Check Current Log Level
+
+If the analysis seems incomplete (e.g., no DEBUG output when needed):
+
+```bash
+docker exec ecm-ecm-1 cat /config/settings.json | grep -i log_level
+```
+
+Suggest changing the level if more detail is needed:
+- More detail: set `backend_log_level` to `DEBUG` in settings
+- Less noise: set to `WARNING` if logs are overwhelming
+
+## Notes
+
+- Container logs go to stdout/stderr; Docker manages retention (no rotation configured)
+- `~/ecm/output.log` is a large persistent capture — always use tail/grep, never read it whole
+- Log injection is protected (newlines sanitized via `log_utils.py`)
+- Frontend logs are browser-console only — not visible in container or file logs
+- The `journal.db` SQLite database stores an audit trail separate from system logs
+- `httpx` lines (HTTP Request logs) appear in log files from the stream prober's outbound requests
+- `--- Logging error ---` lines indicate a Python logging formatter crash — these are bugs worth flagging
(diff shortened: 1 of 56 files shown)


MotWakorb (commented):
# PR #48 Review: EPG Search & Timezone Features for Auto-Creation

**Reviewed:** +1,638 / -200 across 27 files, 12 commits
**Author note:** "Done with gemini-cli, please review carefully."

---

## Task 1: Product Manager Review

### Business Value
This PR adds meaningful functionality that users in IPTV management will value: searching EPG program guides to auto-create channels based on what's airing today, converting timezone data for international stream names, and group-aware isolation to prevent cross-group channel conflicts. These are real pain points for users managing international channel lineups.

### User Experience Concerns

**Good:**
- Fallback template concept is smart -- users can have different naming for EPG-matched vs name-matched streams
- EPG info displayed in execution logs helps users debug why rules matched
- Transform Time action with searchable timezone dropdowns is well-implemented

**Issues:**
1. **Feature discoverability is poor.** Nine new condition types flood the condition selector with no visual grouping or help text. A user seeing "EPG Title (Today)" won't know they need to also add "EPG Source" to avoid false positives. The USER_GUIDE mentions this but the UI doesn't guide users toward it.
2. **"Fallback Name Template (optional)" label is confusing.** Users won't understand "fallback" without context. The hint text ("used if match was found via Name Search instead of EPG") helps, but only after they've already scrolled to it.
3. **No loading indicator** for timezone list fetch -- if the API is slow, the dropdowns show nothing.

### Strategic Alignment
Aligns well with the auto-creation feature expansion. However, the `docker-compose.yml` port changes (HTTPS port 6143) are unrelated to the EPG feature and should be in a separate PR.

---

## Task 2: Developer Review

### Critical Issues

1. **Duplicate `matching_rules` initialization** (`auto_creation_engine.py` ~line 905): The PR adds a second `matching_rules = []` declaration immediately below the original one. While functionally harmless (the second overwrites the first), this indicates sloppy copy-paste and reduces readability.

2. **Duplicate import** (`auto_creation_executor.py`): The PR adds `from datetime import datetime` at the module level AND again right before `_parse_epg_date`. The second import is redundant.

3. **O(n) linear scan for log entry update** (`auto_creation_engine.py` ~line 1273):
   ```python
   for log_entry in results["execution_log"]:
       if log_entry.get("stream_id") == stream.stream_id:
           log_entry["actions_executed"] = actions_log
           break
   ```
   This runs for every matched stream, creating O(n*m) complexity. Should use a dict index.

4. **~140 lines of inline EPG processing** in `run_pipeline()` (lines 133-271): This massive block should be extracted to a private method like `_prefetch_epg_grid()`. It makes `run_pipeline()` harder to read and test.

5. **Side-effectful condition evaluation**: `evaluate()` mutates `StreamContext.epg_match` and `StreamContext.matched_by_epg` as a side effect. If multiple conditions in an AND group evaluate against EPG, the last one's match overwrites earlier ones. This is fragile -- the evaluation order determines which program metadata ends up in the template variables.

6. **Cross-condition peeking** in `evaluate()`: The `source_filter` logic scans all sibling conditions to find an `EPG_SOURCE_IS` condition. This couples condition evaluation to rule-level structure. If a user has an EPG source condition in a different OR group, it still gets picked up, which may be surprising.

### Code Style Issues

- Em-dash character replacements (`—` to `\u2014`) in docstrings throughout test files -- these are cosmetic noise that shouldn't be in this PR
- Several `except Exception` catches that are too broad (e.g., `compute_event_times` timezone fallback)
- `package-lock.json` has many `"peer": true` removals -- these seem unrelated and could indicate a different npm version was used to regenerate the lockfile

### Performance

- **EPG grid fetch**: Fetches up to 5000 EPG data entries even if only one stream has a TVG ID. Consider lazy loading per-stream or caching at the engine level.
- **Timezone endpoint**: Called on every `ActionEditor` mount (even for non-timezone actions). Should be fetched lazily only when `transform_time` is selected.

---

## Task 3: Quality Engineer Review

### What's Tested (15 new tests)
- EPG condition evaluation: title contains, desc matches, any matches, source is, source filter (5 tests)
- Multi-field conditions: name-only, EPG title precedence (2 tests)
- Transform time: success, no match, invalid tz (3 tests)
- Fallback template: no-EPG fallback, EPG primary (2 tests)
- Engine: EPG grid prefetch, execution log efficiency (2 tests)
- Executor: assign_profile_none (1 test)

### Critical Missing Coverage

| Feature | Gap |
|-|-|
| `_parse_epg_date` (ISO & XMLTV formats) | Zero tests |
| `_find_channel_by_name(name, group_id)` | No group_id filtering test |
| `_find_channel_by_regex(pattern, group_id)` | No group-scoped test |
| `_find_channel_by_tvg_id(tvg_id, group_id)` | No group-scoped test |
| `_filter_channels_by_group` helper | Zero tests |
| `from_dispatcharr_stream` with `account_groups` | Zero tests for new group resolution |
| EPG grid dict-format response (`programmes`/`data`) | Only list format tested |
| Negation of EPG conditions | Zero tests |
| `epg_match` overwrite across multiple EPG conditions | Zero tests |
| Frontend: transform_time UI | Zero tests |
| Frontend: fallback template UI | Zero tests |
| Frontend: ConditionEditor EPG source selector | Zero tests |

### Regression Risks
1. **Channel lookup refactoring**: All `_channel_by_name`, `_base_name_to_channel`, `_normalized_name_to_channel`, `_core_name_to_channel`, `_callsign_to_channel` changed from `dict[str, dict]` to `dict[str, list[dict]]`. Every caller now gets lists instead of single dicts. One missed call site = crash. The existing test `test_init_with_channels` was updated (`[0]` added) but group-aware paths are untested.
2. **`matching_rules` duplication**: While harmless, it suggests mechanical editing without understanding.
3. **Short-circuit evaluation change**: The PR adds `break` on first matching OR group. The original code evaluated ALL groups for complete logging. This changes behavior -- some condition results won't appear in logs anymore.

---

## Task 4: Security Engineer Review

| # | Finding | Severity |
|-|-|-|
| 1 | ReDoS risk: user-provided regex patterns execute on potentially long EPG descriptions via `re.search()` without timeout | Medium |
| 2 | `transform_time` timezone/source_field not validated at save time -- only fails at execution | Medium |
| 3 | `zoneinfo.ZoneInfo()` is safe (caught exceptions, no traversal) | Low |
| 4 | Template expansion is non-recursive `str.replace()` -- no injection risk | Low |
| 5 | EPG data in execution log is public broadcast metadata, not PII | Info |
| 6 | HTTPS port exposure (6143) is intentional, user-configurable | Low |

**ReDoS detail**: The `_evaluate_epg_field` method runs user-provided regex against all EPG programs for the day. With potentially hundreds of programs and descriptions up to ~1000 chars, a catastrophic backtracking pattern could freeze the pipeline. Consider `re.TIMEOUT` (Python 3.11+) or enforcing a maximum pattern length.

**Validation gap**: `transform_time.source_tz` and `transform_time.target_tz` are only validated at execution time. A rule with "Invalid/Timezone" will save successfully and only error when the pipeline runs.

---

## Task 5: DevOps Review

### docker-compose.yml
- Adds `ECM_HTTPS_PORT` (default 6143) -- **this is unrelated to EPG features** and should be in a separate PR
- Port mapping uses env vars with defaults: `${ECM_PORT:-6100}:${ECM_PORT:-6100}` -- correct pattern
- Both HTTP and HTTPS ports are exposed and passed as environment variables

### CI/CD
- `pytest.ini` adds `asyncio` marker -- needed for new async tests, no issue
- `package-lock.json` regeneration removes many `"peer": true` flags -- this suggests the contributor used a different npm version which could cause reproducibility issues on CI

### pytz Removal
- The PR replaces `pytz` with stdlib `zoneinfo` throughout `dummy_epg_engine.py` and `auto_creation_executor.py`
- Uses `try: import zoneinfo except ImportError: from backports import zoneinfo` pattern -- correct for Python <3.9 compat
- `pytz.timezone().localize()` replaced with `datetime(..., tzinfo=tz)` -- correct modern pattern
- **However**: The `pytz` import is removed from `dummy_epg_engine.py` but `pytz` may still be required as a dependency elsewhere. Need to verify no other files import it.

---

## Task 6: UI/UX Designer Review

### Visual Consistency
- New CSS classes (`.log-epg-info`, `.log-epg-title`, `.log-epg-desc`) follow existing patterns with proper text overflow handling
- Uses existing design tokens (`--accent-primary`, `--text-muted`)
- Material icons (`event`) for EPG indicators are consistent with the existing icon set

### Usability Issues

1. **Condition selector overflow**: Adding 9 new condition types to the "Stream Metadata" category makes it the largest category by far. Should consider a sub-category like "EPG" within the selector.

2. **Fallback template placement**: Located between the primary name template and the group selector. This breaks the visual flow -- users expect group selection right after naming. Consider placing it immediately below the primary template with a smaller visual footprint (collapsible or inline toggle).

3. **Transform Time action UX**: The "From Timezone" / "To Timezone" side-by-side layout is good. However, "Extraction Regex" is intimidating for non-technical users. A preset dropdown (e.g., "HH:MM", "HH:MM AM/PM") with a custom option would be more accessible.

4. **EPG info in execution log**: Good addition. The truncation at 60 chars for descriptions is reasonable. The `title` attribute on hover shows full text -- good.

### Accessibility
- Fallback template input has `htmlFor` and `id` properly linked
- Variable inserter button has `aria-label="Show fallback variables"` -- good
- Transform Time fields all have proper labels
- Missing: no `aria-live` region for when timezone data finishes loading

---

## Summary Verdict

This PR introduces valuable EPG-based auto-creation features but needs work before merge:

**Must Fix:**
1. Extract ~140-line EPG processing block into a dedicated method
2. Remove duplicate `matching_rules = []` and duplicate `from datetime import datetime`
3. Add tests for group-aware channel lookup functions (`_find_channel_by_name` with `group_id`, `_filter_channels_by_group`, etc.)
4. Add tests for `_parse_epg_date`
5. Fix O(n^2) execution log update -- use a dict index
6. Validate `transform_time.source_tz`/`target_tz` at schema validation time (not just execution)

**Should Fix:**
7. Separate docker-compose.yml HTTPS port changes into their own PR
8. Address `package-lock.json` regeneration (use same npm version as project)
9. Lazy-load timezone list only when transform_time action is selected
10. Add frontend tests for new UI elements (fallback template, transform_time, EPG source selector)

**Nice to Have:**
11. Sub-categorize EPG conditions in the condition selector
12. Add ReDoS protection (pattern length limit or `re.TIMEOUT`)
13. Collapsible fallback template field to reduce visual noise

MotWakorb:
## PR #48 Review: EPG Search & Timezone Features for Auto-Creation

**Author:** titooo7 | **Size:** +2,750 / -1,268 across 28 files, 16 commits | **Tools used:** gemini-cli, then GLM 4.7

### What it adds
- EPG program guide searching for auto-creation rules (match by title, description, etc.)
- Timezone conversion action (`transform_time`) for international stream names
- Group-aware channel isolation to prevent cross-group conflicts
- Fallback name template when EPG match isn't found
- EPG info in execution logs

### Have they improved?

**Somewhat, but core issues persist.** After the first review (requesting changes), several fix commits were pushed. The second review found:

**Still critical (3 issues):**

1. **`get_epg_grid` double-parsing** — The client returns a list but the engine checks for a dict, so `grid_channels` is always empty, triggering a fallback `get_epg_data(page_size=5000)` call on *every* pipeline run. GLM 4.7's comment acknowledges this is "necessary" but that's because the fix was wrong — the client method strips the dict before the engine can use it.

2. **`epg_match` state bleeds across rules** — `StreamContext.epg_match` and `matched_by_epg` are never reset between rules, so rule 2 inherits rule 1's EPG match data, corrupting fallback template logic.

3. **Execution log regression** — Only matched streams are logged now. Users debugging why streams *didn't* match lose all visibility.

**Still major:**
- `account_groups` lookup uses wrong key (`m3u_account_id` vs `channel_group_id`) — group name resolution is completely broken
- Short-circuit OR drops condition logs
- O(N²) `source_filter` scanning all conditions per evaluation

### Commit quality
The commits tell the story — "some more fiexes" (twice, with typo), multiple "fixing some issues" messages. The GLM 4.7 fix commit tried to address the review feedback but introduced a new regression (#1 above) and wrote a justification for the workaround rather than fixing the root cause.

### Verdict
The Gemini-generated code adds real feature value, but the AI-assisted fix rounds are creating a whack-a-mole pattern — fixing one issue while introducing another. The 3 critical issues from the second review still need addressing before this is mergeable.

MotWakorb (commented):
## PR #48 Review Update (Post-Fix Commits)

Issues #2 (state bleed), #3 (execution log), and #4 (account_groups lookup) are **fixed** — nice work on those.

Removing `EPG_SOURCE_IS` is fine — it was new to this PR and never shipped, so no breaking change. Can always be re-added properly in a follow-up.

### Still needs fixing

**1. Negated EPG conditions corrupt match state (Major)**

The evaluation sets `context.epg_match` and `context.matched_by_epg = True` *before* applying negation. So a negated EPG condition like "EPG Title does NOT contain X" that initially matched will still mark the stream as EPG-matched even though the condition evaluates to `False`. This corrupts the fallback template logic — the stream gets the EPG name template instead of the fallback template.

Fix: Only set `context.epg_match`/`matched_by_epg` when the *final* result (after negation) is `True`.

**2. Empty AND/OR groups match everything (Moderate)**

An empty AND group returns `True` ("Empty AND matches"), and an empty OR group does the same. A rule with no conditions in a group would match every stream. Should return `False` for empty groups, or at minimum empty OR groups.

**3. Condition log structure changed (Moderate)**

Condition logs changed from structured per-condition objects (with `type`, `value`, `matched`, `connector` fields) to flat `or_group` blobs with a string like `"[0] stream_name_contains: ... AND [1] epg_title_contains: ..."`. The frontend parses these as structured objects for the execution log UI — this will break that display or lose debugging granularity.

**4. Short-circuit OR logging (Minor, leftover from previous review)**

OR groups now all evaluate (good), but the per-condition detail within each group is flattened into a single string instead of individual log entries. Less important than #3 but related.

### Summary

Three of the original six issues are solidly fixed. The EPG_SOURCE_IS removal is acceptable. But #1 (negation bug) is a real correctness issue that will produce wrong channel names, and #2-3 need attention before merge.

titooo7:
Hopefully the code is now good enough to merge the feature

titooo7:
Another attempt, let's see if this is the final one, lol. 

I appreciate that the different LLM's I tried to use (gemini pro and glm) aren't as good as I wish without a knowledgable guidance that I can't provide.  Therefore please feel free to let me know if I should stop trying for now.

Obviously it doesn't cost me anything to ask gemini-cli to try again after receiving your change request comments, but I also don't want to waste your time :)

PR #77, opened 2026-04-20
Fix stream sorting by provider m3u priority value

stevencoutts:
When trying to sort by provider m3u priority value, streams were not sorted. This appears to fix it.

### DIFF
--- a/backend/auto_creation_engine.py
+++ b/backend/auto_creation_engine.py
@@ -702,13 +702,16 @@ class AutoCreationEngine:
 
                 channel_name = channel.get("name", f"Channel #{channel_id}")
 
-                # Build smart sort key function
-                sorted_streams = _smart_sort_streams(
+                # Respect rule.stream_sort_field (Provider Order, Quality, etc.).
+                # Previously this always used global smart-sort settings, so "Provider Order (M3U)"
+                # only changed the log label and did not reorder by M3U account priority.
+                sorted_streams = _reorder_streams_for_rule(
                     current_streams,
+                    rule,
                     self._stream_stats_cache,
                     stream_m3u_map,
                     channel_name,
-                    settings
+                    settings,
                 )
 
                 # Skip if order didn't change
@@ -723,7 +726,7 @@ class AutoCreationEngine:
                         "rule_id": rule.id,
                         "rule_name": rule.name,
                         "action": f"Would reorder {len(sorted_streams)} streams in '{channel_name}' "
-                                  f"by smart sort ({rule.stream_sort_field})",
+                                  f"by {_stream_sort_rule_label(rule.stream_sort_field)}",
                         "would_create": False,
                         "would_modify": True
                     })
@@ -745,7 +748,8 @@ class AutoCreationEngine:
                                 elif stats.get("probe_status") in ("failed", "timeout"):
                                     deprioritized.append({"id": sid, "name": stats.get("stream_name", f"Stream {sid}"), "reason": stats.get("probe_status")})
 
-                        desc_parts = [f"Reordered {len(sorted_streams)} streams in '{channel_name}' by smart sort ({rule.stream_sort_field})"]
+                        mode_label = _stream_sort_rule_label(rule.stream_sort_field)
+                        desc_parts = [f"Reordered {len(sorted_streams)} streams in '{channel_name}' by {mode_label}"]
                         if deprioritized:
                             reasons = {}
                             for d in deprioritized:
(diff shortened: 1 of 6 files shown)


MotWakorb (commented):
Welcome, Steven — and thank you for the first contribution to ECM. This is a solid PR with a real fix, a clear commit history, tests that pass, and a thoughtful UI hint for the user. Details below; none of the comments are blockers but a couple are worth a follow-up.

## Summary of what changed

Three independent fixes rolled into one PR, each with its own commit — nice atomic structure:

1. **`acd0c8bd` — `backend/stream_prober.py` + `backend/routers/stream_stats.py`:** Dispatcharr has returned either `id` or `stream_id` as the stream identifier across endpoint/version permutations. Building `stream_m3u_map` with `s["id"]` silently produced an empty map when the payload used `stream_id`, so M3U-priority sorting degraded to insertion order. Defensive `s.get("id", s.get("stream_id"))` fixes it.
2. **`733a1e2b` — `backend/routers/stream_stats.py`:** When the compute-sort endpoint is called with `mode == "m3u_priority"`, force `deprioritize_failed_streams = False`. Makes sense — a pure M3U-priority sort should not bury a stream just because probe stats are missing; M3U priority is a property of the account, not the probe.
3. **`e9035a24` — `backend/auto_creation_engine.py` + `RuleBuilder.tsx`:** The big one. Pass 3.5 was hardcoded to call `_smart_sort_streams()` regardless of `rule.stream_sort_field`, so picking "Provider Order (M3U)" in the Rule Builder only relabeled the execution log without actually changing the ordering. New `_reorder_streams_for_rule()` dispatches by field to four new helpers (`_sort_streams_by_m3u_account_priority`, `_sort_streams_by_resolution_height`, `_sort_streams_by_stream_name` with optional natural sort), falling back to smart sort on `smart_sort`/empty/unknown. That's exactly the right shape.

## Verification

I ran the backend test suite against your branch staged into the `ecm-ecm-1` container, comparing failure sets against the dev baseline at `127220a7`:

- **Dev baseline (`origin/dev`):** 2114 passed, 43 failed (all pre-existing environmental/seed failures in `test_password_security`, `test_dispatcharr_provider`, etc. — unrelated to this PR)
- **PR #77:** 2117 passed, 43 failed — identical failure set, **+3 new passes from the tests you added**
- **`tests/unit/test_sort_criteria.py` + `tests/unit/test_auto_creation_engine.py`:** 86/86 passing

**No regressions.** The three new tests in `TestStreamReorderByRule` are clean, readable, and hit the behavior that was broken before. Thank you for including them — that's exactly the bar we hold for sort changes.

## Overlap with PR #74 (merged earlier today)

PR #74 fixed a different sort bug in the same general area: within failed-rank buckets (black_screen, low_fps, failed), primary criteria weren't being applied because the composite sort key tail was `(0,) * len(active_criteria)`. That lived in `backend/stream_prober.py::smart_sort_streams`.

Your PR is **complementary**, not conflicting:
- Your `stream_prober.py` diff is in `_auto_reorder_channels` (lines ~1840), solely the `id`/`stream_id` tolerance. No overlap with PR #74's `compute_criteria_values` helper at the top of the file.
- PR #74's 4 TDD regression tests (`TestWithinBucketPrimaryCriteria` + the updated `test_standalone_failed_still_sorted_by_m3u`) all pass on your branch.
- Your PR is rebased on dev HEAD after PR #74 merged, so `compute_criteria_values` is already in the stream_prober.py you shipped. No merge conflict, no semantic conflict.

In short: PR #74 fixed **within-bucket tiebreaking for smart sort in the scheduled probe path**; this PR fixes **cross-mode dispatch in the auto-creation Pass 3.5 path**. Both land cleanly.

## Review comments

### Non-blocking — worth considering

**1. Cross-fork `dev` → `dev` targeting.** Branch name on your fork is `dev` and you're targeting our `dev`. That works, but it makes local reviewer workflows awkward — `gh pr checkout 77` refuses to fetch into a branch of the same name currently checked out. For future contributions, consider a topic branch like `fix/m3u-priority-sort` off your `dev` — CONTRIBUTING.md recommends this pattern too. Not a blocker here.

**2. `"No sorting"` dropdown option (value `""`) still calls smart sort.** In `_reorder_streams_for_rule`, the first branch is `if not field or field == "smart_sort"`, which sends the empty string down the smart-sort path. That matches the pre-existing behavior (before your PR, *all* non-smart-sort values also fell through to smart sort), so it's not a regression you introduced — but it means the UI label "No sorting" still reorders streams, which could surprise users. Out of scope for this PR, but worth filing a follow-up issue.

**3. Latent bug in `_smart_sort_streams` itself (bd-bqpq0).** Heads-up: the `_smart_sort_streams` function this dispatcher still calls (when `field == "smart_sort"`) has the same within-bucket tuple-of-zeros bug that PR #74 fixed in `stream_prober.py` — tracked as bead `bqpq0` in our backlog. Your PR doesn't need to touch this (it's genuinely separate scope), but users who pick "Smart Sort (default)" on an auto-create rule will still hit that bug until bqpq0 lands. If you wanted to take that one on as a follow-up, the pattern is: extract a `compute_criteria_values()` helper mirroring the one PR #74 added at `stream_prober.py:170`, then call it in the three deprioritized branches of `_smart_sort_streams` (around `backend/auto_creation_engine.py:2177-2187`). Totally optional.

### Nits (feel free to ignore)

**4. `_natural_sort_key` forward-reference.** `_sort_streams_by_stream_name` calls `_natural_sort_key`, which is defined *after* the dispatcher in the file. Python resolves that at call time so it works — and the test coverage confirms it — but for readability, moving `_natural_sort_key` above the sort helpers would be cleaner. Tiny.

**5. Log tag consistency.** Your new helpers log under `[AUTO-CREATE-ENGINE]`, which matches the rest of the file — good. One tiny inconsistency: the provider-order helper logs `"provider order (%s) by M3U priority -> %s"` with a bare arrow, while the other helpers use `"(%s) -> %s"`. Minor; not worth a re-spin.

**6. `_sort_streams_by_stream_name` could use `sorted(..., reverse=order=="desc")` instead of `list(reversed(temp))` after an ascending sort.** Cleaner and one less traversal. Also minor.

## Verdict

**Approving.** The three commits are each a real bug fix, the test coverage is appropriate for the scope, there are no regressions against `origin/dev`, and the code is idiomatic. Thanks for the clear commit messages and for matching our existing logging conventions — this was an easy PR to review.

The comments above are either follow-up territory (bqpq0) or pure polish; nothing needs to change before merge. If you pick up a follow-up PR, we'd love it — and ping us on the issue tracker if you hit any workflow friction. Welcome aboard.

— Claude (reviewer on behalf of the ECM maintainers)

PR #93, opened 2026-04-21
Auto-creation merge_streams: add prune toggle

stevencoutts:
Add a Merge Streams toggle to remove streams that no longer match. When enabled, merge_streams records desired stream IDs per channel and a post-pass prunes the channel’s stream list to keep it in sync.

Made-with: Cursor


MotWakorb (changes requested):
Thanks for the contribution! The feature itself is a welcome addition. A few things to address before this can merge.

## 🚨 Blocker — `.beads/` contamination

The reported +1311/-1524 is almost entirely `.beads/issues.jsonl` (~2700 diff lines), `.beads/metadata.json`, and `.beads/config.yaml`. The config change also adds:

```yaml
sync.remote: "git+ssh://git@github.com/stevencoutts/enhancedchannelmanager.git"
```

That would wholesale overwrite our issue-tracking database and point the project's beads sync remote at your personal fork. Please drop all three `.beads/` files from this PR (and the `sync.remote` line) — they shouldn't ship with feature work.

## Feature review (~240 lines across 6 files)

The actual change is small and well-shaped:

- `backend/auto_creation_schema.py` — `remove_non_matching: bool` validation with default `False` ✅
- `backend/auto_creation_executor.py` — `_merge_prune_desired_by_channel` accumulator + new `prune_merge_streams()` (dry-run aware, logs on failure) ✅
- `backend/auto_creation_engine.py` — "Pass 2.75" placed correctly between merge and re-sort ✅
- `frontend/src/types/autoCreation.ts` + `ActionEditor.tsx` — type + checkbox with hint ✅

## Concerns

**1. Weak test assertion.** `test_merge_remove_non_matching_prunes_channel` only asserts `update_channel.call_count >= 2`. It never verifies the pruned payload — stream `999` (the stale one) could still be in the final list and the test would pass. Please assert on the `streams` list actually passed to the prune call (e.g. inspect `self.client.update_channel.call_args_list[-1]` and verify `999` is gone and the new merged stream is present).

**2. No dry-run test.** The `results["dry_run_results"]` branch of `prune_merge_streams` isn't exercised.

**3. Mixed-rule semantics (please clarify).** Only rules with `remove_non_matching=True` contribute to the desired set. If two merge rules target the same channel — one with prune off, one with prune on — streams added by the prune-off rule will be pruned, because they aren't in `desired`. That's likely surprising to users. Two options:

- Document the constraint in the UI hint ("applies across all merge rules for this channel"), or
- Accumulate all merges targeting a pruned channel regardless of flag, so only streams from this run as a whole survive.

**4. Post-prune append fallback.** In `prune_merge_streams`, the loop that appends desired IDs not in `pruned` assumes merge reliably attached streams to the channel before prune runs. If a merge failed silently upstream, we could send stream IDs to `update_channel` that aren't actually bound there yet. The inline comment notes "should be rare", but worth either gating on confirmation or noting the risk more loudly.

## Summary

- **Must:** Drop the `.beads/` files + `sync.remote` line.
- **Should:** Strengthen the new test to assert the pruned payload; add a dry-run test.
- **Nice:** Clarify mixed-rule semantics in code comment + UI hint.

Thanks again — once the beads files are out and the test is tightened, this looks close to merge-ready.

MotWakorb (approved):
Re-reviewed after the latest push (head `1b1e1907`). All blocker and concern items from the previous round are addressed.

## Verified fixes

**✅ Blocker — `.beads/` contamination removed** (commit `30c8ed34`). PR is now 6 files, +202/-0.

**✅ Test assertion strengthened** — `test_merge_remove_non_matching_prunes_channel` now inspects the actual payload sent to `update_channel`:
```python
assert payload["streams"] == [101, 201]
assert 999 not in payload["streams"]
```

**✅ Dry-run test added** — `test_merge_remove_non_matching_dry_run_logs_only` confirms `update_channel.assert_not_called()` and that `dry_run_results` gets the "Would remove" entry.

**✅ Mixed-rule semantics fixed structurally** — the accumulator split into:
- `_merge_streams_added_by_channel` — tracks **all** merge_streams activity per channel regardless of flag
- `_merge_prune_enabled_channels` — tracks which channels have at least one prune-enabled rule

So a prune rule on a channel now protects everything merged that run, not only streams added by prune-flagged rules. The UI hint also reads correctly: "it will keep only the streams that were merged into that channel during this run". Nice fix.

**✅ Post-prune append fallback removed** — the dangerous "append desired-but-missing to the update payload" loop is gone. Replaced with a `missing = desired - set(current_streams)` warning log. No risk of pushing unbound stream IDs to `update_channel`.

## Minor observation (non-blocking)

In `_execute_merge_streams`, the line:
```python
self._merge_streams_added_by_channel.setdefault(channel["id"], set()).add(stream_ctx.stream_id)
```
runs when `channel` is found, which includes the `max_streams` skip path and the "already present" path. For the already-present case this is the desired behavior (idempotent re-runs). For the `max_streams` skip, the stream gets tracked as "desired" but was never added; the new `missing` warning will surface this cleanly, so no harm. Consider renaming `_merge_streams_added_by_channel` to something like `_merge_streams_desired_by_channel` in a follow-up — the current name slightly overpromises. Not worth blocking on.

## Status

CI green (Backend Tests + Frontend Tests). MERGEABLE / CLEAN. Approving.

PR #94, opened 2026-04-21
fix(auto-creation): copy rule settings when duplicating

stevencoutts:
Duplicate rule now preserves stream/channel sort, normalization groups, orphan cleanup, and option toggles so copied rules match originals.

Made-with: Cursor

### DIFF
--- a/frontend/src/hooks/useAutoCreationRules.test.ts
+++ b/frontend/src/hooks/useAutoCreationRules.test.ts
@@ -464,7 +464,7 @@ describe('useAutoCreationRules', () => {
       expect(duplicate!.name).toContain('Original Rule');
       expect(duplicate!.name).toContain('Copy');
       expect(duplicate!.id).not.toBe(original.id);
-      expect(duplicate!.enabled).toBe(true);
+      expect(duplicate!.enabled).toBe(false);
       expect(duplicate!.run_on_refresh).toBe(true);
       expect(duplicate!.stop_on_first_match).toBe(true);
       expect(duplicate!.sort_field).toBe('quality');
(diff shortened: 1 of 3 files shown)


MotWakorb (changes requested):
Thanks for the fix! The 9 new fields (sort config, normalization groups, orphan action, etc.) are exactly the gap — cross-checked against the `AutoCreationRule` type and it's complete. CI is green. One blocking concern and two small nits.

## Blocking — the `enabled` line change

```diff
-      enabled: false, // Disabled by default
+      enabled: originalRule.enabled,
```

This flips a deliberate safety default. Previously, duplicating an enabled rule produced a disabled copy, so the user could tweak conditions/targets before activating. With the rest of your PR's changes (conditions, actions, and `target_group_id` now all preserved), the duplicate matches the same streams as the original immediately — risking duplicate channel creation, double processing, and user confusion ("why is my copy already running?").

Also worth flagging: the backend's own `POST /rules/{rule_id}/duplicate` endpoint (`backend/routers/auto_creation.py:414-461`) also hardcodes `enabled=False`. Pre-PR the two duplicate paths (frontend hook `createRule` vs. backend endpoint) were aligned on safe-by-default; post-PR they'd diverge. Keeping them aligned is worth more than the 1-click "re-enable after duplicate" ergonomic win.

Could you revert just this one line back to `enabled: false` and keep the rest of the PR as-is? If there's a specific user request driving the flip, happy to discuss — but since it's not called out in the PR description, my read is it's a side effect of the broader "preserve settings" pass rather than an intentional UX change.

## Nit 1 — test coverage gap for nullable fields

The new assertions all cover the populated case. Worth one more assertion block with `sort_field: null`, `sort_regex: null`, `stream_sort_field: null`, `normalization_group_ids: []` on the original, to round-trip the nullable path. Your current test wouldn't catch a bug where `?? null` silently became `?? 'quality'`.

## Nit 2 — mock default alignment

In \`frontend/src/test/mocks/handlers.ts\` around line 311, the mock default is \`orphan_action ?? 'none'\` — the real backend defaults to \`\"delete\"\` (see \`backend/models.py:1453\` and \`to_dict\` at \`backend/routers/auto_creation.py:703\`). Using \`'none'\` as the mock default means tests exercise a value the real API never returns by default. Low-stakes, but aligning reduces future surprise.

Ready to re-review as soon as these land. Thanks again — good catch on the settings gap overall.

MotWakorb (approved):
Thanks for the quick turnaround, Steven — all three items look good:

1. `enabled: false` restored at `frontend/src/hooks/useAutoCreationRules.ts:179`, with a nice clarifying comment on the safety rationale. This now matches the backend duplicate endpoint's behavior.
2. New test `round-trips nullable/empty sort config fields` covers `sort_field`, `sort_regex`, `stream_sort_field` as `null` and `normalization_group_ids: []` — and asserts `enabled: false` in that path too.
3. Mock handler default aligned to backend at `frontend/src/test/mocks/handlers.ts:152` (`orphan_action ?? 'delete'`).

Both CI suites green. Nice contribution — the fuller copy of sort/normalization/orphan settings is a real UX win. Appreciate the clean revision. PO will handle the merge.

PR #99, opened 2026-04-21
Feature/bulk edit auto creation rules

stevencoutts:
**Summary**
This PR introduces bulk editing for Auto-Creation Rules so users can select multiple rules and apply changes across them in one action. It adds a dedicated bulk edit modal with per-section “Apply …” toggles, a new backend bulk-update endpoint, and end-to-end typing/support in the frontend API layer.

**What’s included**

- UI

> - Multi-select rules in the Auto-Creation table
> - “Bulk edit” action opens a modal that shows representative values
> - Per-section “Apply …” toggles so users can choose exactly which groups of settings to update

- Backend

> New POST /api/auto-creation/rules/bulk-update endpoint to update multiple rules at once
> - Shared helper logic for applying updates safely/consistently
> - Router test coverage for the bulk-update path

- Frontend plumbing

> - New/updated types and API client method
> - Hook to call bulk-update and refresh UI state after apply

**Why**
Editing Auto-Creation rules one-by-one is slow when making the same adjustments across many rules. Bulk edit reduces repetitive work while preserving control by letting users opt-in to specific setting groups per operation.

**Test plan**

- Select multiple Auto-Creation rules
- Open Bulk edit
- Toggle one or more “Apply …” sections, change values, and apply
- Verify only the selected rules change, and only the opted-in sections update
- Run the existing backend test suite covering the bulk-update route


MotWakorb:
Thanks for this @stevencoutts — clean refactor extracting `_apply_rule_scalar_updates` for reuse, good a11y on the select-all indeterminate, and the per-section Apply toggles are the right UX pattern. We ran it through a multi-persona review (security, architect, engineer, code reviewer, QA, UX, SRE, DB, TW, PM) — findings below.

## Must-fix before merge (4)

1. **`probe_on_sort` cross-section collision.** `probe_on_sort` is a single column on `AutoCreationRule` (`models.py:1430`), but `BulkRuleSettingsModal.tsx` L707-709 and L714-716 both write it — from the channel-sort section when `sortField === 'quality'` AND the stream-sort section when `streamSortField === 'quality'`. If a user Applies both sections, the second write silently clobbers the first. Two ways out:
   - Add a second column (e.g., `stream_probe_on_sort`) server-side and split the UI bindings; **or**
   - Enforce mutual exclusivity in the modal and surface that to the user.

   Either is fine; flag which you pick in a reply and we can unblock fast.

2. **Add `max_items` on `rule_ids`.** `BulkUpdateAutoCreationRulesRequest.rule_ids: List[int]` has no length cap — a caller posting 100k ids forces 100k SELECTs before rollback. One line in the Pydantic field:
   ```python
   rule_ids: list[int] = Field(..., min_length=1, max_length=500)
   ```

3. **CHANGELOG entry.** All recent feature PRs (#107, #94, #98, #101) land a `CHANGELOG.md` `[Unreleased] / Added` entry in the same PR. Suggested text:
   > Auto-Creation rules — bulk edit. Multi-select rules and apply per-section changes (options, normalization groups, channel/stream sort, orphan cleanup, Merge Streams prune) in one action. Adds `POST /api/auto-creation/rules/bulk-update`; omitted fields are left unchanged. (PR #99)

4. **`docs/api.md` row.** Add to the Auto-Creation endpoint table (L311-334), matching the existing style:
   > `POST /api/auto-creation/rules/bulk-update` | Apply the same field changes to multiple rules (omitted fields unchanged)

## Non-blocking follow-ups (we'll track as beads; no need to address here)

- Mixed-value indicator when selected rules have different values for a field (prevents silent overwrite surprise)
- Apply-button count affordance + optional confirm summary (destructive-action UX)
- Replace per-id `session.query(...).first()` loop with single `filter(id.in_(rule_ids))` fetch (N+1 → 1)
- Add tests for: 404-mid-batch rollback, duplicate-id rejection, `validate_rule` failure rollback
- Frontend tests for `BulkRuleSettingsModal` (seed-from-sample effect, select-all indeterminate, patch assembly)
- `getNormalizationRules()` AbortController + surface fetch failures (currently silent `.catch()`)
- Journal `entity_id=0, entity_name="bulk"` convention — consider a dedicated bulk action_type so audit queries aren't surprised

## Ship notes

- CI is green. The `BEHIND` status is stale PR metadata — a rebase on `dev` after the must-fixes will refresh it.
- No conflicts with the in-flight bd-eio04 normalization work (wave 2 touches adjacent but different code paths in `auto_creation_*.py`).

Happy to pair on #1 if helpful. Thanks again for the contribution.

stevencoutts:
Thanks for the thorough review — all 4 must-fix items are now addressed on stevencoutts:feature/bulk-edit-auto-creation-rules (commit 62e4e1b).

- probe_on_sort collision: I chose option (2) (single column + enforce consistency). The modal now sets probe_on_sort once; when both Channel Sort + Stream Sort apply quality probing, the two checkboxes are kept in sync, and we block submit with a clear message if they somehow diverge.
- rule_ids cap: Added Field(..., min_length=1, max_length=500) to BulkUpdateAutoCreationRulesRequest.rule_ids.
- CHANGELOG: Added the requested [Unreleased] / Added entry for bulk edit.
- docs/api.md: Added the POST /api/auto-creation/rules/bulk-update row.

Note: empty rule_ids now correctly fails request validation with 422; I updated the existing router test accordingly.

MotWakorb:
Thanks @stevencoutts — all 4 must-fixes are in and clean. Engineer review passes. One additional ask from code review before merge:

## One more round (tests)

The prior must-fix list prevented silent overwrite via the UI guard + sync + validate logic, which is defense-in-depth — good. But the code-review pass caught that **no test locks the fix in**. This is the same "regression-lock" gap the original team review surfaced on the bulk-update path: the code is right today, but without a test, the next refactor can silently re-introduce the bug.

Please add three small tests (all one-liners + small fixtures):

1. **`probe_on_sort` mismatch regression test** (the critical one). Frontend RTL preferred since the validation lives there. Shape:
   - Render the bulk modal with two rules selected, both `applyChannelSort`+`applyStreamSort` checked with `quality`, but `probeOnSort=true` while `streamProbeOnSort=false`.
   - Click Apply. Assert the mismatch error message renders and no PATCH request is sent.
   - Alternative: a backend test asserting the bulk endpoint accepts exactly the value the UI sent for `probe_on_sort` and doesn't silently derive it from two sources.

2. **`rule_ids` `max_length=500` rejection**. Mirror `test_rejects_empty_rule_ids` — POST with 501 ids, assert 422.

3. **Partial-failure rollback**. Bulk-update with 3 valid ids + 1 missing id; assert no rows were mutated after the 404. This is the test the rollback logic most needs.

## Non-blocking follow-up beads

Tracking these post-merge (not needed in this PR):

- Mixed-value indicator: seed effect pulls `sample.probe_on_sort` into both `probeOnSort` and `streamProbeOnSort` state vars; if selected rules disagree, Apply silently overwrites divergent values. Applies to every sample-seeded field. Original UX review caught this too — ship an indeterminate/mixed-value UX in a follow-up.
- `journal.log_entry(entity_id=0, entity_name="bulk")` — prefer a distinct `action_type="bulk_update"` so audit filters can find bulk entries without string-matching the description.
- Replace `rulesRef.current = rules` in render body with a `useEffect` write.
- Upgrade `.query(...).get(id)` → `session.get(AutoCreationRule, id)` for SQLAlchemy 2.x.

## Ship notes

- CI is green.
- No wave-2 bd-eio04 overlap; clean merge.

Happy to merge as soon as the 3 tests land. Thanks again.

stevencoutts:
Added regression-lock coverage per the follow-up request (commit 08bd696 on feature/bulk-edit-auto-creation-rules):

- Frontend RTL: BulkRuleSettingsModal blocks submit and shows an error when channel/stream Quality probing values mismatch, and confirms onApply isn’t called.
- Backend: rejects rule_ids > 500 with 422.
- Backend: bulk-update rolls back when any rule id is missing (404 mid-batch leaves all rows unchanged).

MotWakorb:
Thanks @stevencoutts — team ran another pass and the 3 regression tests are in and the underlying fixes are all implementation-correct. CI is green. Three small test tweaks before we merge, then we're good.

## Must-fix before merge (3 small test items)

1. **Add `test_rejects_duplicate_rule_ids`.** The handler has an explicit duplicate-id guard (`if len(rule_ids) != len(set(rule_ids)): raise 400`) that's currently unverified. Without coverage a refactor could silently re-introduce double-mutation + double-counting on the same row. Shape:
   ```python
   def test_rejects_duplicate_rule_ids(test_client, test_session):
       r = make_rule(test_session)
       resp = test_client.post(
           "/api/auto-creation/rules/bulk-update",
           json={"rule_ids": [r.id, r.id], "enabled": False},
       )
       assert resp.status_code == 400
       assert "duplicate" in resp.json()["detail"].lower()
   ```

2. **Move `missing_id` to the end of the rollback-test id list.** Currently the missing id is 3rd-of-4, so `r3` is never mutated before the 404 fires — the rollback path only covers `r1`/`r2`. Putting the missing id last exercises the full dirty-all-then-rollback path:
   ```python
   # was: [r1.id, r2.id, missing_id, r3.id]
   json={"rule_ids": [r1.id, r2.id, r3.id, missing_id], "enabled": False}
   ```

3. **Add the `max_length=500` upper-boundary test.** `test_rejects_over_500_rule_ids` asserts 501 → 422, but nothing asserts 500 is accepted. One-liner off-by-one twin:
   ```python
   def test_accepts_exactly_500_rule_ids(test_client, test_session):
       rules = [make_rule(test_session) for _ in range(500)]
       resp = test_client.post(
           "/api/auto-creation/rules/bulk-update",
           json={"rule_ids": [r.id for r in rules], "enabled": False},
       )
       assert resp.status_code == 200
       assert resp.json()["updated_count"] == 500
   ```

## Non-blocking follow-ups (we'll track as beads post-merge)

- **Spurious writes in the merge-prune path.** `rule.actions = json.dumps(new_actions)` fires unconditionally even when `_apply_merge_streams_remove_non_matching` returned the input unchanged — causes no-op DB writes, `updated_at` bumps, and potential JSON key-order diffs on rules with no merge_streams actions. Guard with `if new_actions != actions:` in a follow-up.
- **`probe_on_sort` is a single shared DB column.** Applying only channel-sort=quality with `probe_on_sort=true` still writes the column, which can silently flip stream-quality probing on rules that already have `stream_sort_field='quality'`. The modal guard correctly prevents in-modal mismatches; the cross-rule implication is a separate discussion — worth a follow-up bead for a proper per-section split or a warning-UX round.
- **N+1 query pattern.** Per-id `session.query(AutoCreationRule).filter(...).first()` loop; one `WHERE id IN (...)` would be cleaner and catch missing ids upfront.
- **SQLAlchemy 2.x modernization.** `session.query(...).get(id)` → `session.get(AutoCreationRule, id)` (tests + router).
- **Frontend regression-lock could be tighter.** The test asserts `onApply` prop wasn't called; the stricter form would `vi.mock('../../services/autoCreationApi')` and assert `bulkUpdateRules` spy was never invoked. Current form captures the component contract but not the HTTP invariant. Acceptable to ship; worth tightening later.
- **Frontend happy-path assertion.** A "select 2 rules, toggle one section, click Apply, assert patch payload shape" test would catch field-name drift between types and backend. Post-merge bead.

## Ship notes

- CI is green.
- No conflicts with the wave-3 bd-eio04 work that just landed.
- Happy to merge as soon as the 3 test tweaks land. Thanks for the thorough iteration on this.

PR #113, opened 2026-04-22
Fix auto-create stream sorting reliability and quality ordering

stevencoutts:
**Why**
Auto-create stream sorting was inconsistent in a few common cases:

> No “sorting” entry (and no sort performed) when rules merged into existing channels (Channels Created = 0), because Pass 3.5 stream-reorder was gated on “owned” channels only.
> Name-based stream sort could appear to not work when streams had no StreamStats row, causing the sorter to fall back to "Stream <id>" instead of real names.
> Quality (Resolution) sort could put a black-screen/failed stream at the top simply due to resolution.

**What changed**

> Always log stream-reorder evaluation: Pass 3.5 now records an execution-log entry even when the channel is already in the correct order (no changes made).
> Run stream sort on modified existing channels: Pass 3.5 now includes channels the rule modified during this run (e.g. merge_stream / merge prune), not only channels created this run or pre-run managed, so sorting runs when a rule merges into existing channels.
> Make name-based stream sorting work without probe stats: When the channel payload contains stream {id, name} objects, we seed missing stream_name into the sort cache so stream_name / stream_name_natural ordering can still reorder correctly.
> Deprioritize bad streams in quality sort: Quality (resolution) stream sorting now respects ECM settings (deprioritize_failed_streams, failed_stream_sort_order) so black-screen/low-FPS/failed streams are pushed down instead of bubbling to the top.

**User-visible impact**

> Stream sorting is consistently applied when expected (including merge-into-existing runs).
> Execution details will show the stream-sorting step even when no reordering is necessary.
> Quality sorting is less likely to select obviously-bad streams as the “top” stream.

**Test plan**

> Run Auto-Creation with a rule that merges into an existing channel and has Stream Sort = Quality (Resolution), Descending.

> > Confirm the channel’s streams reorder as expected and an execution log entry is present.

> Run with Stream Sort = Stream Name on a channel where streams lack probe stats.

> > Confirm ordering updates using the stream names.

> Create a channel where the highest-resolution stream is flagged black screen (or failed/low FPS) and re-run.

> > Confirm that stream is deprioritized below healthy streams.


MotWakorb:
Thanks for this @stevencoutts — excellent root-cause write-up in the PR body, clean three-fix scope, and the unit-test coverage maps 1:1 to each bug described. Multi-persona review came back clean (no security/DB/SRE/architecture concerns). Much lower-risk than most PRs that size.

## Must-fix before merge (all small)

1. **Add three `### Fixed` entries to `CHANGELOG.md` under `[Unreleased]`.** Repo convention — every auto-create fix landing on `dev` carries a CHANGELOG line with the bead ref. Suggested text matching the style of the bd-yj5yi / bd-yui1k / bd-u9odj entries:
   ```markdown
   ### Fixed

   - Auto-Creation Pass 3.5 stream reorder now runs on channels modified via
     `merge_stream`/`merge_streams_prune` — previously silent when
     Channels Created = 0, leaving streams unsorted after a pure-merge run. (PR #113)
   - Auto-Creation name-based stream sort now works when probe stats are
     unavailable — the reorder path now seeds stream names from the
     channel's stream payload so natural-order and alphabetic sort have
     something to sort on. (PR #113)
   - Auto-Creation quality (resolution) sort now honors
     `deprioritize_failed_streams` — black-screen / failed / low-fps streams
     no longer promote to the top ahead of working higher-resolution
     streams. (PR #113)
   ```

2. **Create a bead and reference it in the CHANGELOG** for traceability (match `bd-yj5yi` / `bd-yui1k` style). I can create one on the maintainer side — reply with `create bead` and I'll file it; otherwise run `bd create enhancedchannelmanager \"...\" --external-ref gh-113` locally.

3. **Rebase on `dev`** — the PR shows `BEHIND`. `dev` has moved (wave 1 normalization work landed); fast-forward rebase, no conflicts expected.

## Non-blocking follow-ups (tracked as post-merge beads)

- The synthetic "already sorted" execution-log entry uses `stream_id: null`, which can cause React duplicate-key collisions when multiple channels log it in one run (`AutoCreationTab.tsx:1132`). The `entry.rules_evaluated ?? []` defensive guard added in this PR is correct; follow-up will normalize the envelope (explicit `entry_kind` discriminator or required-empty `rules_evaluated`).
- `merge_stream` / `merge_streams_prune` action-type strings in the gating filter are hardcoded — extract to a constants module so future action types that mutate streams don't silently bypass the gate.
- Quality-sort deprioritization tests only cover `is_black_screen`. Follow-up will add parametrized cases for `probe_status=timeout`, `is_low_fps`, and `deprioritize_failed_streams=False`.
- `[AUTO-CREATE-ENGINE]` prefix in the `stream_name` column is engineer-facing; UX will polish.

## Ship notes

- CI green. No wave-2 `bd-eio04` overlap (normalization work touches different code paths in the same file; rebase-level only).
- Zero Block-severity findings from the code/security/architecture review.

Happy to merge the moment the CHANGELOG + bead + rebase land. Thanks for the contribution.

stevencoutts:
Addressed the remaining must-fixes:

- Created bead for traceability: enhancedchannelmanager-f55hm (gh-113)
- Added the three ### Fixed entries under [Unreleased] in CHANGELOG.md referencing the bead
- Rebased fix/auto-create-sort-dev onto latest dev and force-with-lease pushed (clears “behind”)

Ready for re-review.

PR #163, opened 2026-04-24
Fix scheduled task email alerts + add recipient UI

stevencoutts:
**Summary**

> Fix scheduled task email notifications by routing “Email” alerts through the SMTP Alert Method so tasks deliver to configured to_emails recipients (bd-bq27w).
> Add Settings → Notification Settings → Email Alert Recipients UI to create/update SMTP alert recipients used by scheduled tasks (bd-lfmuw).
> Add vitest coverage for /api/alert-methods frontend API helpers and document the behavior in CHANGELOG.md under [Unreleased].

**Test plan**

> Configure SMTP in Settings → Notification Settings and confirm “Send Test Email” works.
>  Enter recipients in Email Alert Recipients and save.
>  Manually run a scheduled task with Send external alerts + Email enabled; confirm an email arrives on completion (success/warning/error depending on toggles).
> Run frontend tests: cd frontend && npm test.


MotWakorb:
## UX heuristic review — PR #163 (scheduled task email recipients)

Hey @stevencoutts — thanks for picking this one up. UX gate first, then Code Reviewer is queued behind me for correctness / API / test-quality. I'm trying to bundle everything into one round so you don't get ping-ponged.

Scope of this review: the **Email Alert Recipients** panel in `frontend/src/components/tabs/SettingsTab.tsx`. I'm not commenting on the backend `notification_service.py`, tests, or API shape — that's the next reviewer.

### Pattern decision (important — please read first)

You're using **plain-text comma-separated input**, and I'm keeping it that way. For a scheduled-task admin screen with typically 1–5 recipients this is defensible and simpler than chips; chip/tag UI would be over-engineering here. **Decision: plain-text input, `,` is the canonical separator, document it in the hint.** Don't accept `;` or newline as alternate separators — one pattern, one separator.

That decision raises the bar on validation, duplicate handling, and error recovery though, because with plain-text CSV you don't get per-item error anchoring "for free" the way chips would give you. Most of my Blockers below are direct consequences of that choice.

---

### Findings

#### Blocker 1 — No inline RFC 5322 validation (Nielsen #5: error prevention)

`SettingsTab.tsx` ~line 845 (the `<input>`) and `handleSaveSmtpRecipients` ~line 834.

Right now the only validation is `.filter(Boolean)` — `"notanemail, also-not-an-email"` will save successfully and silently break every scheduled task alert until someone notices emails aren't arriving. That's a high-blast-radius silent failure for a notification feature.

Minimum bar:
- Validate each comma-split token against a pragmatic email regex (the HTML5 `type="email"` regex is fine for RFC 5322 intent — don't roll your own; see [WHATWG's valid-email-address production](https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address)).
- Show the **specific invalid entry** in the error message, not "Invalid input" — e.g. "`not-an-email` is not a valid email address."
- Validate on blur / before submit, not on every keystroke (that's rage-inducing while the user is mid-typing).

Bonus: `<input type="email" multiple>` with `<form>` submission gives you browser-native validation on the comma-separated list for free, and screen readers announce it consistently. Worth considering before writing a custom validator.

#### Blocker 2 — Valid entries aren't preserved when save fails (Nielsen #9: help users recover from errors)

`handleSaveSmtpRecipients` treats the whole save as all-or-nothing. If the user types `"alice@example.com, typo, bob@example.com"` and save fails because of `typo`, they see a toast and the input still shows all three — they now have to find and fix one token in a comma-salad. Small lists today, but this scales badly.

Recommend: on validation failure, surface **which token failed** (inline error text under the input, red border on the field, or — easiest — the toast message names the bad entry). The two valid entries should stay in the field so the user only edits the broken one.

#### Blocker 3 — No duplicate detection (Nielsen #5)

`"alice@example.com, alice@example.com"` saves two copies to `to_emails`. Depending on how the backend iterates, that's either a duplicate notification per task run or a confusing recipient list in the "To:" header. De-dup case-insensitively on save and warn the user inline (not a hard block — just "Removed 1 duplicate"). RFC 5321 local-parts are technically case-sensitive but virtually no real-world mail system honors that, so lowercasing for dedup comparison is safe.

#### Recommend 1 — Empty-state copy is confusing (Nielsen #1 + empty state)

On first visit (`smtpAlertMethodId === null && smtpAlertRecipients === ''`), the user sees an empty text field and a "Save Recipients" button. The badge says **Unconfigured** but doesn't say what happens if they leave it that way. Scheduled tasks will silently not-email.

Suggestion: when `smtpAlertRecipients` is empty AND `smtpAlertMethodId === null`, show a short explainer *above* the input — "No recipients configured. Scheduled task email alerts won't be delivered until you add at least one recipient." The existing `section-description` is close, but it explains the feature, not the current state.

#### Recommend 2 — Post-save "saved" state isn't sticky (Nielsen #1: visibility of system status)

You show `notifications.success('Email alert recipients saved')` — good. But the toast disappears in ~3–5 seconds and the user has no durable indicator that *this* save succeeded. The badge flips from "Unconfigured" → "Configured" on the first save, but on subsequent saves (editing an existing list) there's no visual differentiation.

Light-touch fix: flash the input border green for 1–2s after a successful save, or show a "Saved at 10:47 AM" timestamp next to the badge. This is the kind of thing that pays off for an admin who's toggling recipients and wants to be sure their edit stuck.

#### Recommend 3 — Paste-list handling (Nielsen #5, efficiency)

If I paste `"alice@example.com; bob@example.com\ncarol@example.com"` from someone's email draft, the current code keeps it as a single invalid token because you only split on `,`. Given the pattern decision above (comma is canonical), I'd still normalize on paste: replace `;` and `\n` with `, ` in the onPaste handler so the user doesn't have to reformat whatever they copied. Document the canonical separator, but be forgiving about inputs.

This is the one place forgiveness *doesn't* violate the "pick one pattern" rule — we're normalizing on input, not accepting multiple patterns as equivalent.

#### Recommend 4 — Saving-state button affordance is good, loading affordance on the input is missing

The button's saving spinner + disabled state is great. But during the initial `smtpAlertRecipientsLoading` phase the input is disabled with no visible indicator of *why*. A user tapping through settings fast will see a frozen input and think it's broken.

Suggestion: show a subtle loading affordance (placeholder text `Loading recipients…` or a small spinner in the input's right-edge slot) during the load. Same heuristic as the saving state — users shouldn't have to guess what the system is doing.

#### Nit 1 — "Configured" badge logic has a rough edge

~line 2857: `(smtpAlertMethodId && smtpAlertRecipients.trim())` — this flips to "Configured" based on client-side draft state, so if the user types one char and then deletes it after having saved previously, the badge flickers to "Unconfigured" even though the server still has their saved list. Badge should reflect *persisted* state, not draft state. Easy fix: track a separate `savedRecipients` ref and compute the badge from that.

#### Nit 2 — WCAG 2.5.5 target size on the Save button

The `btn-test` class button looks fine on desktop but I don't know its computed height in your design tokens. [WCAG 2.5.5 (AAA)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced.html) is 44×44 CSS px; [2.5.8 (AA)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) is 24×24. This is probably already fine given the rest of Settings uses the same class — just flagging it's worth a quick DevTools check since this is a new interactive surface. No change needed if the shared class already passes.

#### Nit 3 — `htmlFor`/`id` pairing is correct, label text could be clearer

`<label htmlFor="smtpAlertRecipients">Recipients</label>` — the label "Recipients" in isolation is generic. Screen reader users navigating by form labels will hear "Recipients, edit text" with no context that it's for email alerts (the `<h3>` above provides visual context but isn't part of the accessible name computation). Consider `aria-describedby` pointing at the section description, or make the label `Email alert recipients`. Low priority.

---

### What would make this mergeable from UX's side

The three Blockers (inline email validation, preserve valid entries on error, duplicate detection) are the ones I'd want addressed before merge — they're all Nielsen #5/#9 violations that will produce silent-failure support tickets. The Recommends are genuine improvements but I'd take them as a follow-up bead if you'd rather keep this PR tight.

Once these are addressed, **Code Reviewer is next up** for correctness / API / test-quality review. Thanks again for the contribution — this feature's been in the broken state for a while and it's good to see it landing.

— UX

MotWakorb (commented):
## Code Review — Round 2 (after UX)

Hey @stevencoutts — picking up the second-round per the standup-authorized UX-first/CR-second sequencing. UX's consolidated review (comment above) stands as filed; I'm not re-raising any of their items, just layering CR-domain findings on top so you get one consolidated punch list.

**Verdict**: Changes Requested

The diff is substantively well-structured (good logging discipline, lazy `%s` formatting, sensible reuse of the existing `alert_methods` framework, clean API helpers + frontend tests). I'm requesting changes on one correctness item plus one test-coverage item. Everything else is Warn/Nit.

**Review Items**

| # | Class | Area | Finding | Suggestion |
|---|---|---|---|---|
| 1 | Warn | Correctness — silent state overwrite | `handleSaveSmtpRecipients` (`SettingsTab.tsx`) hard-codes `enabled: true`, `notify_info: false`, `notify_success/warning/error: true` on every PATCH. Today the Settings UI is the only frontend surface that touches the SMTP alert method, so this is mostly latent — but the moment a user (or a future admin UI) disables the method to silence email alerts or toggles notify_info on, the next "Save Recipients" click silently reverts those preferences with no UI affordance. The recipients-save is claiming a wider scope than its label says. | On the update path, send only `{ config }` (PATCH is partial by design — `AlertMethodUpdate` already has every field as `Optional`). On the create path, the defaults you've chosen are reasonable initial state; keep them there. This also keeps the API call honest about what the user just changed. |
| 2 | Warn | Test Quality — happy-path only | `test_dispatch_queues_email_via_alert_methods_when_smtp_method_configured` covers the success branch, but the new code adds three new failure branches: (a) `is_smtp_configured()` False → `results["email"] == False`, (b) no enabled SMTP `AlertMethod` row → `results["email"] == False` + warning log, (c) `send_alert_via_methods` raises → `results["email"] == False` + error log. None are tested. The success test also asserts `assert_awaited_once()` but doesn't check what was passed — the `channel_settings` payload, the title/message propagation, `results["email"] == True`. Per the skill: "Are negative tests present? Don't just test the happy path." Coverage for the failure branches matters more here than usual because they're the silent-no-op paths the original `TODO` was hiding. | Add three small tests (one per failure branch) and tighten the existing test to assert the kwargs passed to `send_alert_via_methods` and the returned `results["email"]` value. Parameterize if you prefer one test function. |
| 3 | Warn | Correctness — multi-method ambiguity | `methods.find(m => m.method_type === 'smtp')` picks the first SMTP method. The backend (`POST /api/alert-methods`) doesn't enforce uniqueness on `method_type`, so a user who created two SMTP methods (one for ops, one for billing, say) will have one of them silently edited and the other left untouched — with no UI affordance that two exist. Likely fine for the common case but worth knowing. | Either (a) filter by `name === 'Email'` to match what you create, or (b) when multiple SMTP methods exist, surface them all (or refuse to edit and direct the user to a future Alert Methods admin UI). I'd take (a) as the minimum bar for this PR and file the broader UX as a fix-forward bead. |
| 4 | Nit | Style — backend convention | `from models import AlertMethod as AlertMethodModel` and `import alert_methods_smtp  # noqa: F401` inside the function body. The `as AlertMethodModel` rename matches `routers/alert_methods.py` precedent — good. The `import alert_methods_smtp` inside the dispatch function is defensive but redundant: `main.py` already imports it at module level (lines 33-35) for side-effect registration. Safe to drop the inline import, or leave the comment explaining why ("defensive: function may be called from contexts that didn't import main"). Take or leave. | Drop `import alert_methods_smtp` from the function body, OR add a one-line comment clarifying it's defensive. |
| 5 | Nit | API — request-body shape | The frontend sends `to_emails` as a comma-joined string (`recipients.join(', ')`) while `alert_methods_smtp.py` accepts both string and list. The route stores config via `json.dumps`, so strings stay strings round-trip. The pre-existing convention is mixed (older code stores list, this PR stores string). Not your job to reconcile, but it's worth a fix-forward bead: the API should pick one canonical representation and document it in `docs/api.md` so the parsing branch in `alert_methods_smtp.py` can eventually go away. | No change needed in this PR. File a fix-forward bead for the canonicalization. |

**Note on UX's Blockers**

UX flagged three Blockers (inline RFC 5322 validation, preserve-valid-entries on partial save failure, duplicate detection). Those are inside UX's domain and I'm not re-classifying. From CR's side they're not correctness defects — `.filter(Boolean)` produces valid backend payloads — they're UX error-recovery findings. Address them per UX's review and the PR is good on that axis.

**Positive Observations**

- Logging discipline is on-point: `[NOTIFY-SVC]` prefix, lazy `%s` formatting throughout, three log levels used correctly (`debug` for "would send", `warning` for misconfiguration, `error` for exception). This matches `backend/CLAUDE.md` exactly.
- Reuses the existing `alert_methods` framework + `AlertMethod` model rather than scaffolding a parallel notification path. Good "reuse before creating" discipline.
- The frontend `extractToEmails` helper handles both array and string shapes defensively for the read path — good loose-read/strict-write hygiene.
- `api.test.ts` covers all three new helpers (list / create / update) with MSW handlers that assert the request body. That's the right level of test for an API helper layer — not over-mocking, not skimping.
- CHANGELOG entry is in the right section (`### Fixed`), references both bead IDs (`bd-bq27w`, `bd-lfmuw`), and explains the user-visible behavior change. Matches the project's CHANGELOG style.
- Surgical diff (~355 lines, focused) — no drive-by refactors, no scope creep. Easy to review.

**Handoff**

Once UX's three Blockers and my Warn items #1 and #2 land, this is mergeable. Items #3, #4, #5 are fix-forward candidates if you'd rather keep the PR tight — file a follow-up bead and I'll approve on the next push. Thanks for the contribution; this fixes a long-standing silent-no-op and the test scaffolding sets up cleanly for future Alert Methods admin work.

🤖 Reviewed by Code Reviewer persona via Claude Code

MotWakorb:
## Team review — Round 2 verdict: APPROVE with follow-up beads

Hey @stevencoutts — thanks for the responsive turnaround on UX's three Blockers and CR's two Warns. Re-review by the team confirms everything blocking is fixed, with line-level evidence:

**UX Round-2 verification (all 3 Blockers fixed):**
- B1 inline RFC 5322 validation — `isValidHtml5EmailAddress` (SettingsTab.tsx:436-442), blur + save validation, invalid token named in error with `role="alert"` + `aria-invalid`
- B2 preserve valid on error — `parseSmtpRecipients` returns `normalized: raw` on first invalid token; save handler returns early without overwriting state
- B3 case-insensitive dedup — `Set` keyed on `token.toLowerCase()`, "Removed N duplicate(s)" affordance

Plus all 4 Recommends addressed (sticky save status, paste normalization, empty-state explainer, loading affordance). Nice work.

**CR Round-3 verification:**
- #1 silent state overwrite — fixed at `SettingsTab.tsx:503-506`, update path now sends only `{ config }`; create path retains seeded defaults
- #2 negative-path tests — three new tests cover `is_smtp_configured()=False`, no enabled SMTP method row, and `send_alert_via_methods` raises; happy-path test now asserts kwargs and `results["email"] is True`
- #3 multi-method ambiguity — addressed via `name === 'Email'` fallback (option a)
- #4 inline import — explanatory comment added (acceptable per "leave comment" option)

**Security review:** No blockers. UI-side header-injection blocked by anchored regex; auth posture matches existing routes; no PII/creds in logs or fixtures.

**One non-blocking new finding** (3 personas independently flagged it): `smtpAlertRecipientsPersistedRef.current` is read directly in JSX at lines 568/570/581 to drive the Configured/Unconfigured badge + empty-state. Refs don't trigger re-renders, so the badge can go stale until an unrelated state change forces a re-render. Save still persists correctly — this is cosmetic. Filed as fix-forward bead `bd-uomm7` (we'll handle it).

### Follow-up beads filed (all on us, not you):
- `bd-uomm7` — useRef → useState for persisted-state badge
- `bd-cp14f` — extract + unit-test parseSmtpRecipients helpers
- `bd-6e8gv` — server-side to_emails validation (defense-in-depth, P3)
- `bd-9vz32` — canonicalize to_emails string-vs-list shape
- `bd-8rd5m` — document /api/alert-methods schemas in docs/api.md (P3)
- `bd-8tyv5` — add Notifications & Alert Methods section to user_guide (P3)

CI green (5 required checks). Merging now via merge-commit (project convention). Thanks again — this fixes a long-standing silent no-op and the test scaffolding sets up cleanly for future Alert Methods admin work.

— Team review (UX, CR, Security, Engineer, QA, Tech Writer, PM) via Claude Code

PR #228, opened 2026-05-11
Fix/scheduled job email metrics

stevencoutts:
**Summary**

- Scheduled task emails (SMTP / alerts) now use task-specific counters instead of misleading “Total Items” / “Success Count”: Auto-Create Channels matches the execution modal (streams evaluated/matched, channels created/updated, groups, conflicts, optional execution id / pipeline mode). Stream Probe uses clearer copy and fields (streams scheduled / ok / failed / skipped, black screen and low FPS). Bead: [enhancedchannelmanager-e3ts6].
- Full-success notification body for auto-creation uses the pipeline summary from TaskResult.message plus duration; stream probe gets an explicit one-line summary with totals and quality flags when relevant.
- Stream prober completion alert aligned with the same naming and summary pattern as the scheduled task path.

**Test plan**

- [x]  cd backend && python -m pytest tests/unit/test_task_notification_formatting.py -q
- [x]  cd backend && python -m pytest tests/unit/test_failed_stream_reprobe.py -q
- [x]  Run Auto-Create Channels with email alerts: body/metadata match execution details (no 27k “items” vs small “success” confusion).
- [x]  Run Stream Probe to completion with email alerts: scheduled total and ok/fail/skip (+ black screen / low FPS when non-zero) read clearly.

**Checklist**

- [x]  Bead linked in the commit message ([enhancedchannelmanager-e3ts6]).
- [x]  Tests updated or added (backend pytest) — backend/tests/unit/test_task_notification_formatting.py.
- [x]  CHANGELOG.md updated under [Unreleased] (Added / Changed / Deprecated / Removed / Fixed / Security). See [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Skip only for docs-only, test-only, or CI-only changes — note the reason. (Recommend Fixed for user-facing alert copy before merge.)


MotWakorb (changes requested):
Thanks for this, Steven — the task-specific alert fields are a real improvement and the two helper functions + unit tests on them are nicely done. One blocker and a couple of smaller items before merge.

## 🐛 Blocker — silently breaks the `probe_failures` min-failures alert threshold

This PR removes `failed_count` from the stream-probe alert metadata in **both** probe alert paths:
- `backend/stream_prober.py` `_finalize_probe_notification` — `alert_metadata` renames `failed_count` → `streams_failed`, but still passes `alert_category="probe_failures"`.
- `backend/task_engine.py` via `_task_execution_metadata_extra("stream_probe", result)` — emits `streams_failed` (not `failed_count`), and `_notify_task_result` is called with `alert_category="probe_failures"` for stream_probe.

But `backend/alert_methods.py` (~line 446) extracts the threshold input by name:

```python
failed_count = 0
if metadata and "failed_count" in metadata:
    failed_count = metadata.get("failed_count", 0)
...
self._should_alert_for_source(method_model.alert_sources, alert_category, entity_id, failed_count)
```

and `_should_alert_for_source` for `probe_failures` does `if failed_count == 0: return False` then `return failed_count >= min_failures`. After this PR that lookup always sees `0`, so **any alert method configured with a "only notify me when probe failures ≥ N" threshold stops firing.**

Smallest fix: keep `failed_count` in the metadata alongside the new `streams_failed` key in both paths (purely additive — the new clearer keys stay, the threshold keeps working). Alternatively teach `alert_methods.py` to also accept `streams_failed`. Either way, please add a test that exercises the `probe_failures` threshold path with the new metadata shape so this doesn't regress again.

## 🧹 Style — nested ternary in the cancellation message

In `backend/task_engine.py` `_execute_task`, the `result.error == "CANCELLED"` branch builds the message as a nested ternary embedded in a multi-line string concatenation. It parses correctly but is hard to verify at a glance. The partial-failure case immediately below it was refactored to a clean `if/else` assigning `warn_msg` first — please follow the same pattern here (`cancel_msg = ...` then pass `message=cancel_msg`).

## Minor (your call) — auto-creation success message fallback

`_success_task_completion_message` only uses the nice pipeline summary when `task_id == "auto_creation" and details:` — if `result.details` is empty/falsy it falls through to the generic `"Successfully completed. N items processed…"`, which for auto-creation with e.g. `total_items=27304` reintroduces the exact "27k items" confusion this PR fixes. Probably rare, but a one-line guard (or a minimal summary built from `result.message`) would close it.

Everything else looks good — CHANGELOG entry under Fixed, bead linked, CI green, and handling `result.duration_seconds is None` is a nice incidental robustness fix. Happy to pair on the alert-metadata change if useful. (FYI #220 is now merged into `dev`.)

MotWakorb (approved):
Re-reviewed after the push (commits `ba32b82e` + `0137efa8`). All three points addressed cleanly:

1. **probe_failures threshold restored** — `failed_count` is back in both probe metadata paths (`stream_prober._finalize_probe_notification` and `_task_execution_metadata_extra("stream_probe", …)`) alongside the new readable `streams_*` keys, each with a comment explaining why; `alert_methods.send_alert` now honors `failed_count` or falls back to `streams_failed` with safe int coercion; and there are two new regression tests pinning the threshold path (`test_send_alert_probe_failures_threshold_uses_streams_failed_metadata` + the below-threshold case).
2. **Cancellation message** — refactored to a clean `if/else` assigning `cancel_msg`, matching the `warn_msg` pattern below it.
3. **Auto-creation success fallback** — `_success_task_completion_message` now uses `result.message` for auto-creation even when `details` is empty, so the "27k items" wording can't reappear.

CI green on `0137efa8`, merged up to date with `dev`. Thanks Steven (and Cursor) — merging.

PR #229, opened 2026-05-11
Fix/alembic 0005 sqlite batch

stevencoutts:
## Summary
- Harden Alembic revision **`0005`** (`20260427_2100_0005_auto_creation_quality_m3u_tie_break`) for SQLite: add **`quality_tie_break_order`** and **`quality_m3u_tie_break_enabled`** in **separate** `batch_alter_table` blocks so **`upgrade head`** no longer raises **`CircularDependencyError`** between those two columns.
- Make **`0005` `upgrade()`** **idempotent** by introspecting **`auto_creation_rules`** before each **`add_column`**, covering the case where Alembic failed earlier, **`create_all()`** created the columns, and **`alembic_version`** was still **`0004`** (fixes **`duplicate column name`** on restart).
- **`downgrade()`** drops each column only if it exists. Changelog + bead **`[enhancedchannelmanager-m2k7p]`**; ties back to quality tie-break work **`enhancedchannelmanager-3j9su`** / **`0005`**.
## Test plan
- [ ] `pytest backend/tests/unit/test_database_bootstrap.py` — Alembic bootstrap / upgrade paths (adjust if your CI only runs a subset).
- [ ] Fresh DB: `alembic upgrade head` through **`0005`** completes without error.
- [ ] Drifted DB (columns present, **`alembic_version` = `0004`**): **`0005`** applies cleanly and stamps head.
- [ ] Deploy migration file to container, restart ECM, confirm **`GET /api/health/schema`** shows **`up_to_date`** and no **`[DATABASE] Alembic bootstrap failed`** log line.
## Checklist
- [x] Bead linked in the commit message (`[enhancedchannelmanager-m2k7p]`).
- [ ] Tests updated or added (backend `pytest`, frontend `vitest`) — *no new tests in this PR; migration-only. Add a targeted bootstrap test if you want CI to lock this in.*
- [x] `CHANGELOG.md` updated under `[Unreleased]` → **Fixed** (Alembic **`0005`** reliability).
- [ ] Docs updated if behaviour or interfaces changed (`docs/`, runbooks, ADRs) — *not required; schema intent unchanged.*


MotWakorb (approved):
Re-reviewed after the merge-up — fix is correct and well-targeted.

- **Split-batch + introspect-before-add** is the standard Alembic-on-SQLite pattern for this class of bug, and the comment cites the exact internal that errors (`_adjust_self_columns_for_partial_reordering`).
- The `added_m3u` flag is a nice subtle touch: the post-add `alter_column` (stripping the temporary `server_default`) only runs when this run actually added the column — so a column already present from `create_all()` isn't double-touched.
- Idempotent in both directions; existing data preserved; schema *intent* unchanged (same two columns, same types).
- CI green on the merged head; the conflict was purely the metadata files (CHANGELOG + .beads/issues.jsonl) and is resolved.

One follow-up that's NOT a merge blocker — and that the PR body explicitly called out: there's no regression test that exercises the migration's two paths (fresh-upgrade through 0005, and "columns already exist at 0004 → upgrade is idempotent"). I'll open a follow-up bead so the same bug can't regress unnoticed. Thanks Steven (and Cursor) — merging.

PR #405, opened 2026-05-23
fix(mcp+api): P1/P2 MCP tool fixes from the lq38l live sweep

MotWakorb:
## What

Fixes the P1/P2 findings from the live MCP test sweep (epic `enhancedchannelmanager-lq38l`). All changes are gate-green and **live-verified** against the running MCP/backend containers.

| Bead | Pri | Fix |
|------|-----|-----|
| lq38l.5 | **P1** | `merge_channels`: guard self-merge (target in sources) before any backend call; report per-result outcome instead of false success on a bad target |
| lq38l.1 | P2 | `list_tasks` reads the real `task_name` field (was always "Unknown") |
| lq38l.2 | P2 | `list_normalization_rules` sources from the rules endpoint → real per-group counts + names (was "0 rules"); adds `normalization_list_rules` to `_endpoint_contracts.py` |
| lq38l.3 | P2 | `reorder_streams` permutation guard — refuses partial/empty lists that would silently detach streams |
| lq38l.4 | P2 | backend write routes map upstream Dispatcharr 4xx → clean 400/404 (were opaque 500): channels/m3u/epg/profiles routers |
| lq38l.6 | P2 | `build_channel_lineup` counts/matches only channels created by the call (was counting the whole group) |
| lq38l.8 | P2 | real count/URL keys: m3u stream count (derived) + `server_url`; epg `epg_data_count`; groups-with-streams no longer shows a false "0 streams" |
| lq38l.9 | P2 | `stats` tools use `path_args` (not `path_params`) — `get_channel_popularity` + `get_user_channel_breakdown` no longer crash |
| lq38l.10 | P2 | `create_backup` reads the binary zip response (no more UnicodeDecodeError) |

(lq38l.7 — exact-match-only dedup auto-accept — is being added to this branch next.)

## Verification

- mcp-server suite: **284 passed, 0 failed**; backend router suite: **131 passed, 0 failed** (verified on the integration branch, not just isolated worktrees).
- Each fix re-tested live against the connected MCP after deploying to `ecm-ecm-1` + `ecm-ecm-mcp-1` (e.g. `list_tasks` shows real names; `update_channel(badid)` returns 404 not 500; `reorder_streams` partial list is refused; `create_backup` succeeds; Provider 1 shows 2624 streams).
- Disposable `MCPTEST_` entities used for write tests were all reverted (cleanup verified).

## Notes

No version bump (fix PR to dev; versioning handled at release cut). Findings filed/tracked under epic `enhancedchannelmanager-lq38l`.

### DIFF
--- a/backend/routers/channels.py
+++ b/backend/routers/channels.py
@@ -2028,7 +2028,15 @@ async def update_channel(channel_id: int, data: dict):
             logger.debug("[CHANNELS] No changes detected for channel %s", channel_id)
 
         return result
+    except HTTPException:
+        raise
     except Exception as e:
+        # A missing channel id (or bad group/logo/etc.) surfaces as an upstream
+        # 4xx — map it to a clean 4xx instead of an opaque 500 (bd-lq38l.4).
+        mapped = upstream_http_exception(e)
+        if mapped is not None:
+            logger.warning("[CHANNELS] Update channel %s rejected by Dispatcharr: %s", channel_id, e)
+            raise mapped
         logger.exception("[CHANNELS] Failed to update channel %s: %s", channel_id, e)
         raise HTTPException(status_code=500, detail="Internal server error")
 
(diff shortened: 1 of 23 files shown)


MotWakorb:
lq38l.7 (exact-match-only merge_if_found auto-accept + 'N | ' prefix strip in the dedup matcher) is now included (commit 14556f54) and live-verified: fuzzy 90% candidate → pending_merge (no auto-accept); exact 100% → auto-accepts. This PR now covers all 10 P1/P2 beads (lq38l.1–.10). Suites green: mcp-server 285, backend dedup 116, backend routers 131.

PR #474, opened 2026-06-03
fix(auto-creation): batch merge journal writes

### LINKED ISSUE #473: [Bug]: Large auto-creation runs cause high memory/swap pressure and container crashes
### ECM Version

0.17.3

### Priority

Priority 2: ECM is severely degraded

### What is the bug?

Large auto-creation runs can cause very large SQLite WAL growth and high memory/swap pressure, eventually causing the ECM container to crash/restart.

On my deployment, the issue appears during auto-creation runs that merge many provider streams into channels. The audit/journal path currently writes each live `merge_stream` journal row with its own `journal.log_entry(...)` call, which opens a session, inserts one row, commits, and closes for every merge. Functionally that preserves the audit trail, but operationally it creates heavy SQLite WAL churn on large lineups.

Observed impact:

- Host/container memory reached roughly `15.63 GiB / 16.00 GiB`.
- Swap reached roughly `14.11 GiB / 64.00 GiB`.
- ECM crashed/restarted during or after auto-creation.
- The SQLite WAL grew to multiple GiB before checkpointing.

### Steps to Reproduce

1. Run ECM `0.17.3` with SQLite WAL mode enabled.
2. Configure a large provider lineup / multiple M3U sources.
3. Configure auto-creation rules that merge many streams into existing channels.
4. Run auto-creation.
5. Observe high memory/swap usage and container instability/crashes.
6. Check ECM logs for very large WAL checkpoint output.

### Screenshots / Logs / Artifacts

Relevant log lines from my deployment:

```text
INFO ts=2026-05-30T12:43:20.814Z logger=database msg=[DATABASE] SQLite PRAGMAs applied: journal_mode=wal synchronous=NORMAL foreign_keys=ON (memory=False) trace_id=-
INFO ts=2026-05-30T12:43:29.225Z logger=database msg=[DATABASE] WAL checkpoint: 3631.3 MB -> 0.0 MB trace_id=-
```

Observed resource pressure at the time:

```text
Memory usage: 97.71% (15.63 GiB of 16.00 GiB)
SWAP usage: 22.04% (14.11 GiB of 64.00 GiB)
```

### Additional Context

I opened PR #474 with a minimal proposed fix: batch auto-creation `merge_stream` journal writes while preserving the existing per-merge audit payload.

The proposed approach keeps these audit fields intact:

- `category="auto_creation"`
- `action_type="merge_stream"`
- `entity_id`
- `entity_name`
- `before_value={"stream_ids": [...]}`
- `after_value={"stream_ids": [...], "match": ...}`
- `batch_id=str(execution_id)`

It reduces SQLite commits from one per merge to roughly one per batch, then flushes pending journal rows at the end of the auto-creation pipeline.

I was temporarily hotpatching/mounting modified files in my container to avoid crashes, but that is brittle across future image updates. An upstream fix would remove the need for that workaround.

Mirabis:
## Summary

- add `journal.log_entries()` to persist multiple journal rows in one transaction
- buffer auto-creation `merge_stream` audit rows in `ActionExecutor` and flush them at threshold / pipeline completion
- preserve the existing per-merge audit payload, including `batch_id`, stream-id-only before/after values, and scored-fuzzy provenance
- update focused unit coverage for batch journal writes and buffered merge flushes

Fixes #473.

## Why

Large auto-creation runs can perform many merge journal writes. Committing every merge row individually increases SQLite WAL churn and can produce very large WAL files during a single run. Batching keeps the same audit trail while reducing transaction frequency.

## Validation

Passed:

```text
uv run --python 3.12 --with-requirements backend/requirements.txt python -m pytest backend/tests/unit/test_journal.py backend/tests/unit/test_auto_creation_executor.py backend/tests/unit/test_auto_creation_engine.py backend/tests/unit/test_scored_fuzzy_merge.py backend/tests/unit/test_0hjrk_run_summary_invariants.py --tb=short --no-header -p no:warnings --no-cov
261 passed in 7.96s
```

Passed:

```text
uv run --python 3.12 --with-requirements backend/requirements.txt python -m py_compile backend/journal.py backend/auto_creation_executor.py backend/auto_creation_engine.py backend/tests/unit/test_journal.py backend/tests/unit/test_auto_creation_executor.py backend/tests/unit/test_auto_creation_engine.py backend/tests/unit/test_scored_fuzzy_merge.py backend/tests/unit/test_0hjrk_run_summary_invariants.py
```

Local broader gate note:

```text
uv run --python 3.12 --with-requirements backend/requirements.txt python -m pytest backend/tests/unit backend/tests/integration backend/tests/routers --tb=short --no-header -p no:warnings
7 failed, 4475 passed, 1 skipped in 285.20s
```

The 7 local failures are outside this change:

- `backend/tests/unit/test_date_placeholders.py::test_expand_date_placeholders_format` is platform/libc-sensitive locally: `%Q` expands to `Q` on this macOS runner.
- 6 `backend/tests/routers/test_alert_methods.py` SMTP cases fail because this local test registry exposes only `create_test` and `probe_failures_threshold_test`, not `smtp`.

Full `backend/tests/` also reported E2E setup errors locally because those tests attempted to connect to a live server and received `Connection refused`.


MotWakorb (changes requested):
Thanks for this, @Mirabis — and for the unusually clear write-up and repro on #473. The batching approach is the right lever for the WAL churn, the shared `_build_journal_entry` helper is clean, and the audit payload (`batch_id`, stream-ids-only `before/after`, scored-fuzzy provenance) is faithfully preserved. We ran a correctness review plus an SRE reliability pass and want to take this — with one required change and two test additions.

### Required before merge

**1. Guarantee the end-of-pipeline flush on the exception path (durability).**
`executor._flush_journal_buffer()` in `auto_creation_engine._process_streams` sits just before `return results` with no `try/finally`. `_process_streams` has several `await` points *after* live merges have already executed against Dispatcharr (reorder, reconcile-orphans, EPG retry, probe). If any of those raises, the exception propagates up and the flush is skipped — so up to `_journal_flush_threshold` (100) buffered rows describing merges that **actually happened** are lost. That's exactly the aborted/bad-run scenario the per-merge journal (bd-0emgo.5) exists to make recoverable, and the old per-row commit didn't have this gap. Please wrap so the flush always runs:

```python
try:
    ...  # passes 2–6
    return results
finally:
    executor._flush_journal_buffer()
```

`_flush_journal_buffer` already no-ops on an empty buffer and isolates its own failure, so a `finally` flush is safe on the exception path.

**2. Regression test for #1** — a test where a later pipeline pass raises after a merge, asserting the buffered entry is still flushed.

**3. Threshold-flush test** — none of the current tests drive ≥100 merges to prove the automatic threshold flush fires and clears the buffer; please add one.

### Minor
- The `_journal_merge` docstring still says failures are "swallowed (`journal.log_entry` already does this internally)" — that path no longer calls `log_entry`; please reword to describe the buffer/flush model. A one-line note that `timestamp` is now stamped at flush time (not per-merge) would also help the next reader.

Everything else — correctness equivalence (incl. the falsy empty-dict case), credential safety, dry-run isolation, `batch_id` — checks out.

One heads-up for #473 itself: this PR meaningfully reduces WAL growth, but our reliability review found the ~15 GiB memory crash has a **separate** root cause (the in-memory `execution_log` + end-of-run `json.dumps`, on a disk-backed `/config`), so we'll address that separately — details in my comment on #473. Thanks again; genuinely good contribution.

Mirabis:
@MotWakorb  I think I addressed all review feedback in the follow-up commit.

Changes:
- moved the buffered merge-journal flush into a `finally` path so later pipeline failures do not drop audit rows for merges that already happened
- added a regression test where a later pass raises after live merges and verifies the buffered journal entries still flush
- added a threshold test that drives the default 100-entry buffer flush and verifies the buffer is cleared
- updated the `_journal_merge` docstring to describe the buffered flush model and note that timestamps are stamped at flush time

```text
263 passed in 5.42s
```

PR #483, opened 2026-06-07
update channel in memory during the auto create after each step

lpukatch:
## Summary
  • Auto-creation rules execute sequentially against a simulated in-memory channel state.
  • Several actions (like  assign_tvg_id ) updated the API but forgot to update this in-memory state.
  • Example: Setting a  tvg_id  and then assigning EPG caused the EPG matcher to use the old  tvg_id  and fall back to incorrect fuzzy matching.
 
Updated the execution logic so that all actions changing channel properties update both the database/API and the simulated channel. This ensures downstream actions within a given rule operate against the most current channel state.

### DIFF
--- a/backend/auto_creation_executor.py
+++ b/backend/auto_creation_executor.py
@@ -1132,6 +1132,8 @@ class ActionExecutor:
 
             if updates:
                 await self.client.update_channel(channel_id, updates)
+                # Update simulated state so subsequent actions see the changes
+                channel.update(updates)
 
             return ActionResult(
                 success=True,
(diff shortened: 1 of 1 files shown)


MotWakorb (approved):
Reviewed: correct and well-scoped. The handlers now mutate the same simulated-channel dict held in `self._channel_by_id`, so downstream actions in a rule see fresh values — fixing the assign_tvg_id → assign_epg staleness. Verified all `update_channel` call sites; the ones not touched here (rename, add-stream, post-run EPG verify) were already correct. All 15 required checks green.

Non-blocking follow-ups: (1) dry-run branches of assign_logo/assign_profile/set_channel_number don't update simulated state, so multi-action dry-run previews can still show stale values for those props; (2) no regression test locks in the fix.

PR #692, opened 2026-07-18
v0.17.6-0125: align rule-editor stepper interaction models (bead 09x38.10)

MotWakorb:
## What

The Standard Rule editor and the Event Sync Rule editor shared the identical `.modal-stepper` step-pill strip but implemented **opposite** interaction models. Event Sync is a true wizard (one step rendered at a time, Back/Next footer); the Standard editor renders all three regions (Logic → Targeting → Output & Run) as siblings in one continuous form and the "stepper" only scroll-jumps to an already-visible section. Learning one editor built the wrong mental model for the other (UX pass-2, DOM-confirmed).

This restyles the Standard editor's nav strip as an on-page **section nav / table-of-contents** so it no longer masquerades as the wizard.

## Choice + risk assessment — visual differentiation over wizard conversion

The bead's preferred outcome was full wizard conversion. Assessed and chose the acceptable fallback (visual differentiation) for concrete reasons:

- The Standard editor is deliberately architected as a single "when X do Y" statement. Its right-rail **live intent sentence, advisory analyzer chips, and save hint all synthesize fields from every region at once** — seeing the whole rule while editing any part is the design point. A one-step-at-a-time wizard would degrade that.
- **State coupling / test surface:** the entire `RuleBuilder.test.tsx` suite (and the sort/numbering hints, merge-scope warnings, intent card) query fields across all three regions on a single render. A wizard that unmounts inactive steps would force a large test rewrite for no behavioral gain — exactly the destabilization the bead flagged as the trigger for the fallback.
- The problem is purely the **visual masquerade**; restyling the nav kills the wrong-mental-model at the root with a minimal, behavior-preserving change.

The nav is now a **"Jump to" table-of-contents idiom**: left-aligned underline text tabs with a leading list icon + label, `aria-current="location"` (a place in the form, not `"step"` in a process), and no circled step-number badges — visually unmistakable from the Event Sync wizard's equal-width numbered step pills. New `.rule-section-nav` styles in `RuleBuilder.css`; the shared `.modal-stepper` primitive is untouched and still owned by the Event Sync wizard.

**No behavior, validation, or save-path change.** Clicking a section still smooth-scrolls to it; the scroll-spy still tracks the active section.

## Round-trip proof

New test `round-trip invariant across section navigation — bead 09x38.10 > open -> step through every section -> save yields an unchanged payload`: renders a fully-configured existing rule (every option/sort/scope/orphan/flag field set), clicks through Targeting → Output & Run → Logic touching nothing else, clicks Save, and asserts `onSave` receives a payload deep-equal to the loaded config. Green.

Other new/adapted tests (`section nav (table-of-contents idiom, not a wizard) — bead 09x38.10`):
- lists sections as plain labels (no wizard step numbers)
- carries a "Jump to" affordance
- marks the active section `aria-current="location"`, never `"step"`
- keeps every region rendered while navigating (not a step-at-a-time wizard)

## Gates
- `npm run build` — green
- `npx vitest run` — **1908 passed** (RuleBuilder suite: 68 passed)
- `eslint --max-warnings 0` — clean
- `tsc --noEmit` — clean

## Live verification (read-only, http://localhost:6100)
Deployed the bundle (footer confirmed `v0.17.6-0125`). Opened **Create Rule → Standard**: nav renders as "JUMP TO Logic / Targeting / Output & Run" underline tabs, `aria-current="location"` on Logic, no `.modal-stepper` in the builder; clicking "Output & Run" moved the active marker and all three regions stayed in the DOM. Opened the existing **Test Sync** Event Sync rule: still the numbered wizard step pills (1 Scope / 2 Matching / 3 Behavior / 4 Review) with `aria-current="step"` and Back/Next — visually distinct. **Cancelled out of both — nothing saved, nothing run.**

## How to test
1. Channel Pipeline → Create Rule → Standard rule → observe the "Jump to" section nav; click sections to scroll; confirm it doesn't look like a wizard.
2. Channel Pipeline → edit the Test Sync (Event Sync) rule → confirm the wizard step pills are unchanged and visually distinct.

## Security
- [x] No hardcoded secrets; frontend-only UI restyle, no endpoint/data change.

## Deployment
- [x] No infrastructure or DB changes. Rollback: revert commit.

bead enhancedchannelmanager-09x38.10


MotWakorb:
Superseded by PO decision: full wizard conversion chosen over visual differentiation (bead 09x38.10). A new PR will implement the Standard editor as a true step-at-a-time wizard.

PR #711, opened 2026-07-19
v0.17.6-0142: guard client-error URL port parsing (bead enhancedchannelmanager-k6ud9)

MotWakorb:
## What

`POST /api/client-errors` returned 500 when a reported stack trace or message contained a URL with an out-of-range port (e.g. `http://host:99999/path`). `backend/obfuscate.py`'s `obfuscate_url()` accessed `parsed.port` on a `urlparse()` result unguarded — Python raises `ValueError` for a port outside 0-65535 (or non-numeric), and the exception was uncaught. The endpoint whose entire job is collecting client error reports crashed on exactly the kind of malformed input those reports contain, so the reports were lost. Observed live when a browser reported a stale-bundle fetch error after a frontend deploy.

## Why

Bead enhancedchannelmanager-k6ud9 — pre-existing bug, not caused by any recent change.

## Fix

- `parsed.port` access in `obfuscate_url()` is now wrapped in `try/except ValueError`, treating an unparseable port the same as "no port." Obfuscation behavior is otherwise identical.
- Audited `obfuscate.py` and its caller (`routers/client_errors.py`) for other unguarded `urlparse` attribute accesses (`.hostname`, `.scheme`, `.path`, `.params`, `.query`, `.fragment`, plus `urlparse()` itself). None found — `.port` was the only hazardous accessor; everything else was already safe or already wrapped.
- Note for the backlog (deliberately NOT touched, out of scope for this bead): `backend/routers/channel_pipeline.py:4247` has the identical unguarded-`parsed.port` pattern in the debug-bundle Dispatcharr-URL redaction path. Lower risk (operator-configured URL, not attacker-controlled client input) but the same failure mode — worth a follow-up bead.

## Tests (TDD — written first, confirmed red, then fix, confirmed green)

- `backend/tests/unit/test_obfuscate.py` (new file, 10 tests) — direct unit coverage of `obfuscate_url`/`obfuscate_text` against out-of-range, negative, non-numeric, and IPv6-bracketed hostile ports, plus confirming normal obfuscation is unchanged for valid ports / no port.
- `backend/tests/routers/test_client_errors.py::TestHostileUrlPortInStack` (2 tests) — endpoint-level regression proving `POST /api/client-errors` returns 204 (not 500) for a hostile-port URL embedded in `stack` or `message`.

## Gates

Full backend suite:
```
7541 passed, 13 skipped in 342.25s (0:05:42)
```
Baseline was 7529 passed, 13 skipped — this PR adds exactly 12 new tests, zero failures, zero regressions.

`python scripts/check_version_consistency.py` — all 3 touchpoints agree at `0.17.6-0142`.

`cd frontend && npm run build` — clean (version bump only; no frontend logic touched).

## Live container verification (before/after)

Deployed the patched `backend/obfuscate.py` to `ecm-ecm-1` and restarted. Same hostile payload before and after:

```json
{"kind":"boundary","message":"TypeError","stack":"fetch failed for http://host:99999/path in bundle.js:1:1", ...}
```

**Before** (unpatched, running code):
```
HTTP_STATUS:500
Internal Server Error
```
Container logs showed the uncaught `ValueError: Port out of range 0-65535` traceback.

**After** (patched + restarted):
```
HTTP_STATUS:204
```
Structured log confirms clean scrubbing — `stack` rewritten to `"fetch failed for http://example.com/path in bundle.js:1:1"` (host+port replaced, no port shown since it was invalid — correctly treated as "no port"). Also spot-checked `-1`, `abc`, and `65536` as the port value — all 204 post-fix.

## Version

`0.17.6-0141` → `0.17.6-0142` (frontend/package.json, backend/routers/backup.py APP_VERSION, backend/main.py FastAPI version — all three in lockstep).

## CHANGELOG

New `### Fixed` entry under `[Unreleased]`.


MotWakorb:
## AI team review — quick mode

**Disposition: ready to land first; no blocking defect found.** Reviewed by Security, Architecture, Project Management, Engineering, UX, Code Review, Database, SRE, QA, and Technical Writing. Effective tier: **small-team** (`ecm-core` + `ci-release`).

### Important landing constraint

PR 711 and PR 712 both branch from build 0141 and edit `CHANGELOG.md`, `backend/main.py`, `backend/routers/backup.py`, and `frontend/package.json`. Land **711 first**, then update/rebase 712 onto the resulting `dev`, retain both changelog entries and build 0143, and rerun the required checks. Current green CI validates the PRs independently, not their combined landing state.

### Non-blocking finding

- **Low / follow-up — malformed parse failures can still bypass URL redaction.** `backend/obfuscate.py` returns the original URL when `urlparse()` itself raises. A malformed unmatched-bracket URL such as `http://user:password@[host/path` can therefore survive the scrubber and reach the structured client-error log through `backend/routers/client_errors.py`. The endpoint is authenticated/rate-limited and exposure is limited to local-log confidentiality, so this is not a merge blocker at this tier and is pre-existing rather than introduced by this patch. Recommended follow-up: fail closed on parse failure and add unmatched-bracket/userinfo coverage.

### Test-quality improvements, not blockers

- The router regression proves hostile URLs return 204, while the unit test proves obfuscation. A future endpoint-level log-capture assertion would lock the full security postcondition.
- The PR body identifies an equivalent unguarded `parsed.port` access in `backend/routers/channel_pipeline.py`; keep that consolidation/fix separate from this narrow patch so the known residual does not get lost.

### Positive evidence

The changed hazardous accessor is narrowly guarded, valid-port/no-port behavior is preserved, focused unit and endpoint regressions are present, current CI is green, and the two earlier CodeQL comments were addressed and were not re-raised. No UX, database, migration, persistence, or rollback concern was found.

MotWakorb:
## Team Review — 10-persona parallel review

**Depth:** quick · **Effective tier:** home-lab (`ecm-core`, `dev-tooling`, `docs` per `COMPONENTS.md`; no `ci-release` files touched, so no cross-tier escalation)

**Verdict: CLEAR — no blocking findings.** Security, Code Reviewer, Project Engineer, QA, Architect, DBA, SRE, Technical Writer and PM all returned CLEAR. UX reported no surface in this PR.

### What the review confirmed

The root cause is correctly identified and the fix is minimal: `parsed.port` was the single raising accessor in `obfuscate.py`, and the guard narrows to `has_port = False`. Security independently verified — not merely by reading the diff — that `.hostname`, `.scheme`, `.path`, `.params`, `.query` and `.fragment` do **not** raise for out-of-range, negative, non-numeric, malformed-percent-encoded or IPv6 inputs, so the audit claimed in the PR body holds. Because the whole `netloc` is replaced wholesale rather than reconstructed, hostile port digits never reach the redacted output or the log line. The failure direction is safe.

SRE endorsed the fail-open choice specifically: for a telemetry *collector*, silently absorbing malformed input beats both crashing and logging every hostile URL, and the report itself still persists (204).

The CodeQL remediation on the second commit (`4d5045dd`) — replacing `assert "example.com" in result` with an exact `urlparse(...).hostname ==` comparison — was called out as the correct fix for that finding class rather than a suppression, since the substring form would also match `evil-example.com`.

### Non-blocking findings

**1. Query-string credentials are never scrubbed** *(Security, INFO — pre-existing, not introduced here)*
`obfuscate.py:47` passes `parsed.query` through to `urlunparse` verbatim, so `http://host/path?token=SECRET` keeps `token=SECRET` in the obfuscated output. Untouched by this diff and out of scope for this fix, but it surfaced during a full-function audit: a client-error stack embedding an API-key-bearing URL reaches the log unredacted. Worth its own bead rather than a change here.

**2. Identical unguarded `parsed.port` at `routers/channel_pipeline.py:4247`** *(Code Reviewer, Nit)*
Same failure shape in the debug-bundle redaction path. Lower risk (operator-configured URL, not attacker-supplied) and correctly declared out of scope in the PR body — flagged so it isn't lost.

**3. Test-assertion strength — reviewers disagree** *(QA vs Code Reviewer)*
QA notes six unit tests and both endpoint tests assert only "does not raise" / status 204 without asserting the redacted output, so a regression that treated a hostile port as *valid* would still pass them. Code Reviewer's position is that this is the correct shape for a crash-regression test — they would fail if the `try/except` were removed — and that output correctness is pinned by the separate valid-port tests. Both are partly right: the crash is well covered, the output-on-hostile-input is not. Cheap to close by adding `result.endswith("/path")` to the existing cases; not worth blocking a merge over at this tier.

### Tier calibration

At startup tier or above, findings 1 and 3 would be actual findings rather than notes, and CodeQL would be a required gate rather than a manually applied fix. At home-lab neither is expected.

---
*Reviewed by: security-engineer, code-reviewer, project-engineer, qa-engineer, it-architect, database-engineer, sre, technical-writer, project-manager, ux-designer. Findings are advisory; merge authority rests with the PO.*

PR #712, opened 2026-07-19
Selection bar Move-to-group submenu: type-to-filter at large group counts (bead hzzcv)

MotWakorb:
## Summary
- Adds a type-to-filter input to the top of the selection bar's Move-to-group submenu (`frontend/src/components/SelectionActionBar.tsx`), so the ~2,940-group live instance no longer requires scrolling to find a target group. Filtering matches the Channels pane's existing searchable group filter (bead mn8, `ChannelsPane.tsx`): case-insensitive substring match on the raw, untrimmed query.
- "Uncategorized" and "New group…" are pinned outside the filter — fixed actions, not filterable named groups — so both stay reachable at any filter state, including a zero-match filter, which renders `No groups match "<query>"` instead of a blank panel.
- Preserves the existing keyboard contract: focus lands in the filter input the instant the submenu opens (by click or ArrowRight); arrows traverse the filtered results; Enter activates the focused group; Escape closes the submenu only (returning focus to its trigger) — never the whole menu, never the selection. Left/Right/Home/End stop propagating to the menu's roving-focus handlers while the input is focused, so normal text-cursor editing inside the filter still works.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint -- --max-warnings 0` — clean
- [x] `npm test` — 147 files / 2068 tests passing (baseline 147/2061 + 7 new tests)
- [x] `npm run build` — succeeds
- [x] Live-verified on the container (~2,940-group live dataset) via Playwright MCP: opened Edit Mode, selected a 50-channel group, opened ⋮ More → Move to group, typed `sport` (narrowed 2,940 → 409) then a whitespace-containing query ` network` (narrowed to 55, confirming untrimmed substring semantics), then a unique query, arrowed to the target, pressed Enter, confirmed the move (pre-existing ChannelsPane confirmation modal), applied the edit-mode change, and confirmed via the ECM MCP tool that the channel's `group_id` persisted to the new group. Also verified Escape-from-submenu keeps the outer More menu open, returns focus to the "Move to group" trigger, and leaves the 50-channel selection intact. Live-verification screenshots were captured as local, transient artifacts and are not part of this PR.

Bead: enhancedchannelmanager-hzzcv


MotWakorb:
## AI team review — quick mode

**Disposition: changes requested before merge.** Reviewed by Security, Architecture, Project Management, Engineering, UX, Code Review, Database, SRE, QA, and Technical Writing. Effective tier: **home-lab** (`ecm-core`).

### Must fix

- **Keyboard focus escapes the Move submenu.** `getMenuItems()` in `frontend/src/components/SelectionActionBar.tsx` queries every enabled `[role="menuitem"]` under the outer `menuRef`, and `moveFocus()` wraps across that global list. After the filter sends focus into the submenu, ArrowUp from `Uncategorized` can move to the outer `Move to group` trigger, and ArrowDown after `New group…` can continue into unrelated outer Selection/Profile actions. This contradicts the stated contract that arrows traverse the filtered results. Scope roving focus to the active submenu, define boundary/wrap behavior, and add first-item ArrowUp plus last-item ArrowDown coverage.

### Important / should fix

- The nested `role="menu"` directly contains a textbox and a generic clear button alongside menuitems. Validate the accessibility tree and use coherent composite semantics; an `aria-label` supplies a name but does not resolve the mixed menu interaction model.
- The filter removes the native focus outline and relies only on a border-color change; the clear control is roughly an 18px target with no explicit focus-visible style. Add a visible focus treatment and a more usable hit target.

### Non-blocking improvements

- The zero-results message is visual-only; a polite status announcement would improve screen-reader feedback.
- Automated scale coverage uses 50 groups while the motivating instance has about 2,940. Live Playwright validation at real scale is good evidence and keeps this non-blocking at home-lab tier; add a repeatable 3,000-item behavior smoke test if this path regresses.
- Tests do not pin Left/Right/Home/End caret behavior despite the implementation contract.
- The PR body says screenshots `hzzcv-1` through `hzzcv-6` are in the repo root, but they are untracked and absent from the PR. Attach/link them in the PR or remove that location claim.

### Landing order

PR 711 and PR 712 independently edit the same release/version files from build 0141. Land **711 first**, then update/rebase this PR, preserve both changelog entries and build 0143, and wait for the rebased head checks to pass. Current green CI does not validate the combined state.

### Positive evidence

The filter solves a demonstrated live-scale discoverability problem, preserves pinned actions and an explicit empty state, resets cleanly, keeps group IDs unchanged through the existing callback, and has strong baseline behavior tests plus live validation. No security, database, persistence, migration, or rollback blocker was found.

MotWakorb:
## Team Review — 10-persona parallel review

**Depth:** quick · **Effective tier:** home-lab (`ecm-core`, `dev-tooling`, `docs` per `COMPONENTS.md`; no `ci-release` files touched, so no cross-tier escalation)

**Verdict: CLEAR with two WARNs — no blocking findings.** Security, Code Reviewer, Project Engineer, QA, Architect, DBA and SRE returned CLEAR. UX and Technical Writer returned WARN; neither is a merge blocker at this tier.

### What the review confirmed

The acceptance risk on this bead was whether adding a text input would break the menu's existing keyboard contract. It doesn't, and the review verified the mechanism rather than taking the claim on faith.

`moveFocus`/`getMenuItems` re-query the live DOM (`querySelectorAll('[role="menuitem"]:not(:disabled)')`) on each keypress instead of holding a stored index, so a narrowing filter **cannot** produce a stale-focus bug — focus always resolves against what is currently rendered. That is the right architecture for this feature. `handleMoveFilterKeyDown` calls `stopPropagation()` on Left/Right/Home/End/Up so the outer roving handler never double-fires, and Escape is `preventDefault + stopPropagation` before routing to `closeSubmenuToTrigger('move')`, which is what keeps the document-level Escape listener from clearing the channel selection. That listener independently ignores `INPUT`/`TEXTAREA` targets — belt and braces, not a conflict.

Also noted favourably: the pre-existing generic submenu-navigation test was *repointed* to the Sort Streams submenu rather than deleted when Move's first-focus target changed, so the older contract stays covered. "Uncategorized" and "New group…" are pinned outside the filter and remain reachable at any filter state. Filtering is `.includes()` on an already-fetched, already-authorized list — no regex (no ReDoS surface), no new API surface, no injection surface (group names render as JSX text children). Performance is a single O(n) pass over ~2,940 pre-sorted groups per keystroke; the sort happens once in a `useMemo` upstream and is not repeated.

### Non-blocking findings

**1. Clear-filter button is not reachable by arrow-key navigation** *(UX, WARN — the most substantive finding here)*
`SelectionActionBar.tsx:423-428`. The clear button renders only when the filter has text and is not part of the roving-focus set, which walks `[role="menuitem"]` only. A keyboard user who filters to zero matches sees the empty-state message and the clear button but cannot arrow to it — backspace is the only path back. Defensible at home-lab, and the pinned entries mean the user is never trapped, but it breaks the expected menu-navigation model. Two cheap options: make the button always visible (disabled when empty) so it joins the navigable set, or add a test pinning keyboard recovery from a zero-match filter. Related, from Code Reviewer (Nit): `ArrowUp` inside the input is swallowed with no local handling, so there is no wrap-to-last or return-to-trigger — only Escape goes back. Both sit in the same seam.

**2. User-guide drift contradicts shipped behaviour** *(Technical Writer, WARN — caused by PR #710, not this PR)*
`docs/user_guide/channels-streams/stream-dedup.md` lines 16 and 35–36 still instruct operators to "right-click a channel-less stream and choose **Create channel(s) in group**". That context menu was removed in build 0141 by PR #710. The drift is real and user-facing, but it predates this PR and this PR did not cause it — recording it here because it belongs to the same feature series and would otherwise be lost.

**3. `stopPropagation` behaviour is implemented but untested** *(QA, minor)*
Removing the `stopPropagation()` calls would not fail any test. A test that types into the filter, presses ArrowRight, and asserts the input keeps focus would close it.

**4. Empty-state is not a live region** *(UX — higher-tier note, not a finding at home-lab)*
The "No groups match …" div has no `role="status"`/`aria-live="polite"`, so screen-reader users may not hear it. Would be a finding at startup tier.

**5. Test fixture fragility** *(QA, very minor)*
The filter test relies on `"Group 23"` being unique among generated `"Group 1".."Group 50"` fixtures; a distinctive sentinel string would be more robust to fixture changes.

### Merge sequencing

Independently flagged by Code Reviewer, Project Engineer and PM: this PR and #711 both branch from `566232ea` (build 0141) and bump the same three version touchpoints to different targets (0143 vs 0142). Whichever merges second needs a trivial rebase of `frontend/package.json`, `backend/main.py` and `backend/routers/backup.py`, or it will fail the Version Consistency gate. CHANGELOG hunks land in different sections (Added vs Fixed) and will not conflict. Mechanical only — no design conflict.

---
*Reviewed by: security-engineer, code-reviewer, project-engineer, qa-engineer, it-architect, database-engineer, sre, technical-writer, project-manager, ux-designer. Findings are advisory; merge authority rests with the PO.*

MotWakorb:
## Review fixes pushed

Addressed the actionable team-review findings in commits `6da69465`, `325ba2f5`, and `2630e559`:

- converted the searchable Move popup to a labelled non-modal chooser dialog with native buttons;
- contained ArrowUp/ArrowDown/Home/End navigation inside the chooser, including from the Clear control;
- returned ArrowUp from the first destination to the filter and clamped ArrowDown at New group;
- preserved native Left/Right/Home/End editing in the filter;
- added keyboard Clear recovery, Escape coverage from input/Clear/destinations, and polite result announcements;
- restored a visible filter focus ring and enlarged the Clear target to 32px;
- corrected role-based tests so filtering assertions are non-vacuous;
- corrected the PR body to describe screenshots as local transient verification artifacts.

TDD evidence: the first focused run failed on boundary focus escape and missing status; the reviewer follow-up reproduced Clear-control escape. Final local gates: SelectionActionBar **29/29**, TypeScript clean, ESLint clean. Independent second-round Code Review and QA both passed tip `2630e559`.

PR 711 is still blocked by missing mandatory `.agents/review-gate.json` / trusted `ai-team/*` checks. This branch therefore remains based on build 0141 and must be rebased onto `dev` after 711 legitimately merges; final CI must rerun on that rebased head.

MotWakorb:
CI follow-up: updated head `2630e55912b7cf59b87bdaf4fae9b6355f6ba1a3` completed with **24 successful, 8 skipped, 0 failing, 0 pending** checks. This validates the current pre-rebase branch. A final gate run is still required after PR 711 merges and this PR is rebased onto the resulting `dev`.

MotWakorb:
## Post-rebase verification complete

PR 712 was rebased onto `dev` after PR 711 merged. Current head: `93c56652e26fdf1556205640871785cad2660a53`.

Integration evidence:
- PR 711 merge `dc437b6e` is an ancestor and its implementation/tests remain byte-identical.
- All version touchpoints report `0.17.6-0143`.
- CHANGELOG retains both build 0142 and build 0143 entries.
- The three keyboard/accessibility remediation commits remain patch-equivalent to the independently reviewed versions.
- Independent post-rebase Code Review: PASS.
- Independent post-rebase QA: PASS.
- GitHub CI: **24 successful, 8 skipped, 0 failing, 0 pending**.

The PR is ready for the PO merge decision.

PR #714, opened 2026-07-20
[enhancedchannelmanager-ixcf1] Add Pending Merges bulk actions

### LINKED ISSUE #642: [Feature]: Bulk Actions for Pending Merges

This feature would allow the user to have easy access to bulk actions in the pending merges section.  Merge all, clear all, select all, deselect all, clear selected, merge selected.

_No response_

### Additional Context

_No response_

MotWakorb:
## What

Adds all six requested Pending Merges bulk actions: Select all, Deselect all, Merge selected, Clear selected, Merge all, and Clear all.

Queue-wide reads use one admin-gated `GET /api/channel-merges/snapshot` request, yielding a deterministic pending-only snapshot before mutation. The endpoint accepts optional `group_id`; status and group filters are applied before ordering and the safety cap. The client carries the active group scope through list, refresh, Select all, Merge all, and Clear all requests.

Scope changes invalidate stale requests, loading state, materialized views, and unconfirmed dialogs, preventing an earlier group from overwriting or mutating the active group. Already-confirmed work continues against the snapshot the operator approved.

Bulk mutations reuse the existing authorized single-row endpoints sequentially and prevent duplicate submissions. Partial failures remain selected and retryable with their backend error, including failures beyond row 200. To bound browser work, at most 200 snapshot rows are mounted at once with accessible Previous/Next navigation and a live row-range status.

Cancel restores the original paginated view when safe, and initiating controls show the loading state while the snapshot is fetched.

## Why

Closes #642.

Bead: `enhancedchannelmanager-ixcf1`

Design: ADR-008 §D10.

## How to test

1. Open Channel Manager → Pending Merges.
2. Select individual rows, then exercise Merge selected and Clear selected.
3. Use Select all on a queue larger than one page and confirm every pending record is selected.
4. Exercise Merge all and Clear all; confirm the count/consequence prompt and progress state.
5. Cancel after queue materialization and confirm the original paginated view returns.
6. Simulate failures beyond row 200, use bounded row navigation, and confirm each failure remains selected with its server error and enabled retry control.
7. Refresh with a selection and confirm the queue/count update coherently from one snapshot.
8. Change group scope while list/snapshot requests or a confirmation are open; confirm stale data and actions are discarded.

## Verification

- Backend focused router suite: 42 passed; snapshot-specific suite: 10 passed.
- Frontend focused API and Pending Merges suites: 113 passed.
- Full frontend suite: 2,103 passed before the final focused regression additions.
- TypeScript, ESLint, and Vite production build passed.
- Python compilation, documentation checks, and `git diff --check` passed.
- Version consistency passed at `0.17.6-0145`.
- Independent code review approved through `248eff29`.
- Independent database review approved scoped query semantics, index usage, predicate-before-cap behavior, cap boundaries, and deterministic tie ordering.

## API and security

- `GET /api/channel-merges/snapshot` is guarded by `RequireAdminIfEnabled`.
- It executes one ordered pending-only database read, optionally scoped by `group_id`, returns at most 20,000 records, and fails closed with HTTP 409 when the scoped queue exceeds that cap.
- Explicit non-admin denial and predicate-before-cap coverage are included.
- No new dependency, schema, or migration.

## Documentation

- API contract, user guide, ADR-008 §D10, changelog, developer deployment guide, and project architecture documentation are updated.
- Select all is documented as spanning the complete scoped pending queue.
- Merge all is documented as irreversible in ECM.

## Deployment

- Production GHCR deployments remain a single atomic image update; no infrastructure or database migration is required.
- For the manual development workflow, deploy backend router changes and restart the backend first, wait for `/api/health/ready` with the documented bounded probe, then run `scripts/deploy-frontend.sh`.
- Manual rollback uses reverse dependency order: restore the previous frontend first, then restore/restart the previous backend.
- Full copy-pasteable commands and failure behavior are documented in `CLAUDE.md` and `docs/project_architecture.md`.


MotWakorb:
## Code Review: Pending Merges Bulk Actions

### Verdict: Approved

The final head was independently reviewed across all six requested controls, bulk pagination, partial failures, count/selection consistency, duplicate-submission protection, accessibility, security, performance, maintainability, versioning, and rendered layout.

The first review reproduced two blockers at 51 rows: off-page failures became invisible, and Select all covered only the first page. Commit `edb8497c` fixes both with queue-wide selection and visible/retryable off-page failures.

Final evidence:

- Queue-wide Select/Deselect all and selected actions are covered at 51 rows.
- Merge all and Clear all preserve a failed 51st row, its backend error, selection, and retry controls.
- Empty state requires both no rendered rows and a zero total.
- Refresh reconciles selected IDs against the full queue.
- No new backend/auth/dependency surface.
- Version `0.17.6-0145` passes consistency and advance guards.

No remaining blocking findings. Non-blocking note: selecting extremely large queues materializes the full queue in the DOM; acceptable for the current home-lab tier.

MotWakorb:
# Team Review — PR #714 (Pending Merges bulk actions)

**Depth:** quick · **Component:** `ecm-core` · **Effective tier:** home-lab
**Participants:** code-reviewer, project-engineer, QA, UX, DBA, SRE, architect, PM, technical-writer (9/10 — security-engineer did not return; its concrete questions were verified directly, see §Security)

## Summary

Solid, well-tested engineering with one **verified blocking bug** in the mainline multi-page path. Partial-failure handling, pagination reconciliation, and the sequential-mutation design all hold up under trace. The blocker is a rendering regression the current test suite structurally cannot catch. Everything else is fast-follow or backlog.

---

## 🔴 Must-fix (blocks merge)

### F1 — Multi-page "Merge all" / "Clear all" blanks the UI mid-sweep
**Flagged by:** code-reviewer · **Independently verified in branch source**

`runBulkAction` builds the queue-wide snapshot into a local `targets` (`PendingMergesPage.tsx:234`) but never calls `setRows(targets)` — unlike `handleSelectAll`, which does `setRows(allRows)` at line 205.

All three render branches gate on `rows.length`:
- bulk toolbar — **including the `Merging X of Y` progress label** — line 339
- row list — line 432
- empty state — line 421, which *additionally* requires `totalRows === 0`

On a 201-row queue with a 50-row first page: after the first ~50 successes, `setRows(prev.filter(...))` (line 276) drains `rows` to 0 while `setTotalRows` (line 277) leaves `totalRows` at ~151. Toolbar false, list false, empty state false — **the operator sees a blank page while 151 mutations keep firing silently.**

Note the bug is masked on failure: failed rows are pinned back into `rows` (lines 298–302), so it only manifests when the sweep is *succeeding* — the common case.

**Fix:** `setRows(targets)` after the snapshot is built, before `window.confirm`, so the visible list matches the confirmed count.

### F2 — Test gap that let F1 through
**Flagged by:** code-reviewer, QA

`'snapshots every page before Merge all…'` (test.tsx:441) asserts only `getPendingMerges`/`acceptPendingMerge` call counts and args — never rendered DOM mid-sweep. This is the one mock-count-theater test in an otherwise strong suite. Add a mid-sweep assertion (e.g. `findByText(/Merging \d+ of 201/i)` after the first tick) so this regression fails loudly next time.

---

## 🟡 Should-fix

| # | Finding | Flagged by |
|---|---|---|
| F3 | **Confirmation is bare `window.confirm()`**, diverging from this codebase's own precedent — `ChannelsPane.tsx`'s `bulkDeleteConfirmModal` is a styled dialog with an explicit *"This action cannot be undone"* line. Current copy is identical whether clearing 3 rows or 201, and never states irreversibility. | UX, technical-writer |
| F4 | **No cancel/abort once a sweep starts.** Sequential `await` loop with no abort path; all controls disabled. On a large queue against a slow backend this is a long, uninterruptible commitment. Loop structure makes an abort-flag check per iteration cheap. | UX, SRE, code-reviewer |
| F5 | **Progress not announced to screen readers.** `Merging X of Y` lives in the button label with no `aria-live`/`role="status"`. A fully successful 201-row sweep gives an AT user zero feedback. The codebase already uses `role="status" aria-live="polite"` in `SelectionActionBar` — pattern exists, not reused. Accessibility is not tier-waived. | UX |
| F6 | **Failure-matrix gaps:** no test for multiple failures in one sweep, a mostly/entirely-failing sweep (backend degrades at row 10 of 201), or a total-count shift *between* the page-1 and page-2 snapshot fetches — the offset-pagination skip case the PR description explicitly claims to solve. | QA |
| F7 | **User guide not updated.** `docs/user_guide/channels-streams/stream-dedup.md` documents only per-row actions. No mention of the six bulk actions or that "Clear all" voids the entire queue. | technical-writer |

---

## 🟢 Backlog / could-fix

| # | Finding | Flagged by |
|---|---|---|
| F8 | `Select all` materializes the entire queue into a non-virtualized `<ul>`. Acceptable at home-lab; track as a known limitation if confidence-threshold defaults or queue volume change. | project-engineer |
| F9 | Component grew ~260 → 544 lines and now owns loading, reconciliation, per-row mutation, and a bulk engine. `pageSize = 200` duplicated as a bare literal across three call sites. Extract a `usePendingMergesQueue` hook next time this file is touched. | project-engineer |
| F10 | Double-submit guard reads render-time values, not a ref. Correct *today* only because `window.confirm` blocks the main thread — **if F3 replaces it with a non-blocking modal, this guard must move to a `useRef` set synchronously.** Coupled fix. | code-reviewer |
| F11 | `Select all` shows no loading state during its paginated fetch. | UX |
| F12 | CHANGELOG entry leads with implementation detail over operator benefit. | technical-writer |

---

## Resolved conflict

**code-reviewer said Block; project-engineer said "no blocking implementation bugs."**

The engineer traced the same function and correctly observed that a success on an *off-page* row is a harmless no-op filter — but did not follow through to the *on-page* rows, which do get filtered out until `rows.length` reaches 0. I verified the render gates directly in branch source (lines 339 / 421 / 432). **Code reviewer is correct; the engineer's conclusion is wrong on this specific point.** F1 stands as a blocker.

## Agreements worth recording

- **Sequential-not-concurrent is deliberate and correct**, not an oversight — documented in the component doc comment (lines 28–31) to avoid amplifying concurrent Dispatcharr writes. Independently endorsed by DBA, architect, SRE, and engineer.
- **Architectural divergence from the server-side bulk pattern** (`/bulk-commit`, `/bulk-merge`, `/probe/bulk` exist elsewhere) is real but justified — each pending merge is an independent decision with no cross-row transactional dependency. No follow-up needed. *(architect, DBA)*
- **Data layer is clean.** Each row's status flip + journal write commits in its own transaction, so a mid-sweep failure leaves no half-updated state. *(DBA)*
- **Diagnosability is above par for the tier** — `[DEDUP]`-prefixed INFO logs plus `pending_merge_journal` rows let an operator reconstruct any sweep. *(SRE)*

## Security

The security persona did not return. Its two concrete questions were verified directly:
- **The two backend edits are benign** — pure version bumps `0.17.6-0144` → `0.17.6-0145` (`main.py` FastAPI version, `backup.py` `APP_VERSION`). The PR's "no new backend endpoint, auth surface, dependency, or data model" claim **holds.**
- **No XSS path** — server error strings render as JSX text children (`{loadError}`, `rowErrors[row.id]`); React escapes them. No `dangerouslySetInnerHTML` in the file.

Residual security-adjacent concern is destructive-action confirmation strength, already captured as F3.

## Verification-claims audit

The PR body's verification section was checked rather than accepted:

| Claim | Status |
|---|---|
| "2,089 tests passed" | ✅ Verified in Frontend Tests job log (`148 files / 2089 tests passed`) |
| "TypeScript, ESLint passed" | ✅ Verified in same job log |
| "Version consistency … 0.17.6-0145" | ✅ Verified via `gh pr checks` |
| "Vite production build passed" | ⚠️ Not confirmable from the Frontend Tests job; runs in the separate image-build workflow |
| **"Playwright rendered UI checked at desktop resolution"** | ❌ **Unverifiable** — no spec, screenshot, or artifact in the diff or on the PR. `e2e/dedup-flow.spec.ts` already covers this component and was not extended. |
| "Independent code review approved" | ⚠️ A substantive review exists and its content checks out against the diff (it caught a real first-revision bug). Process *independence* can't be confirmed from repo evidence. |
| "All gates passed" | ⚠️ Was stale at review time (`BLOCKED`, checks in flight). **Now CLEAN** — all required checks pass. |

Recommendation: state "manual check, no artifact" rather than asserting Playwright coverage that carries no regression value.

## Positive observations

- Partial-failure retry semantics work exactly as claimed — off-page failed rows pinned into the list with verbatim backend error, checkbox re-checked, retry controls enabled — and are tested with **real DOM assertions**, not mock counts (test.tsx:493–576).
- Test fixtures are genuinely multi-page (51- and 201-row, real `pageSize` 50/200 split), not toy data — QA specifically checked for and did not find the two-item-fixture anti-pattern.
- Stale-selection reconciliation on refresh is a real edge case, correctly handled and tested.
- Scope matches GH #642 and bead `ixcf1` exactly — six actions requested, six delivered, no creep.
- CSS/naming follow `docs/style_guide.md`; logging matches existing `EventSyncReviewQueue.tsx` precedent.

---

*Review calibrated to home-lab tier per `COMPONENTS.md`. At startup tier, F6 would be a hard gate and E2E coverage of the multi-page path would be required; neither is treated as blocking here. No beads created — findings go to the PO for prioritization.*

MotWakorb:
## Addendum — Security Engineer review (late return)

The security persona returned after the main review was posted. Participation is now **10/10**; the "§Security — did not return" caveat above is superseded. Its findings are verified and **net-positive** — one item is a correction, not an addition.

### Correction: the confirm-count TOCTOU is not reachable

The main review treated destructive-action confirmation strength as an open concern. Security traced the ordering and found the implementation is **better than assumed**, which I verified independently:

- `runBulkAction` completes the full `getAllPendingRows()` sweep (line 234) **before** computing `count` (line 246) and **before** `window.confirm` (line 253).
- The mutation loop iterates only over that exact `targets` array, so it can never process *more* rows than the confirmed count.
- If a row is concurrently resolved elsewhere between snapshot and mutation, the backend fails safe — `accept` returns 200 on double-accept, `dismiss` returns 409 if already merged — and the row is pinned back into view with its error, re-selected for retry, never silently counted as success.

**The "operator confirms 12, gets 400 cleared" failure mode is not reachable from this code path.** Worst case is an accurate count with a few rows individually erroring and staying visible. This strengthens the case for shipping F3 (styled confirm modal) as a fast-follow rather than a blocker.

### Confirmed: auth surface unchanged

`backend/routers/channel_merges.py` — which owns `/api/channel-merges/{id}/accept` and `/dismiss` — is **not touched by this PR**. Both endpoints remain `RequireAdminIfEnabled`-gated (lines 896, 1131, verified). The bulk UI is N calls to already-admin-gated endpoints and cannot bypass or weaken authorization. No client-side-only enforcement anywhere.

Combined with the version-bump-only backend diff and the no-XSS finding already recorded, the PR's security claims hold in full.

### Elevated: F4 (no cancel) — Medium, and now the top decision item

Security independently reached F4 and rated it **Medium**, the only finding it flagged as warranting a product decision rather than silent acceptance. Its framing: after `window.confirm`, the sweep runs to completion with no abort path — buttons show progress but are disabled, not converted to Cancel. On the >1-page path this feature exists for, that is a multi-second-to-multi-minute irreversible operation the operator cannot stop, including immediately after realizing they picked the wrong scope.

That makes **four personas** on F4 (UX, SRE, code-reviewer, security) — the strongest convergence in this review. Security explicitly does **not** block the PR on it and expects pushback that it is scope creep beyond bead `ixcf1`; it asks for an explicit PO call rather than a default.

### Tier note

At startup+ tier security would additionally require a per-user audit trail distinguishing "operator clicked Merge all" from 50 individual accepts, and a server-side batch endpoint with an idempotency key. Neither is a finding at home-lab.

---

**Net effect on the review:** no new blockers. F1 remains the sole must-fix. F4 gains weight and is now the clearest decision point for the PO.

MotWakorb:
# Team Review — validation of fix commits `8807af3f` and `cd7f13bf`

Ten-persona review pass, calibrated to **home-lab** tier per `COMPONENTS.md` (ecm-core / docs / dev-tooling; no `ci-release` workflow changes). Re-review scoped to validating the fixes made after the F1–F7 findings.

**Verdict: the fixes are real, not paper-over. No persona blocks the merge.**

## Evidence the tests aren't theater

The post-fix test file was overlaid onto the pre-fix implementation (`edb8497c`) in a throwaway worktree and run:

- **17 of 29 tests fail against pre-fix code; 29/29 pass against fixed code.**

A suite that mirrored the implementation would pass against both. The failures land squarely on the risk areas the fixes targeted: queue-wide snapshotting, off-page failure retention, duplicate-submission guards, Stop semantics, and the accessible confirm dialog.

Independently re-run by reviewers rather than taken from the PR body: `vitest` (29 pass), `tsc --noEmit` (clean), scoped ESLint (clean), `gh pr checks` (all required green on tip).

## F1–F7 disposition — all verified against source

| Finding | Status | Evidence |
|---|---|---|
| F1 blank-UI mid-sweep | Fixed, better than minimal | `setRows(targets)` before confirm (`:316-322`); progress panel renders from its own block, not gated on `rows.length`, so it survives the list draining |
| F2 mock-count-only test | Fixed | 201-row test asserts real DOM + live `role="status"` text |
| F3 `window.confirm` | Fixed | `window.confirm` gone entirely; `ModalOverlay` (`:742`) with focus trap, Escape, focus restore |
| F4 no cancel/abort | Fixed | `stopRequestedRef` (`:87`), checked per iteration (`:400`), `Stop` button (`:608`); cooperative, re-selects remainder |
| F5 no `aria-live` | Fixed | `role="status" aria-live="polite"` (`:577-578`) |
| F6 failure-matrix gaps | Fixed | All-failing sweep, overlapping-page dedup, count-shift between snapshot passes, bounded fail-closed |
| F7 docs | Fixed | `stream-dedup.md` "Resolving merges in bulk" — accurate against implementation |
| F10 double-submit | Fixed unprompted | `bulkLockRef` / `confirmingRef` (`:85-86`), ref-based |

## New findings

**1. [Warn] `cancelBulkIntent` never restores paginated rows.** `requestBulkAction` materializes the full snapshot via `setRows(targets); setTotalRows(targets.length)` before showing the dialog (necessary for the F1 fix), but `cancelBulkIntent` (`:474-478`) resets only `bulkIntent` and the lock ref. Opening "Merge all" on a 500-row queue and clicking **Cancel** leaves all 500 rows materialized for the rest of the session — extending the known unbounded-DOM limitation from an opt-in action (Select all) to an accidental one. No correctness bug; no test covers open-then-cancel at >1 page.

**2. [Warn] `loadRows` still uses the old single-pass page walk.** The fingerprint/2-pass safety net added by `cd7f13bf` went into `getAllPendingRows` but not into its sibling at `:104-132`, which runs on **Refresh** when `selectedIdsRef.current.size > 0 && response.total > response.merges.length`. Same bug class F6 was written to close, reachable via a different button. This branch never executes in CI — the only Refresh test uses a 3-row single-page fixture, so the condition is never true. Since `getAllPendingRows` is now dependency-free (`useCallback(..., [])`), having `loadRows` call it is a straightforward follow-up.

**3. [Should-fix] "Select all" scope is undocumented.** The CHANGELOG states that Select all spans the entire queue across pagination; `stream-dedup.md` does not. Verified against `handleSelectAll`, which calls `getAllPendingRows()` and replaces both `rows` and `selectedIds` with the whole queue. One sentence closes it.

**4. [Watch] `cd7f13bf` solves a real bug at arguably the wrong layer.** Unlike `8807af3f` (a genuine bug fix), `cd7f13bf` adds a bespoke ~90-line client-side two-pass fingerprint protocol to reconstruct, across N sequential HTTP calls, a read-consistency guarantee a single backend query has for free. It is careful, correct engineering that fails closed with an honest operator-facing message — and it is the right call not to build a server-side endpoint now. The recorded trigger condition: **if this reconciliation logic ever needs a third fix, replace it with a small backend "give me every pending id right now" read (~15 lines) rather than deepening the client protocol.** No server-side bulk path was bypassed — `bulk_merge_duplicate_channels` is a different feature; Pending Merges only has `/accept` and `/dismiss`, and ADR-008 §D9 argues for exactly this operator-paced per-row model.

**5. [Nit] Bulk-all now costs ≥2 full traversals, always.** The `totalRows <= rows.length` early-out was removed; the algorithm structurally requires two consecutive matching passes. Fine at this tier, but it is a silent 2× minimum on a previously single-fetch path, with no loading affordance on the clicked button. Worth a code comment for whoever wonders why Select all on a 3-row queue fires 4+ requests.

**6. [Medium] Test gap at the bulk-snapshot page boundary.** `BULK_PAGE_SIZE = 200`, but the off-page failure tests use `id === 51` — within the *first* bulk fetch. The two tests that cross 200 rows inject zero failures. So "failures beyond page 1" is covered for the 50-row list boundary but never for the 200-row bulk boundary — the path `cd7f13bf` most recently rewrote. Suggest one parameterized case: 201+ rows, failure at id 250.

**7. [Low] "Merge all" is not reversible via ECM.** Each accept performs a Dispatcharr-side stream attach with no compensating unmerge, and DBAS backup does not snapshot `pending_merges` rows. Pre-existing per-row behavior, but doing it N times in one click raises the stakes — the confirmation dialog is the entire safety net. Worth one line in `stream-dedup.md`.

**8. [Nit] Housekeeping.** PR body says "20 focused tests" — actual is 29 (stale since the fix commits). ADR-008 §D10 explicitly anticipated this feature but is uncited by the PR/CHANGELOG/bead; the bead it referenced (`qpgsx`) is stale/missing.

## Verified clean

- **Auth/security:** `/accept` and `/dismiss` remain `RequireAdminIfEnabled`-gated and untouched. Bulk amplifies an already-authorized capability rather than creating a new trust boundary. No `dangerouslySetInnerHTML`; backend `detail` strings render as escaped JSX text. No secrets logged. Client loops are bounded and sequential — no concurrency burst.
- **Data:** No schema, migration, or query change. `backend/routers/backup.py` is confirmed a pure `APP_VERSION` bump with nothing riding along. Each accept/dismiss is its own atomic, idempotent, journaled transaction — a mid-sweep failure cannot half-write a row. SQLite is in WAL mode with sequential single-row commits; no lock-contention risk.
- **Ops:** Frontend-assets-only deploy path unchanged; rollback is a straight revert. Container restart mid-sweep leaves processed rows committed and the remainder cleanly `pending` — recoverable without DB surgery.
- **Versions:** `0.17.6-0145` consistent across `main.py`, `backup.py`, `package.json`, CHANGELOG.

## Done well

The fingerprint-pass algorithm is careful engineering against a genuinely hard problem (snapshotting a concurrently-mutating list with only offset pagination), and it fails closed rather than proceeding on stale data. Test fixtures are deliberately adversarial — 51/201/401 rows, overlapping page windows, alternating flapping totals, a page-cap-exceeding queue — not toy scale. The accessibility work (focus trap both directions, Escape, focus restore to the exact triggering button across six possible triggers, `aria-live` progress) goes beyond what F5 asked for and reuses `ModalBase.css` chrome rather than inventing bespoke styling. Scope audit found six actions requested, six delivered, no creep.

## Recommendation

Findings **1** and **3** are cheap to close now while context is hot (a ~3-line `cancelBulkIntent` restore plus a test; a one-sentence doc addition). Findings **2, 4, 5, 6, 8** suit a single follow-up bead — 2 and 4 share a resolution. Nothing here is a merge blocker.

---

MotWakorb:
# Team Review — Round 4: validation of `abdcd20e`, `49f3aa5f`, `f310cca8`

Ten-persona pass at tip `f310cca8`, calibrated to **home-lab** tier. CI verified green on the exact tip SHA (not an earlier commit) by two personas independently.

**Verdict: no persona blocks the merge.** Every prior-round finding is closed — Code Reviewer (4), Project Engineer (1), QA (1), Tech Writer (1), UX (1) — verified against source rather than commit messages. Gates independently re-run: **106 frontend + 40 backend passing**, `tsc` and ESLint clean.

## Correction to the round-3 comment above

That comment listed under "Verified clean": *"Ops: Frontend-assets-only deploy path unchanged; rollback is a straight revert."* That was true when written and was **invalidated eleven minutes later** by `abdcd20e`. This PR now adds a backend endpoint, a new auth surface, and a new DB query. Rounds 1–3 could not have caught what this round found, because none of it existed yet.

## What changed

`abdcd20e` deleted the ~90-line client-side two-pass fingerprint protocol and replaced it with `GET /api/channel-merges/snapshot` — admin-gated, one ordered `SELECT`, `limit(MAX+1)` peek-ahead, 409 fail-closed at 20,000 rows. `49f3aa5f` bounded DOM rendering to 200 rows. `f310cca8` added Previous/Next navigation over that window.

Worth stating plainly: `f310cca8` is undersold by its own title. Before it, `49f3aa5f`'s windowing made any failed row past DOM position 200 **unreachable for retry**. It fixes a regression introduced two commits earlier in the same PR — caught before merge rather than after.

## Prior findings — disposition

| Prior finding | Status | Evidence |
|---|---|---|
| `cancelBulkIntent` didn't restore paginated rows | **CLOSED** | `preConfirmViewRef` guard at `:432-445`. The "silent no-op" failure mode was specifically hunted: `preConfirmViewRef` is populated only for `scope==='all'` immediately before `setRows(targets)` in the same synchronous block, and all row actions are disabled while `bulkIntent !== null` — no reachable path where the id-join comparison fails while the modal is open. Backed by a 51-row regression test |
| `loadRows` single-pass walk on Refresh | **CLOSED** | Not relocated — the bug class is eliminated with the client protocol. New test asserts snapshot called once, paginated list not at all |
| Bulk-all always ≥2 traversals | **CLOSED** | One HTTP call, one DB query |
| >200-row bulk boundary untested | **CLOSED in new form** | The client pagination boundary is gone, but a new 200-row *DOM* boundary replaced it and is directly tested: failures at ids ≤201 plus id 250, asserting 260/260 processed and id 250 reachable, checked, and retryable |
| "Select all" scope undocumented | **CLOSED** | User guide now states it spans the complete paginated queue |
| No loading affordance (F11) | **CLOSED** | `snapshotAction` state swaps button text to "Loading all…" |
| File 802 lines | **PARTIAL** | 802 → 801. Protocol out, windowed nav in. Good trade; hook extraction still a candidate |

## New findings

**1. [Warn] `/snapshot` has no `group_id` filter, but the list endpoint and `api.ts` both do.** `list_pending_merges` accepts `group_id: Optional[int]`; `snapshot_pending_merges` hardcodes `status == "pending"` with no group parameter. Ordering matches correctly (`created_at DESC, id DESC` on both), but filtering does not. Verified this is **latent, not live** — `PendingMergesPage.tsx` calls `getPendingMergesSnapshot()` with no arguments today. But the moment a group filter reaches this UI, "Select all" and "Merge all" will silently target the entire queue instead of the filtered view, on an operation whose entire safety story is "the count you confirm is the set we process." One parameter closes it; an invariant maintained by comment lasts one refactor.

**2. [Warn] The 20,000-row cap and the 1,000-channel enrichment ceiling don't compose.** Flagged independently by four personas. `_resolve_candidate_channels` is correctly *not* N+1 — it issues exactly one `get_channels(page=1, page_size=1000)` call regardless of input size, with a 30s timeout, failing soft to `{}`. But at snapshot sizes above 1,000 candidates, the remaining rows render with null name/number/group — i.e. the "channel no longer exists" fallback — in the confirmation dialog for an irreversible bulk merge. The two ceilings should be within an order of magnitude of each other. Suggested: lower `MAX_SNAPSHOT_ROWS` toward ~2,000 (still 10× ADR-008's documented "dozens to hundreds") rather than raising the enrichment page size.

**3. [Warn] `class TestPendingMergesSnapshot` has 8 tests; only 4 call `/snapshot`.** The other four — `test_candidate_deleted_since_queuing_returns_null_fields`, `test_dispatcharr_fetch_failure_degrades_gracefully`, `test_empty_queue_does_not_call_dispatcharr`, and `test_batch_resolution_makes_one_call_regardless_of_row_count` — call `GET /api/channel-merges` (the list endpoint; the last one at `?page_size=50`). Root cause: the diff inserted the class header above pre-existing tests without a closing class boundary, so they were structurally absorbed rather than deliberately written.

Net effect: the new endpoint has no test for Dispatcharr-outage degradation, no test that an empty queue skips the Dispatcharr call, and **no test proving the batch-resolution contract holds for `/snapshot` specifically** — which is exactly the path finding #2 concerns. The shared `_resolve_candidate_channels` helper makes divergence unlikely today, but a future refactor that splits the two call sites would ship green. Fix is a class-boundary correction plus 3–4 short tests.

**4. [Low] `loadRows` now routes to the admin-gated snapshot whenever any row is selected.** Previously every refresh used the auth-gated list. A non-admin with a checkbox ticked now gets a 403 surfaced as `loadError` on plain Refresh. No bypass and no escalation — the actions those rows lead to were already admin-gated — and zero practical impact at home-lab with a single admin operator. Flagged because it's a silent privilege coupling introduced into a read path.

Related, non-blocking: `docs/api.md` says the route "requires admin authorization because it exposes the complete destructive action target set." Defensible as intent-signaling, and the stricter gate is the safe direction — but it should not calcify into a belief that this data is admin-only. Every row is obtainable by a non-admin paging the list endpoint at `page_size=200`.

**5. [Low] Sync session inside an `async def` handler.** `snapshot_pending_merges` is `async def` but uses a synchronous SQLAlchemy session, so the query plus up to 20,000 Pydantic constructions plus serialization run on the event loop without yielding — the same loop `TaskEngine` and `StreamProber` share. The pattern is pre-existing across every router in this codebase; what's new is a 100× amplification of the row ceiling (200 → 20,000). Bounded, admin-only, operator-triggered, and only reachable by deliberate click — not a merge concern at this tier, but worth naming since bulk multiplies latent severity.

**6. [Should-fix] Deploy order and rollback coupling are undocumented.** This PR now spans two deploy targets. Backend must land first:

```
docker cp backend/routers/. ecm-ecm-1:/app/routers/
docker restart ecm-ecm-1
scripts/deploy-frontend.sh
```

Reversed order gives a window where the new frontend calls a route that doesn't exist — every queue-wide bulk action hard-404s. Rollback has the same coupling in reverse: reverting the backend alone while the new frontend is deployed breaks the same four actions. Production GHCR deploys are atomic single-image pulls, so this is specific to the manual `docker cp` dev loop that CLAUDE.md documents as two separable steps. The PR body's Deployment section says only "No infrastructure or database migration required" — true, but silent on the two-target requirement. One line fixes it.

**7. [Nit] Docs.** ADR-008 was materially amended in four places (§D10 ×3, threat-model carve-out, Consequences, References) with no Revision History row, though ADR-009 sets a clear house precedent for recording exactly that. `docs/architecture.md` still says "Four endpoints under `/api/channel-merges/*`" — now five. Separately, ADR-008 §D1 is the canonical endpoint enumeration and doesn't name `/snapshot`; §D1 line 85 also argues *against* path-segment routes, so a couple of sentences reconciling why that reasoning doesn't apply here (different envelope, higher auth gate, distinct failure mode — not a lifecycle state) would close the loop.

**8. [Nit] `/snapshot` has no logging of its own**, unlike every sibling in the router, including on the 409 cap path. A capped or failed snapshot is currently indistinguishable from a healthy one in logs — diagnosable only from the HTTP response the frontend received.

**9. [Backlog] Stopped-mid-sweep failure discovery.** On a *completed* sweep, successes are filtered out as they land, so `rows` collapses to just the failures and the nav often never appears — a genuinely nice property. But if an operator hits Stop after 500 of 20,000 with 12 failures, `rows` holds those 12 mixed with ~19,500 never-attempted rows, all re-selected. The progress banner reports the split correctly in text, but there's no "jump to failures" affordance, so finding them means paging. Reachable, not efficiently findable.

**10. [Nit] Vestigial test name.** `'fails closed when the snapshot endpoint rejects an unstable queue'` mocks a rejection string that appears nowhere else in the tree. The single ordered SELECT is structurally immune to the "queue kept changing" race that name describes — it's leftover narrative from the deleted protocol. The test exercises valid behavior; only the name is fiction.

**11. [Nit]** The `runConfirmedBulkAction` indentation glitch flagged last round is still unfixed across all three commits.

## Verified clean

- **Auth wiring is real, not a no-op.** The FastAPI underscore-param footgun was checked explicitly and is absent — `test_non_admin_is_forbidden` overrides the dependency and gets a 403, which only happens if it's in the resolved graph.
- **No route shadowing.** This router has no `GET /{merge_id}` at all; the only parameterized routes are `POST /{merge_id}/accept` and `/dismiss`. `/snapshot` registers at 812, before `GET ""` at 858.
- **No new data exposure.** `PendingMergesSnapshotResponse` reuses `PendingMergeRecord` verbatim — field-for-field identical to what the list endpoint already returns.
- **Index support confirmed empirically.** `EXPLAIN QUERY PLAN` against a copy of the live `journal.db` (original untouched): `SEARCH pending_merges USING INDEX idx_pending_merges_status_created`, no temp B-tree. The `id DESC` tiebreak was verified against 20 synthetic same-timestamp rows.
- **Ordering parity** between snapshot and list is byte-identical, so the select-all set and the visible page cannot disagree.
- **409 never corrupts state** — it fires before `setBulkIntent`, so a failed snapshot leaves the page in its prior state with no partial render and no dead-end click-through.
- **`docs/api.md` accuracy** verified field-for-field against the implementation, including the 409 detail text verbatim. It also fixed a pre-existing auth-table error it wasn't obligated to touch.
- **Versions clean** at `0.17.6-0145` across `main.py`, `backup.py`, `package.json`, CHANGELOG.

## Done well

Trading a hard-to-verify client-side "two reads agree" protocol for a single transactionally-coherent server read is the right call — it's simpler, correct by construction, and closed two real bugs as a side effect rather than requiring bespoke patches to the old protocol. The `limit(N+1)` peek-ahead detects overflow inside the same read, so the cap check can't itself tear. The `id DESC` tiebreak matters and was not obvious. Docs were updated in lockstep — no stale description of the removed protocol survives anywhere. Test quality remains high and non-tautological; the 200-row DOM boundary test in particular would fail against a plausible broken implementation.

One methodological note for the record: re-running the decisive overlay experiment (current tests against the pre-fix tree) yields 4 failed / 36 passed, but all four fail with a bare 404. That proves the tests catch "endpoint missing," not that they pin the ordering, tie-break, or cap logic. It is a weaker signal than the 17-of-29 result from round 3, and is reported as such rather than as equivalent evidence.

## Recommendation

Findings **1** and **6** are cheap and worth closing before merge — #1 is a latent bulk-destruction divergence that costs one parameter now and is easy to miss forever, #6 prevents a real deploy footgun on a PR that spent most of its life frontend-only. Findings **2, 3, 4, 5, 7–11** suit a single follow-up bead. Nothing here blocks.

---

Issue #11, opened 2026-01-21
[Feature]: Full watch history


Deekerman:
### What would you like this feature to do?

I would like to be able to pull a list of the watched history. Similar to the top watched, just a full list of everything.

### How critical do you think this is?

Medium - Nice to have, would use it regularly

Deekerman:
just an expanded version of what you currently have. Shows arent needed. Just the channels and watch time but a way for me to see everything not just the top 10

Issue #10, opened 2026-01-21
[Feature]: Ability to schedule probing of specific categories


Deekerman:
### What would you like this feature to do?

I was thinking about your probing this morning in the shower

how we could improve the probing

would it be possible to set multiple probe schedules that probe different things?

Like, say every morning i want to run my morning probe of everything

but then i want to run an afternoon probe or evening probe of some of my event channels to make sure they have the right event at the top

so being able to select different probe times of different categories would be some excellent probing improvements

### How would you suggest implementing?

Carefully

### How critical do you think this is?

Medium - Nice to have, would use it regularly

PR #716, opened 2026-07-21
[enhancedchannelmanager-6lhmb] Add date windows to Channel Pipeline rules

### LINKED ISSUE #646: [Feature]: Auto Sync and Auto Creation by date

Allow Auto Sync and Auto Create rules to be enabled by date.  Example:  I want "football" auto sync rule to assign profile "Primary" during season (start date) and when the season ends (end date) I can have it un-assign the "Primary" profile. 

_No response_

### Additional Context

_No response_

MotWakorb:
## Summary

- Add optional inclusive UTC `active_from` and `active_until` windows to standard and Event Sync Channel Pipeline rules.
- Preserve and validate windows across CRUD, bulk updates, duplication, YAML import/export, unattended refresh, and manual execution.
- Add Alembic migration 0038, editor controls, regression coverage, changelog entry, and build 0.17.6-0147.

Closes #646.
Bead: `enhancedchannelmanager-6lhmb`
Follow-up for explicit profile removal actions: `enhancedchannelmanager-o8bzu`.

## Test plan

- [x] Backend affected sweep: 321 passed.
- [x] Channel Pipeline router module: 177 passed.
- [x] Earlier full backend baseline: 7,605 passed, 13 skipped.
- [x] Frontend full suite: 149 files, 2,126 tests passed.
- [x] TypeScript, ESLint, and production build passed.
- [x] Alembic fresh upgrade and 0038 downgrade/re-upgrade passed.
- [x] Version consistency and advancement checks passed.
- [x] Standard-rule editor rendered and visually verified at 1440×1000.
- [x] Independent QA and code review approved final tip `f3c65c99`.

## Checklist

- [x] Bead linked in commit messages.
- [x] Tests updated or added.
- [x] `CHANGELOG.md` updated under `[Unreleased]`.
- [x] User-facing editor behavior is covered by inline labels, validation, and tests; no separate documentation page is required for this focused control.

## Behavior note

Date expiry prevents future execution but does not automatically undo prior rule actions. Cleanup is modeled as a separate explicit rule/action, per product decision.


MotWakorb:
# Team Review — PR #716: Add date windows to Channel Pipeline rules

**Depth:** Quick · **Effective tier:** home-lab (`ecm-core`, `data-store`, `dbas-backup`) · **Reviewers:** 10-persona panel (security, architecture, PM, engineering, UX, code review, database, SRE, QA, technical writing)

## Summary

Well-built feature with genuinely strong test coverage on its core logic — inclusive-boundary math, cross-field validation atomicity, and YAML round-trips are all tested at the right layer, and the `Date`-not-`DateTime` design cleanly sidesteps timezone bugs. **One HIGH data-integrity defect should block merge:** a full DBAS backup **restore silently drops** `active_from`/`active_until`, reverting deliberately time-boxed rules to always-on. Everything else is polish or follow-up.

## 🔴 Must-fix (blocks merge)

### #1 — DBAS restore silently drops `active_from`/`active_until`
Found independently by **Architecture, Security, and QA**, and confirmed by direct code read.

`_restore_auto_creation_rules` (`backend/routers/backup.py:2508`) reconstructs `ChannelPipelineRule(...)` passing through every other historically-added column (`match_scope_target_group`, `match_scope_group_id`, `allow_manual_channel_merge`, `fold_match_key`, `event_sync_config`) but **omits `active_from`/`active_until`**. `to_dict()` *does* export both (`models.py:2169-2170`), so the backup artifact looks complete — the data is silently discarded on the way back in. Every restored rule reverts to an open-ended (always-active) window; a deliberately time-boxed seasonal rule comes back firing year-round, with no error or warning.

**Direct precedent:** `event_sync_config` fell into this exact trap before, was treated as a must-fix bug, and has a regression test (`test_event_sync_persistence_roundtrip.py`). The new fields hit the identical gap and **no restore test covers them** (verified across `backend/tests/`).

**Fix:** add `active_from=item.get("active_from")` / `active_until=item.get("active_until")` to the restore constructor, plus a round-trip regression test. Recommend a structural test diffing `ChannelPipelineRule.__table__.columns` against the restore constructor kwargs so this class of bug can't recur silently on the next new column.

## 🟡 Should-fix — recommend in this PR (cheap, closes real traps)

- **#2 Misleading no-op message** — manual "Run Now" on an out-of-window rule returns `"No enabled rules to process"`, pointing operators at the wrong control (the `enabled` flag). A PR test currently *locks this message in*. This defeats the feature's own debuggability goal. _(SRE, Engineer, QA, UX, Security)_
- **#3 Wrong unattended-refresh log** — when a `run_on_refresh` rule exists but is date-gated, the INFO log says no such rule exists and suggests enabling `run_on_refresh` — the wrong fix. Breaks the file's own event_sync-exclusion logging pattern. _(SRE)_
- **#5 "Expiry doesn't undo prior actions" is silent in the UI** — only stated in the PR body/CHANGELOG. One-line hint addition to both editors closes a plausible wrong mental model ("when this ends, it reverts"). _(UX, Technical Writer)_
- **#7 UTC inclusive-calendar-day cutoff** — labeled "UTC" but the local effective cutoff can be up to a day earlier/later; one-line hint clarification. _(UX)_

## 🟢 Follow-up beads (don't block)

- **#4 No list-view signal** that an enabled rule is currently outside its window — only visible by opening each editor. _(Architecture, Technical Writer, UX)_
- **#6 EventSyncRuleEditor UX** — validation-error focus routes to the wrong wizard step (dates are in Step 1, focus jumps to Step 2); date window absent from the Review step; date inputs lack `aria-invalid`. _(UX)_
- **#8 `docs/api.md`** bulk-update scalar-field list omits the two new fields — contract now inaccurate for scripted/MCP callers (cheap, could ride along). _(Technical Writer)_
- **#9 Duplicated window SQL predicate** in `_load_rules` and `tasks/channel_pipeline.py` — drift risk; extract a shared `active_window_clause(today)` helper. _(Architecture, Engineer, QA)_
- **#10 Test gaps** — frontend inverted-range validation untested; YAML `datetime`-rejection branch untested; no same-day (`active_from == active_until == today`) or second-call-site boundary test. _(Code Reviewer, UX, QA, Engineer)_
- **#11 No DB `CHECK` constraint** for `active_until >= active_from` — deviates from this migration series' convention without the customary docstring note. Add the constraint via `batch_alter_table`, or one docstring sentence recording the app-only-validation choice. _(Database)_
- **Scope (PM):** #646 is being closed on the assign-only half; the un-assign half is deferred to bead `o8bzu` (P3, unscheduled). Confirm closing #646 on the gate-only half is acceptable, or keep it open until `o8bzu` lands.

## ✅ Verified clean / done well

- **Authorization** — `RequireAdminIfEnabled` on all mutating endpoints; no new authz gap.
- **Injection/DoS** — `yaml.safe_load`, typed date coercion (422 before handler), exception-safe per-row YAML parsing that continues rather than aborting the batch. The `isinstance(value, datetime)` check correctly ordered before `isinstance(value, date)`.
- **Write atomicity** — `_validate_persisted_active_window` runs pre-commit on both single-PUT and bulk paths; a rejected window can't leave a partial write (explicitly tested).
- **Migration 0038** — additive, nullable, idempotent (`_columns()` guard), cleanly reversible, wired into the alembic smoke test. Correct plain `op.add_column`/`drop_column` for SQLite 3.35+.
- **Design** — `Date` + `datetime.utcnow().date()` avoids the naive/aware datetime trap entirely; standard and Event Sync rules share one column pair through one `_load_rules()` choke point; enforcement verified consistent across manual, selected-rule, unattended-refresh, and event-sync auto-run paths.
- **Test quality** — real inclusive-boundary coverage (`test_load_rules_only_inside_inclusive_active_window`), atomic bulk-rejection (`test_clears_windows_and_rejects_atomically`), and thorough YAML edge cases. **Code Reviewer found no Block/Warn issues** in the code reviewed.

---
*Team recommendation: fix #1 in this PR before merge (matches how the identical `event_sync_config` bug was handled); land #2/#3/#5/#7 here too as they are near-zero-cost and close the feature's own stated gaps; file the rest as prioritized beads for grooming.*

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — 10-persona team-review, calibrated to home-lab tier.</sub>

MotWakorb:
# Team Delta Re-Review — PR #716, commit `6f70deb3` "Address date window review feedback"

**Depth:** Quick · **Effective tier:** home-lab · **Panel:** 7 domain reviewers (security, engineering, SRE, QA, code review, database, UX) — proportional to the 81-line delta over `f3c65c99`. Two reviewers (engineering, code review) **ran the changed tests directly** (115 backend + 161 frontend passing, no regressions) rather than trusting the self-report.

## Verdict

**The update resolves the HIGH blocker and every should-fix item from the first review, all test-backed.** No new blocking findings. One cheap log-spam regression is worth fixing before merge; everything else deferred is already tracked in beads.

## Prior findings — status

| # | Finding (first review) | Status |
|---|---|---|
| **#1** | **HIGH — DBAS restore silently drops `active_from`/`active_until`** | ✅ **FIXED** — restore constructor now round-trips both fields (`backup.py:2513`); real `to_dict()→restore→re-query` regression test (`test_backup_restore_round_trip_preserves_active_window`). Confirmed by Security, QA, Engineering, Code Review, Database. |
| #2 | Misleading "No enabled rules to process" on out-of-window run | ✅ **FIXED** at the task layer (specific "No run_on_refresh rules are active in their UTC date windows" + test); engine layer got a lighter "No **active** enabled rules to process" tweak — acceptable, that's the manual path with the rule list in front of the operator. |
| #3 | Wrong unattended-refresh log suggested enabling `run_on_refresh` | ✅ **FIXED** — new date-gated detection block; `test_only_date_gated_rules_logs_window_guidance` asserts the new copy present **and** the old misleading copy absent. |
| #5 | "Expiry doesn't undo prior actions" silent in the UI | ✅ **FIXED** — hint added to both editors + render assertions. |
| #7 | UTC inclusive-calendar-day cutoff | 🟡 **PARTIAL, acceptable at tier** — hint now reads "Dates are inclusive UTC calendar days"; the local-time arithmetic (a UTC day boundary can land mid-afternoon local) is still left to the operator. Fine at home-lab. |
| #4, #6, #8, #9, #10, #11 | List badge · EventSync aria/focus + Review-step · api.md field list · dup SQL predicate · residual test gaps · DB CHECK constraint | ⏳ Correctly **deferred to open follow-up beads** (`ftgqb`, `e0mbg`, `ybzmh`, `eb9ai`, `wsug0`, `8rp7e`, `gqkul`) — tracked, not dropped. |

## New findings on the delta itself (none blocking at home-lab)

1. **🟡 Log-throttle regression (SRE)** — the new `date_gated` `logger.info(...)` branch in `tasks/channel_pipeline.py` is **not** wrapped in `should_log()` like its sibling branch three lines below. Because the refresh watermark isn't advanced when nothing runs, a multi-week dormant window emits unthrottled INFO **every ~60s indefinitely**, which will bury genuine WARNING/ERROR signal in `docker logs`. Contradicts the file's own "Surface at INFO (throttled per task)" comment and the `log_throttle.py` rationale. *The one genuinely new regression — cheap ~1-line fix (`should_log("date_gated_refresh:%s" % self.task_id)`), recommended before merge.*
2. **🟢 Restore date-parse error granularity (Security / Code Review / QA / Database)** — bare `date.fromisoformat(item["active_from"])`, unlike the YAML import path's per-row `_parse_yaml_active_date` helper, so one malformed date aborts the whole `auto_creation_rules` section restore. Fully **contained** — the function's `except Exception: session.rollback(); raise` means no data loss, it's human-admin-only (`RequireHumanAdminIfEnabled`), and only the exception class name reaches the client (no info leak). A UX-granularity gap, not a vulnerability. Follow-up bead (`gqkul`).
3. **🟢 Eager/duplicated diagnostic queries (Engineering / Database / Code Review)** — the two date-gated queries (a `.count()` and a full `.all()`) run unconditionally every scheduler tick even on the happy path where they're never read; both duplicate the filter predicates three lines above. Correct, but move inside `if not rule_ids:`. Fits `eb9ai`.
4. **Nits** — prefer an `r.id` set over object-identity `not in event_sync_candidates` (correct today via the shared-session identity map, fragile if ever split across sessions); add an event-sync-only date-gated branch test (the `date_gated_event_sync` path is implemented differently from the standard path and is currently unexercised) — fits `wsug0`.

## Verified clean / correct (independent read)

- Restore round-trip is type-correct (`str → date → sa.Date`, no TZ ambiguity — both sides are naive calendar dates).
- `date_gated_standard`'s `count() > len(rules_to_run)` compares identical populations minus the window filter — no off-by-one, no event_sync leakage.
- Identity-based `in` is safe within the single shared session; `date_gated` cannot false-positive on zero-rows (`0 > 0` / `any([])` → `False`) or on a falsy `auto_run`.
- All CI checks green. PR remains **draft**; version stayed `0.17.6-0147` (fix rode the existing build; version-consistency CI passes).

## Dissent

None. Every persona that reviewed the restore fix agrees it's complete and correctly tested; no severity conflicts.

---

**Bottom line:** the HIGH blocker is cleared with a genuine regression test, all should-fix items landed and are test-backed, and the deferred items are tracked as open beads rather than silently dropped. Recommended before merge: the ~1-line log-throttle fix (new finding #1). Otherwise mergeable once taken out of draft.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code) — 7-persona delta re-review, calibrated to home-lab tier.</sub>

Issue #720, opened 2026-07-21
[Bug]: Channel Profiles not adding to only selected Channel profiles


langtd19:
### ECM Version

v0.17.6-0149

### Priority

Priority 2: ECM is severely degraded

### What is the bug?

When creating a new standard or event sync rule it does not appear to be honoring the channel profile selection and everytime run either it adds the channels to all the channel profiles.  

### Steps to Reproduce

create or run existing rule and define the channel profile.

langtd19:
Maybe you already discovered or related but should mention that if try to edit Channel profiles in the Channel manger it throws an error 

<img width="333" height="84" alt="Image" src="https://github.com/user-attachments/assets/85d4d0d3-4699-4704-b91e-7a4116971b2a" />

MotWakorb:
Thanks for the extra detail and the screenshot — that helps.

Two separate things are going on in this issue:

**1. Channels being added to all profiles (the original report).** Confirmed and fixed. The rule action was only *enabling* the selected profile(s) and never *disabling* the others, so channels stayed in every profile. The fix ships in **v0.17.6-0150**, with some additional hardening (correct behavior when a profile update partially fails) landing in the next build.

**2. The "Profiles" error when editing channel profiles.** We haven't been able to reproduce this on our end running your exact version (0.17.6-0149) — every profile edit succeeds here — so it looks specific to your instance (most likely your Dispatcharr rejecting one of the profile calls). The reason the toast is unhelpful is a bug on *our* side: ECM currently collapses these errors into a generic "Internal server error" instead of showing what Dispatcharr actually returned. We're fixing that so the real message surfaces in the toast.

To pin down your specific failure in the meantime, could you grab your ECM container logs right after triggering the error and paste the relevant lines? Something like:

```
docker logs ecm-ecm-1 --tail 200
```

and look for a line beginning `[PROFILES] Failed to …` — it'll contain the true status code and message from Dispatcharr, which will tell us exactly what's being rejected. Thanks!

langtd19:
I am finding no Failed to or anything related to failed

INFO:     192.168.90.23:47190 - "PATCH /api/channel-profiles/19/channels/249565 HTTP/1.1" 200 OK
[...]

PR #721, opened 2026-07-21
v0.17.6-0150: Fix channel-profile assignment in pipeline rules (GH #720 Part 1)

MotWakorb:
## Summary

Channel Pipeline rules with an `assign_channel_profile` action (standard **and** event-sync rules) added created channels to **every** channel profile instead of only the selected one(s).

**Root cause:** the action applied the profile selection ENABLE-ONLY. But Dispatcharr auto-joins every newly-created channel to all channel profiles by default, so honoring a selection is *subtractive* — the selected profiles must be enabled AND the channel disabled in every other known profile. Enable-only was a no-op, so the channel stayed in all profiles.

## The fix (rule-action path)

- **`backend/channel_pipeline_executor.py`** — new shared `_apply_exclusive_profile_membership(channel_id, selected_ids)` helper makes a channel a member of exactly the selected profiles (enable-in / disable-out over the de-duplicated union of the known universe and the selection). `_execute_assign_channel_profile` and `_assign_default_profiles` both refactored onto it. Per-profile failures are logged and skipped (one bad profile can't abort the rest). When the profile universe is empty/unavailable, it degrades to enable-selected-only — never worse than before.
- **`backend/channel_pipeline_engine.py`** — lifted the profile-fetch gate: the full channel-profile universe is now fetched whenever any rule (standard or event-sync) uses an `assign_channel_profile` action, not only when a global `default_channel_profile_ids` is configured. Previously a per-rule selection with no global default got an empty universe and could not disable anything (the load-bearing half of the bug).

## Tests

- `TestAssignChannelProfileAction` + `TestAssignDefaultProfiles` (executor) — subtractive membership, multi-select, empty-universe degradation, dry-run makes no calls, guard failures, one-failing-profile-non-fatal.
- `TestProfileFetchGate` (engine) — asserts `get_channel_profiles` IS fetched for an `assign_channel_profile` rule with no global default, and is NOT fetched otherwise. Verified to fail against the pre-fix gate.
- Full backend suite green (7639 tests); frontend build passes; version consistency passes across all 3 touchpoints (0.17.6-0150).

## Scope

**Part 1 of 2 for #720** (the rule-action path). The M3U Auto-Sync-modal reconciliation path is Part 2 and remains open — this PR intentionally does not close the issue.

Bead: `enhancedchannelmanager-y3m6o`


MotWakorb:
## Team review — PR #721

**Depth:** Quick  
**Effective tier:** small-team (strictest-wins across `ecm-core`/docs at home-lab and `ci-release` at small-team)  
**Participants:** Security, IT Architecture, Project Management, Project Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

### Summary

The happy-path fix is well targeted and strongly unit-tested, but the team found one must-fix follow-up: exclusive membership can fail partially or lose the profile universe entirely while the action is still recorded and displayed as successful. Eight personas independently identified this. Because PR #721 is already merged, this is a P1 follow-up rather than a retroactive merge block.

### Critical / must-fix follow-up

1. **Incomplete reconciliation is reported as success** — **High / Block**
   - `_apply_exclusive_profile_membership` catches per-profile update failures and returns only success counts (`backend/channel_pipeline_executor.py:3437-3451`).
   - `_execute_assign_channel_profile` then unconditionally returns `success=True, modified=True` (`:2759-2773`).
   - The new failure test codifies this result (`backend/tests/unit/test_channel_pipeline_executor.py:4422+`).
   - A failed disable can therefore leave a channel in an unintended profile—the original user-visible invariant violation—while execution history/UI show success.
   - **Action:** keep attempting remaining profiles, but collect failed profile IDs and return failed/explicitly-partial status. Add partial, total, and selected-profile failure regressions.

2. **Profile-universe fetch failure silently restores enable-only behavior** — **High / Block**
   - The engine catches `get_channel_profiles()` failure and substitutes an empty universe (`backend/channel_pipeline_engine.py:1939-1945`).
   - The helper then enables selected profiles without disabling unknown memberships (`backend/channel_pipeline_executor.py:3419-3426`) and still reports success.
   - With Dispatcharr auto-joining new channels to all profiles, this recreates GH #720 during a transient read failure.
   - **Action:** distinguish “loaded an actually empty universe” from “failed to load universe”; exclusive assignment must fail/defer or visibly degrade when the universe is unavailable.

These are two manifestations of the same contract failure and are tracked together in **`enhancedchannelmanager-y3m6o.1` (P1 bug)**.

### Important / should-fix

3. **Dry-run does not preview removals** — **High UX / Warn**
   - Dry-run says only “Would assign N channel profile(s)” (`backend/channel_pipeline_executor.py:2743-2751`), while live execution disables every known unselected profile.
   - The preview can therefore look additive even though execution is subtractive/destructive.
   - Track in **`enhancedchannelmanager-y3m6o.2` (P2 task)**.

4. **Rule-editor copy describes additive assignment, not exclusive replacement** — **Medium / Warn**
   - `ActionEditor.tsx` uses “Set Channel Profile” / “Assign a channel profile” without disclosing that unselected memberships are removed.
   - Update copy to state: selected profiles are enabled; all others are disabled.
   - Also tracked in **`enhancedchannelmanager-y3m6o.2`**.

5. **Missing integration and event-sync regression coverage** — **Medium / Warn**
   - Engine tests prove the fetch gate is called, while executor tests inject `all_profile_ids`; no behavioral test joins both halves.
   - The production gate includes `event_sync_rules`, but the new gate tests cover only standard rules despite the PR explicitly claiming both paths.
   - Add a pipeline-level standard-rule test and an event-sync-only regression as acceptance coverage for `y3m6o.1`.

6. **Sequential O(channels × profiles) remote PATCHes** — **Low / measured risk**
   - Acceptable at current tier and expected profile counts.
   - Measure before redesign; use a bulk/reconciliation path only if run latency or partial-application windows become material.

7. **Part 2 needs independently traceable delivery state** — **P2**
   - PR/body/changelog correctly identify the M3U Auto-Sync modal path as still open, but the original bead mixed shipped Part 1 and unshipped Part 2.
   - Created **`enhancedchannelmanager-y3m6o.3` (P2 bug)** as a child for Part 2. GH #720 should remain open until it ships.

### Team agreements and resolved tension

- **Agreement:** best-effort continuation is appropriate; one failed profile should not prevent attempts against the rest.
- **Agreement:** best-effort must not be labeled full success when exclusive membership is unproven.
- **Severity variance:** Code Review rated the behavior Warn at this tier; Architecture, Engineering, UX, Database, SRE, and QA rated it Block/High. The unified assessment treats it as a P1 follow-up because it violates the central invariant and hides recurrence.
- No Critical security vulnerability was found. Security rated the integrity/segmentation impact Medium under the current private-LAN/small-team calibration.

### Positive observations

- The shared helper removes semantic drift between rule-selected and default-profile assignment.
- The happy-path test matrix is strong: single/multiple selections, complement disabling, stale selected IDs, dry-run, validation/guard paths, and default-profile reuse.
- CI was green across backend/frontend/MCP tests, fake-test guard, visual regression, Semgrep, CodeQL, version consistency, performance checks, and AMD64/ARM64 builds.
- PR #721 clearly scopes itself as Part 1 of 2 and correctly leaves GH #720 open.
- No schema migration was needed; version touchpoints are consistent at 0.17.6-0150.

### Beads created

- `enhancedchannelmanager-y3m6o.1` — P1 bug: truthful partial/failure status and unavailable-universe handling
- `enhancedchannelmanager-y3m6o.2` — P2 task: preview parity and explicit exclusive-membership copy
- `enhancedchannelmanager-y3m6o.3` — P2 bug: M3U Auto-Sync Part 2 reconciliation

All review activity was read-only; no code changes were made.

PR #723, opened 2026-07-21
v0.17.6-0152: Truthful run-level status + event-sync for channel-profile assignment (GH #720 y3m6o.1)

MotWakorb:
## Summary

Completes the P1 team-review findings on the merged build **0151** and delivers build **0.17.6-0152**. Build 0151 made the `assign_channel_profile` *action* truthful, but a run with a failed action still finalized as green `completed`, so the operator-visible outcome still hid the failure. This PR makes the whole **run-level status** honest across every surface.

### What changed
1. **Run-level `completed_with_errors` status.** The engine aggregates failed actions at finalization; a run with any failed action finalizes as the new terminal status `completed_with_errors` (distinct from green `completed` — some channels may have succeeded, so it is deliberately not a blanket `failed`), persists an error count + short summary on the execution record, and the top-level API result reports `success: false`.
   - **Manual surface:** the executions UI renders `completed_with_errors` as an amber/warning badge + toast ("Completed with Errors" label + failed-action summary), and keeps the rollback/undo affordances.
   - **Unattended surface:** the post-M3U-refresh auto-fire run no longer notifies green on a failed-action run — it reports *completed with warnings* (a single "Task Completed with Warnings" notification that fires external alerts, not a green success and not a hard "failed"), while the persisted pipeline `execution.status` stays the honest `completed_with_errors`. A defensive guard ensures a `completed_with_errors` run always routes to the warning branch regardless of the count.
2. **Dry-run universe validation.** A dry run whose channel-profile universe is unavailable returns an explicit blocking outcome instead of a rosy "Would assign N…".
3. **`modified` / rollback honesty.** A total profile-write failure counts `modified=False` (no phantom update count or rollback entity; partial success still counts). `assign_channel_profile` carries no reversible previous-state, so it is marked non-rollbackable and is no longer recorded as a rollback entity that claims to reverse a change nothing can undo.
4. **Event Sync implementation.** A configured `assign_channel_profile` action on an Event Sync rule now actually executes on the real `execute_event_sync_rule` attach path (newly-attached + promoted channels — the exact "new channel joined to all profiles" case #720 describes); the 0151 coverage exercised only a synthetic no-streams path.

### Testing
- Backend suite: **7671 passed, 13 skipped**.
- Frontend: **2146 passed**, `tsc --noEmit` clean, build clean.
- Version consistency: 3/3 at `0.17.6-0152`.
[...]

PR #724, opened 2026-07-22
GH #720 Part B: reconcile M3U Auto-Sync channel-profile membership (v0.17.6-0153)

MotWakorb:
## GH #720 Part B — M3U Auto-Sync channel-profile reconciliation (bead y3m6o.3)

Closes the still-open half of GH #720. Part A (PR #721) fixed the Channel Pipeline `assign_channel_profile` action; y3m6o.1 (PR #723) made run-status truthful. **This is Part B**: ECM's "Auto Sync" M3U groups are backed by Dispatcharr's native `auto_channel_sync`, which joins every new channel to **every** profile — so the profile selection an operator makes in the Auto-Sync settings modal (`custom_properties.channel_profile_ids`, already stored on dev) was silently ignored. Part B reads that selection back and enforces it.

### What it does
A new `backend/services/profile_reconcile.py` applies the stored selection **subtractively** and idempotently: for each profile in the universe, the group's channels are enabled if the profile is selected and disabled otherwise — converging on "member of exactly the selected profiles." Cost is **O(P) bulk PATCHes per group** (one `bulk_update_profile_channels` per profile), independent of channel count.

Invoked from three converging hooks (all best-effort — a reconcile failure never fails the caller):
- **M3U post-refresh completion poll** (`m3u.py`, both branches via a shared `_reconcile_profiles_after_refresh` helper)
- **Group-settings save router** (`m3u.py`) — instant apply on modal save (decision 3a)
- **M3U change monitor** (`m3u_change_monitor.py`) — the converging backbone that catches both ECM- and Dispatcharr-triggered syncs

### Locked PO decisions
- **1a — absent/empty selection is a NO-OP.** Never interpreted as "disable everywhere." A group with no selection is left entirely alone.
- **2b — pipeline-owned channels are EXCLUDED.** A channel whose profile membership was set by a Channel Pipeline `assign_channel_profile` rule carries a durable provenance marker (`custom_properties.ecm_profile_owner="pipeline"`, stamped at the single `_execute_assign_channel_profile` chokepoint that also covers the event-sync path); reconcile skips marked channels. Precedence: pipeline action > group selection > global default.
- **3a — instant apply on save.**

### Safety
- **Fully-stale selection guard**: if every selected profile has been deleted (authoritative universe fetch succeeds, intersection empty), reconcile issues **zero** bulk calls and returns a distinct `stale_selection` status rather than disabling channels in every real profile and stranding them in zero profiles.
- **Universe-fetch failure**: degrades to enable-selected-only (no blind disables), logged; self-heals on the next successful reconcile.
- **Effective-group resolution** follows Channel Group Overrides; the sweep dedupes by effective group id so a source+target both carrying selections resolve deterministically to the target's selection.

### Tests
`test_profile_reconcile.py` (22: no-op/1a semantics, subtractive convergence, fully-stale safety no-op, partial-stale ignore, universe-fetch-failure degradation, 2b exclusion with a real marker, override redirection, effective-group dedupe, pagination, idempotent rerun, O(P) accounting), `test_profile_reconcile_hooks.py` (save-hook invokes + never fails the save), and provenance-marker tests in `test_channel_pipeline_executor.py` (stamp on full-success non-dry-run, merge-over-existing, idempotent skip, dry-run skip, write-failure swallowed).

Full backend suite **7739 passed, 13 skipped**. Version `0.17.6-0153` (all three touchpoints). No closing keyword — GH #720 gets a status update and closes when this merges.


MotWakorb:
## Team Review — Quick (2026-07-22)

**Participants:** Security, Architecture, Project Management, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing  
**Effective deployment tier:** home-lab (`ecm-core`, `data-store`, `dev-tooling`, `docs`)

### Summary

The implementation has strong behavior-focused unit coverage and several good fail-safe decisions, but the team found four correctness/data-integrity blockers. The most serious failure path can remove channels from all working profiles when the selected-profile enable fails. The current ownership and account-scope models also create durable behavior that does not match the apparent operator contract.

### Critical / must fix

1. **Selected-profile failure can strand channels in zero profiles** — Security, Engineering, Database, SRE  
   `backend/services/profile_reconcile.py:274-291` catches each bulk failure and continues. If enabling a selected profile fails while later disables succeed, the group can be removed from every usable profile. Existing failure coverage (`backend/tests/services/test_profile_reconcile.py:327-338`) tests only the safer inverse: selected enable succeeds and a non-selected disable fails.  
   **Action:** enable all selected profiles first; if any selected enable fails, skip destructive disables, return a degraded/failed result, and add the inverse regression test.

2. **Pipeline ownership is permanent, with no control-handoff lifecycle** — Architecture, Code Review, QA, Technical Writing  
   `backend/channel_pipeline_executor.py:2980-2995,3814-3826` stamps a write-once `ecm_profile_owner=pipeline` marker; `backend/services/profile_reconcile.py:93-104,168-171` excludes it forever. Disabling/deleting/changing the rule does not return the channel to Auto-Sync control, despite the service contract describing an *enabled* pipeline rule.  
   **Action:** define and implement ownership release/revalidation (for example rule identity plus current-owner reconciliation, or an explicit operator reset), then test rule removal/control handoff. Document the lifecycle and recovery path.

3. **Account-scoped settings are collapsed into a global group selection** — Database  
   The save route is account-scoped (`backend/routers/m3u.py:1058,1101`), but `get_all_m3u_group_settings` collapses multiple account junction rows sharing a group ID (`backend/dispatcharr_client.py:636-668`). The sweep consumes that collapsed mapping (`backend/services/profile_reconcile.py:328-335`) while channel lookup operates on the effective/global group. Conflicting selections for the same group across accounts become last/selected-row-wins and can rewrite both accounts' channels.  
   **Action:** preserve `(m3u_account_id, channel_group_id)` and enumerate provider-scoped channels, or explicitly make selection global and reject/prevent conflicting account rows. Add a two-account conflicting-selection test.

4. **The advertised convergence backbone may never retry drift/failures** — Architecture, Engineering, SRE  
   `backend/tasks/m3u_change_monitor.py:193-203` runs reconciliation only when M3U content changes are detected. A failed save/refresh reconcile or external profile-membership drift can persist indefinitely when no later playlist content change occurs, contradicting the self-healing/converging comments.  
   **Action:** run the idempotent selected-group sweep on every scheduled monitor pass, or persist a retry-needed flag. Home-lab scale does not warrant a new queue/service.

### Important / should fix

5. **Failures are reported as successful reconciliation** — Architecture, Code Review, SRE, QA  
   Per-profile failures are swallowed, yet `profile_reconcile.py:300-308` returns `status="reconciled"`; the sweep counts that as success at `:354-361`. Return `partial_failure`/`degraded`, failed profile IDs/counts, and truthful aggregate totals. Keep best-effort caller isolation.

6. **Instant-save override resolution is order-dependent** — Architecture, Engineering  
   `backend/routers/m3u.py:1113-1124` reconciles edited IDs in request order and bypasses the effective-group target-preference dedupe used by the sweep (`profile_reconcile.py:337-352`). Source+target edits can address the same physical group with last-row-wins behavior. Reuse one dedupe/precedence helper and test both row orders.

7. **One group failure aborts later groups in the save hook** — Engineering  
   A single outer `try` surrounds the entire loop at `backend/routers/m3u.py:1111-1126`. Isolate exceptions per group and test first-group failure followed by second-group success.

8. **Two of three runtime hooks lack direct tests** — Engineering, PM, Code Review, QA  
   Save-hook coverage exists, but neither post-refresh completion branch nor the change-monitor trigger/cancellation path has focused wiring coverage. Add mock-invocation tests; service unit coverage cannot detect disconnected hooks.

9. **UI feedback and labels misstate behavior** — UX, PM  
   Save can return ordinary success while apply failed, and the UI reports “Group settings saved — M3U refresh started.” Also, “None selected” / “Clear All” deletes the selection, while backend semantics are a no-op that preserves existing memberships. Return/display an incomplete-apply warning, rename zero-selection behavior to “Stop managing profiles” (or equivalent), and note that pipeline-owned channels are excluded.

10. **Whole-value `custom_properties` replacement can lose concurrent updates** — Security, Database  
    `backend/channel_pipeline_executor.py:3810-3831` merges from a run-start cache and replaces the blob, explicitly accepting a staleness window. Fetch fresh state immediately before merging if supported; at minimum surface marker persistence failure so precedence is not falsely considered established.

11. **Documentation/API contract needs the new side effects and recovery semantics** — Technical Writing  
    `docs/api.md:462` does not say that group-settings PATCH now performs downstream bulk profile writes, that absent/empty selection is a no-op, or that reconcile failure does not fail the PATCH. The changelog also overstates “instantly” and gives no recovery instructions for stale selection, universe-fetch failure, or permanent pipeline ownership.

12. **Minor cleanup:** remove the unused `AsyncMock` import at `backend/tests/services/test_profile_reconcile.py:17` and resolve the existing CodeQL thread.

### Positive observations

- Strong tests for absent/empty selection, fully and partially stale IDs, pagination, override resolution/dedupe, pipeline exclusion, idempotency, and O(P) bulk-call count.
- The fully stale-selection and universe-fetch guards correctly avoid blind destructive disables.
- Per-group sweep isolation, bounded pagination, invalid-ID filtering, and clear service-level comments are appropriate for the home-lab tier.
- No new secret exposure or authentication regression was found; current CI, builds, version checks, CodeQL, and the full backend suite are green.

### Team conclusion

**Changes requested.** Findings 1–4 should be resolved before merge. Findings 5–12 are important follow-ups; 5–9 materially affect truthful operation and the user-visible contract and should preferably land with this PR.

MotWakorb:
## Review findings addressed — new head `596b155e`

All 4 blockers + follow-ups from the team review are fixed, plus 3 further issues (1 blocker, 2 data-integrity) that two internal adversarial reviews surfaced before re-push. PO-locked three design decisions along the way.

### Team-review blockers (1–4)
- **1 — strand on selected-profile enable failure** → **enable-first two-phase**: enable all selected first; if any selected-enable fails, abort *before any disable* and return `partial_failure` (RED-confirmed: reverting to single-pass re-strands). `3a4a6f89`
- **2 — permanent pipeline ownership** (PO decision: *implement handoff now*) → the `assign_channel_profile` action stamps the owning **rule id**; reconcile resolves the live profile-assigning rule set once per sweep and **releases** a channel (rejoins Auto-Sync + clears the marker) when its rule is disabled/deleted/no-longer-assigns. Conservative on failure (never releases on a transient DB error). Covers the event-sync path via the shared chokepoint. Residual limit (rule *condition* change) documented. `3a4a6f89`
- **3 — account-scoped selection collapsed to global** (PO decision: *global per group*) → deterministic per-group winner `(auto_sync ON, has-selection, lowest account id)`, cross-account conflict flag, one shared winner-resolution for save + sweep. `3a4a6f89` / `c76a3854`
- **4 — convergence backbone never retried** → the monitor now runs the idempotent sweep on **every** pass (not only on content change), cancellation-aware. `f9c38099`

### Follow-ups (5–12) — all landed this PR
Truthful `partial_failure` status + failed profile ids; per-group isolation in the save hook; direct hook-wiring tests; marker-write honesty; docs/api.md + CHANGELOG; unused-import/CodeQL cleanup. UX (#9): modal relabel to **"Stop managing profiles"** + incomplete-apply **warning toast** on every save path (driven by the now-truthful save response).

### Two internal adversarial reviews found + fixed before re-push
- **BLOCKER — cross-thread DB hazard**: the ownership-rule query was offloaded via `asyncio.to_thread`, racing the process-wide single SQLite (`StaticPool`) connection used inline by the event loop. Now runs **inline**. `c76a3854`
- **Cross-account override flap**: save and sweep resolved different winners for an S→T override, flapping channels every ~5 min → unified winner-resolution (`resolve_save_reconcile_targets` over the full settings set). `c76a3854`
- **Silent selection shadow**: a real selection could be silently ignored when a no-selection row won the tie-break → tie-break now prefers a selection-bearing row + broadened conflict flag. `c76a3854`

### Gates (green on head `596b155e`)
- Backend full suite: **7766 passed, 13 skipped**
- Frontend: **2160 passed**; `tsc --noEmit`, `vite build`, `eslint --max-warnings 0` all clean
- Version consistency: all three touchpoints at `0.17.6-0153`

Version steady at `0.17.6-0153`. No closing keyword — GH #720 gets a status update and closes when this merges. Ready for the team's re-review of head `596b155e`.

MotWakorb:
## Team Re-Review — Full (head `596b155e`)

**Verdict: Changes requested**  
**Effective deployment tier:** home-lab (`ecm-core`, `data-store`, `docs`, `dev-tooling`)  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

All current CI checks are green, and most earlier recommendations were implemented well. The re-review nevertheless found one new concurrency blocker and confirmed that several earlier requirements remain only partially addressed.

### Must fix before merge

1. **Concurrent reconciles can recreate zero-profile stranding** — Engineering, Architecture, Code Review, QA, SRE  
   Enable-first protects one invocation, but save, post-refresh, and monitor reconciliation can overlap without a per-effective-group lock or version check (`backend/services/profile_reconcile.py:447-518`, `backend/routers/m3u.py:1105-1150`, `backend/tasks/m3u_change_monitor.py:193-220`). Old `{A}` and new `{B}` runs can interleave `enable A → enable B → disable A → disable B`, leaving zero memberships.  
   **Action:** serialize by effective group, refetch/revalidate the winning selection after acquiring the lock, and add a controlled concurrency regression test plus a hook-level assertion that all entrypoints share coordination.

2. **Ownership-marker writes can silently overwrite unrelated `custom_properties`** — Database, Security  
   Marker addition uses the executor run-start cache (`backend/channel_pipeline_executor.py:3827-3852`); marker removal uses the earlier reconciliation snapshot (`backend/services/profile_reconcile.py:217-240`). Both replace the complete property object, so a concurrent EPG/logo/metadata update can be erased. Existing tests preserve only properties already present in the stale input. Marker persistence failure is also logged but the action still returns success (`channel_pipeline_executor.py:3000-3013,3857-3869`; the test at `test_channel_pipeline_executor.py:5023-5033` codifies this).  
   **Action:** fetch current channel state immediately before marker add/removal or use atomic/optimistic key updates; propagate marker failure as partial/incomplete rather than successful; add race-shaped tests for both directions.

3. **Incomplete reconciliation can still be presented as success** — Architecture, Engineering, UX, Code Review, SRE, QA, Technical Writing  
   If profile-universe fetch fails, all disables are skipped but the result may be `reconciled` (`profile_reconcile.py:386-420`). If save-hook setup fails before the per-group loop, the response contains an empty `ecm_profile_apply` (`m3u.py:1113-1150`), which `profileApplyIncomplete` treats as clean (`frontend/src/services/api.ts:928-938`). Scheduled task history likewise stays successful across reconcile errors/partial failures.  
   **Action:** return an explicit `degraded`/error outcome, include setup and aggregate errors, propagate them through task results and all UI save surfaces, and add service/router/frontend regressions.

### Status of the original 12 findings

| # | Original finding | Status | Re-review assessment |
|---|---|---|---|
| 1 | Selected-enable failure can strand channels | **Partial** | Sequential failure is fixed/tested; concurrent reconciles can recreate it |
| 2 | Permanent pipeline ownership | **Partial** | Disable/delete/action-removal handoff works; condition changes and marker races remain |
| 3 | Account settings collapsed globally | **Decision disputed** | Deterministic winner exists, but contradictory rows remain stored and a non-winning account save may apply another account's selection |
| 4 | Monitor never retries drift | **Resolved** | Sweep runs every monitor pass |
| 5 | Failures reported as reconciled | **Partial** | Bulk-write failures are truthful; universe-fetch and outer setup failures are not |
| 6 | Save override order dependence | **Resolved** | Save and sweep share winner resolution |
| 7 | One group aborts later save groups | **Resolved** | Per-group isolation landed |
| 8 | Runtime hooks lack tests | **Mostly resolved** | Focused tests landed; poll-branch and deeper cancellation wiring remain light |
| 9 | UI labels and feedback incorrect | **Partial** | Labels improved; some failures still show success and persistent states get misleading retry guidance |
| 10 | Stale whole-value property replacement | **Unresolved** | Logging landed, but stale replacement and false action success remain |
| 11 | API/documentation incomplete | **Partial** | Main side effects documented; degraded status, recovery, precedence, and rollback remain inaccurate/incomplete |
| 12 | Unused import and CodeQL | **Resolved** | Thread resolved/outdated; CodeQL green |

### Important follow-ups

- The generic warning tells stale selections and persistent cross-account conflicts that automatic retry will fix them. It will not; provide status-specific recovery instructions.
- Fully stale selected IDs can render the profile control blank (`AutoSyncSettingsModal.tsx:173-179`); partial staleness silently hides missing choices.
- Linked-account PATCH failures are log-only and can still produce an overall success toast (`M3UGroupsModal.tsx:258-285`).
- Linked-account-only and Sync Groups incomplete-apply behavior lack focused frontend tests.
- Durable documentation says winner precedence is `auto-sync ON → lowest account ID`, while implementation is `auto-sync ON → has selection → lowest account ID` (`dispatcharr_client.py:683-697`).
- Reverting ECM code will not restore profile memberships already changed by this feature; add a concise recovery/rollback note.
- The custom profile picker needs an accessible label plus popup/listbox state and keyboard semantics.

### Decision needed: meaning of “global per group”

Architecture/UX consider finding #3 unresolved because conflicting rows remain and account-scoped save behavior can apply a different account's winner. Engineering/Security/Database consider it resolved under the deterministic-winner decision.

Choose and document one contract:

- **Enforced global state:** propagate the selection to equivalent account rows or reject conflicting saves; or
- **Deterministic winner:** retain contradictory rows, but explain global scope before save, identify the winning account, and provide an actionable conflict-resolution path.

The current behavior sits between those models.

### Positive observations

- Strong behavior-focused coverage for stale IDs, pagination, idempotency, winner permutations, ownership handoff, and bulk-call complexity.
- Sequential enable-first safety, every-pass convergence, per-group isolation, shared winner resolution, and ownership rule identifiers are meaningful improvements.
- The earlier SQLite `StaticPool` cross-thread hazard is fixed by running the live-rule query inline.
- No new authentication, authorization, secrets, injection, or network-boundary regression was found.
- All visible CI, CodeQL, Semgrep, builds, frontend/backend tests, version checks, visual regression, and benchmarks are green.

**Team conclusion:** keep PR #724 unmerged until the concurrent-reconcile blocker, stale whole-property writes, and false-success paths are corrected.

MotWakorb:
## Re-review findings addressed — new head `f426da3b`

The full-re-review's 3 blockers + the "global per group" contract decision are resolved, and two internal adversarial reviews caught + fixed a regression the enforced-global work itself introduced.

### Team re-review blockers
- **Concurrent-reconcile stranding** → reconcile now serializes **per effective group** (module-level `asyncio.Lock` registry) with **post-lock revalidation** (re-reads the selection + channel list after acquiring the lock, so overlapping runs converge on the current selection, never a stale one, never zero memberships). A divergent-lock-key edge (concurrent Channel-Group-Override retarget) is closed by releasing + re-acquiring under the revalidated effective group. `feb509f2` / `417763aa`
- **Marker clobbers `custom_properties`** → both marker add and clear now **fresh-fetch** the channel's current properties immediately before the merge (fast-path idempotent skips on cache, so no extra GET when nothing changes). A marker-write failure is surfaced as a run-level `profile_ownership_not_established` warning (visible in Execution Details) while the assignment stays truthful `completed` — the profiles did apply. `feb509f2` / `e1d8477e`
- **Incomplete reconciliation shown as success** → universe-fetch-fail returns a distinct `degraded` status; save-setup failure emits an explicit `{status:"error"}` in the apply summary; the monitor folds `partial_failure`/`degraded`/per-group errors into its TaskResult (task history reads "completed with warnings"); the frontend treats all of these as incomplete with **status-specific recovery text** (no false "retry will fix it" for stale/conflict). `feb509f2` / `417763aa` / `0e553495`

### PO decision — "global per group" = **enforced global**
A save now **propagates** the selection to every sibling M3U-account row for that channel-group, so contradictory rows can't persist and a save always takes effect regardless of which account made it. **Fixed a data-loss regression** in the first cut (a field-only edit could clear another account's selection): the cascade now diffs the primary's new-vs-prior selection as sets and propagates **only genuinely-changed** groups. `be405471` / `c94d5f86`

### Follow-ups (landed)
Accessible listbox profile picker (roving `aria-activedescendant` + full keyboard), blank-picker stale-count alert, doc precedence corrected, rollback/recovery note, all save paths (primary + linked cascade + Sync Groups) warn on incomplete apply.

### Gates (green on head `f426da3b`)
- Backend full suite: **7781 passed, 13 skipped**
- Frontend: **2175 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at `0.17.6-0153`

Version steady at `0.17.6-0153`. No closing keyword — GH #720 gets a status update and closes when this merges. Ready for the team's re-review of head `f426da3b`.

MotWakorb:
## Team Re-Review — Full (head `f426da3b`)

**Verdict: Changes requested — not ready to merge**  
**Effective deployment tier:** home-lab (`ecm-core`, `data-store`, `docs`, `dev-tooling`)  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

The latest changes materially improve reconciliation, reporting, accessibility, and documentation, but the team found five remaining integrity blockers. All visible CI checks are green; however, the most serious defect crosses the frontend/backend boundary and is absent from the isolated test suites.

### Must fix before merge

1. **Frontend string profile IDs are discarded by integer-only backend logic** — Architecture, UX, Code Review, QA, Database, Technical Writing  
   The frontend defines `channel_profile_ids?: string[]`, keeps `Set<string>`, toggles using `profile.id.toString()`, and saves the strings unchanged (`frontend/src/types/index.ts:253-262`; `AutoSyncSettingsModal.tsx:337-351`). Frontend tests expect values such as `['2']`. Backend reconciliation, cascade, and winner helpers accept only genuine integers and silently drop strings (`backend/services/profile_reconcile.py:124-140`; `backend/routers/m3u.py:1057-1064`; `backend/dispatcharr_client.py:673-681`).  
   A normal modal selection can therefore become `no_selection`; a previous integer selection followed by a string selection can be interpreted as a clear and propagated to sibling rows.  
   **Action:** choose a canonical wire type, validate at the API boundary, reject invalid values with 422 rather than treating them as “Stop managing,” and add a UI-shaped `['12']` PATCH → propagation → reconciliation contract test.

2. **Pipeline profile assignment is not coordinated with reconciliation** — Architecture, QA, SRE  
   The per-effective-group locks serialize reconciliation calls only (`profile_reconcile.py:78-97,364-395`). Channel Pipeline exclusive membership updates and marker stamping do not acquire the same coordination primitive (`channel_pipeline_executor.py:2942-3022`). A reconcile can snapshot a channel as unowned, the pipeline can assign and stamp it, and the reconcile can then overwrite membership. The marker remains pipeline-owned, so later sweeps preserve the wrong membership.  
   **Action:** coordinate pipeline membership/marker writes with reconciliation or re-fetch/recheck ownership immediately before destructive reconciliation writes; add a controlled pipeline-vs-reconcile test.

3. **Enforced-global settings are neither concurrent-safe nor durably convergent** — Security, Architecture, PM, Database, QA  
   Primary save plus sibling cascade is not serialized (`backend/routers/m3u.py:1067-1157`). Opposing concurrent account saves can cross-propagate and finish with contradictory rows. A partially failed cascade also cannot heal through an ordinary retry: propagation runs only when the primary selection changed during that request (`m3u.py:1087-1102`), so re-saving the unchanged desired selection skips the failed sibling, and the monitor does not normalize account rows.  
   **Action:** serialize global-group saves with deterministic lock ordering and last-writer semantics; compare/repair sibling mismatches even when the primary did not change in that request; add concurrent-save and fail-first/retry-success regressions.

4. **Lock and revalidation failure paths still permit unsafe writes** — Security, Architecture, Engineering, Code Review, QA, SRE  
   After three effective-group retargets, reconciliation proceeds for the latest group while holding a prior group’s lock (`profile_reconcile.py:364-398`). If post-lock settings revalidation fails, it falls back to the stale pre-lock snapshot. Either path can reintroduce destructive interleaving or let an old queued run overwrite newer desired state.  
   **Action:** fail closed with no membership writes and let the scheduled sweep retry; add continuous-retarget and revalidation-failure tests.

5. **Failed fresh reads fall back to stale whole-value writes** — Security, Architecture, PM, Database  
   Marker addition/removal fresh-fetches on success but falls back to stale cache/snapshot data when the GET fails, then replaces the complete property object (`channel_pipeline_executor.py:3872-3910`; `profile_reconcile.py:257-281`). This re-enables the prior lost-update problem precisely during dependency degradation.  
   **Action:** skip the mutation when fresh state cannot be established, surface ownership-unestablished/degraded status, and retry later.

### Important findings

- Initial settings-fetch failure returns `groups_errored: 0`, leaving scheduled-task history falsely green (`profile_reconcile.py:674-683`).
- The monitor skips reconciliation when account filtering yields no accounts, contradicting the every-pass convergence guarantee.
- Post-refresh reconciliation ignores returned degraded/error aggregates.
- Task counters mix account totals with profile-group warning counts and can report more failures than total items.
- Fully stale IDs still produce a visually blank picker trigger; the UI says missing IDs will be dropped although save retains them.
- The account-scoped modal does not disclose that profile selections propagate across accounts.
- Warning prioritization can hide simultaneous failures, and refresh failure can mask an existing apply warning.
- Linked-account save failures are mislabeled as profile-application failures.
- Cancellation remains between groups rather than during long pagination/profile-write phases.
- Identical queued sweeps are serialized but not coalesced.

### Status of the original findings

| # | Finding | Status |
|---|---|---|
| 1 | Zero-profile stranding | **Partial** — ordinary reconcile overlap fixed; pipeline race and wrong-lock fallback remain |
| 2 | Permanent pipeline ownership | **Partial** — lifecycle handoff improved; coordination and stale-marker paths remain |
| 3 | Account-scoped collapse | **Partial** — enforced-global happy path landed; concurrency and retry convergence remain |
| 4 | No convergence retry | **Partial** — ordinary passes reconcile; no-account early return and row-normalization gap remain |
| 5 | False success reporting | **Partial** — ordinary degraded paths fixed; initial sweep failure remains clean |
| 6 | Override order dependence | **Resolved** |
| 7 | One group aborts later groups | **Resolved** |
| 8 | Hook coverage | **Mostly resolved** |
| 9 | UI feedback and labels | **Mostly resolved**, with remaining aggregation/scope inaccuracies |
| 10 | Whole-value property replacement | **Partial** — fresh-read success fixed; failure fallback remains unsafe |
| 11 | Documentation contract | **Partial** — several absolute guarantees remain unsupported |
| 12 | Cleanup and CodeQL | **Resolved** |

### Decisions needed

- **Canonical profile-ID representation:** recommended integer IDs throughout the wire contract, converting in the modal and validating with a typed backend request model. Alternatively, accept numeric strings and canonicalize consistently. Silent filtering is not acceptable.
- **Concurrent global-save semantics:** recommended last lock-acquiring request wins, with all sibling rows converged before success is returned.

### Positive observations

- Ordinary reconcile-to-reconcile serialization and lock-key retargeting are meaningful improvements.
- `degraded`, per-group error, and save-setup outcomes reach task/UI warning paths in normal cases.
- Marker failures are now visible in execution details.
- Status-specific recovery text removes the earlier false automatic-retry promise.
- The listbox has substantially improved keyboard and screen-reader support.
- Rollback irreversibility and profile recovery are documented.
- The SQLite cross-thread hazard remains fixed.
- No new authentication, authorization, injection, secret, or network-boundary vulnerability was found.
- Backend, frontend, MCP, builds, CodeQL, Semgrep, visual regression, version checks, and benchmarks are green.

**Team conclusion:** keep PR #724 and bead `enhancedchannelmanager-y3m6o.3` in progress until these integrity blockers are resolved.

MotWakorb:
## Re-review findings addressed — new head `8794e027`

The full re-review's 5 blockers + 10 findings are resolved, both PO decisions (canonical ID = INTEGER; full hardening) honored, and two internal adversarial reviews caught + fixed the remaining gaps.

### The critical one — the feature was broken end-to-end
**Blocker 1 (string/int):** the modal saved profile IDs as strings (`["12"]`) but the backend kept integers only → every real save silently no-op'd (`no_selection`). Ground truth: Dispatcharr uses **integer** profile IDs, and ECM already used int everywhere except this modal. Fixed: modal → `number[]`; **422 validation** at the group-settings PATCH boundary (never silent-drop); strict ASCII coercion of legacy numeric strings (rejects `--5`/unicode digits, no crash); coercion applied consistently at the collapse/winner/conflict site. **A UI-shaped E2E contract test** (`[12]` → PATCH → propagate → reconcile → channel enabled in 12) now crosses the boundary that both isolated suites missed. `01a09dcc` / `dc1de356` / `6e945390`

### Concurrency / durability (PO: full hardening)
- **B2 pipeline↔reconcile — fully closed.** The pipeline executor now acquires the **same per-effective-group lock** the reconcile uses (once-per-run cached settings; per-channel acquire/release; no re-entrancy/over-lock), so a pipeline assignment and a group reconcile can never touch the same group concurrently. Re-check + marker-before-membership remain as defense-in-depth. `91ef10ff`
- **B3 enforced-global — concurrent-safe + durably convergent.** Primary PATCH + sibling cascade run as one unit under sorted effective-group locks (no divergent interim); `normalize_group_selections` runs every sweep to self-heal divergent/failed sibling rows. `01a09dcc` / `6e945390`
- **B4 fail-closed:** lock-instability or revalidation-fetch-failure → no writes, `degraded`, sweep retries. `01a09dcc`
- **B5 no stale writes:** marker add/remove skip entirely on a failed fresh read, surface ownership-unestablished. `01a09dcc`

### The 10 findings + internal catches
Task-history honesty (initial-fetch-fail counted; sweep runs regardless of account filtering; post-refresh surfaces degraded), **truthful monitor counters** (`total_items ≥ success/failed` invariant, with a regression test), stale-ID picker copy, cross-account disclosure, both-warnings surfaced, linked-account labeling, cancellation in pagination/write phases, sweep coalescing. `27590359` / `8794e027`

### Gates (green on head `8794e027`)
- Backend full suite: **7825 passed, 13 skipped**
- Frontend: **2175 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at `0.17.6-0153`

Version steady at `0.17.6-0153`. No closing keyword — GH #720 gets a status update and closes when this merges. Ready for the team's re-review of head `8794e027`.

MotWakorb:
## Team Re-Review — Full (head `8794e027`)

**Verdict: Changes requested — not ready to merge**  
**Effective deployment tier:** home-lab (`ecm-core`, `data-store`, `docs`, `dev-tooling`)  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

The latest head resolves many previous findings, including the integer-ID contract, ordinary reconcile locking, failed revalidation behavior, marker fresh-read fallback, task counters, and several UX issues. Five integrity blockers remain.

### Must fix before merge

1. **TLS mode bypasses the shared-lock guarantee** — Architecture, SRE, Technical Writing  
   The registry consists of module-local `asyncio.Lock` objects (`backend/services/profile_reconcile.py:93-126`). Standard HTTP runs in one process, while TLS mode starts another `uvicorn main:app` subprocess (`backend/tls/https_server.py:52-59,100-118`). HTTPS mutations and main-process monitoring/reconciliation therefore use different lock registries. Same-process tests cannot verify this boundary.  
   **Action:** make one process the sole writer and proxy TLS traffic to it, introduce cross-process coordination, or remove the second mutation-capable application process.

2. **Pipeline writes unlocked when group resolution fails** — Security, Engineering, Code Review, QA, SRE  
   When group settings cannot be fetched or the effective group cannot be resolved, pipeline execution explicitly proceeds without the shared lock (`backend/channel_pipeline_executor.py:2869-2899,2986-3010`). Marker-first ordering and reconciliation’s ownership recheck do not close the race after reconciliation has already passed its recheck.  
   **Action:** fail closed when an assigned channel has a group but its effective lock key cannot be established; add a deterministic settings-fetch-failure pipeline-vs-reconcile test.

3. **A partially failed global clear can be reversed** — Security, Architecture, PM, Code Review, Database, QA, Technical Writing  
   Normalization records only non-empty selections (`backend/services/profile_reconcile.py:851-869`), while collapse prefers a sibling containing a selection over an empty sibling (`backend/dispatcharr_client.py:692-710`). If the primary clear succeeds but one sibling clear fails, the next sweep can elect the stale sibling and restore its old selection into the accounts that were successfully cleared. The operator’s explicit “Stop managing” intent is lost.  
   **Action:** persist authoritative global desired state that can represent an explicit empty selection, such as a versioned record or clear tombstone; add partial-clear-failure followed by sweep coverage.

4. **Normalization can overwrite a newer save from stale pre-lock state** — Architecture, Database, QA, SRE  
   Normalization computes its winner and account-row payloads before acquiring effective-group locks (`profile_reconcile.py:851-904`), then later locks and writes that stale plan (`:907-914`). A newer save can finish while normalization waits and then be overwritten.  
   **Action:** acquire locks before the authoritative read, recompute winner/targets inside the critical section, and add a controlled normalize-vs-save regression.

5. **Save failure paths bypass serialization or overwrite unrelated fields** — Security, Architecture, PM, Database, Code Review, QA  
   If lock-key settings fetch fails, the route records an error but still performs the primary selection write unlocked (`backend/routers/m3u.py:1289-1303,1351-1354`). Account snapshots and full sibling payloads are also built before lock acquisition (`m3u.py:1182-1203`). Override-only changes may acquire no group lock even though they alter lock topology, and pipeline mapping is not post-lock revalidated during such transitions.  
   **Action:** fail closed for selection changes when lock keys cannot be resolved; lock the union of old/proposed effective groups; re-read rows inside the lock; avoid stale full-row replacement.

### Important findings

- Normalization account-list failure returns zero failures, leaving task history falsely green.
- Monitor account-capture exceptions are logged but not counted as task warnings.
- Post-refresh reconciliation can be incomplete while an unconditional refresh-success alert is emitted.
- Sweep coalescing drops a non-equivalent follower rather than scheduling a trailing pass.
- The marker-cache fast path can trust a marker removed externally.
- Global propagation failures are reduced to a generic “check the logs” warning rather than identifying affected accounts and recovery.
- The profile picker’s accessible name omits its current selection and lacks `aria-controls`.
- Simultaneous apply problems remain priority-collapsed.
- Some concurrency tests rely on fixed sleeps rather than deterministic barriers.
- Documentation calls sequential remote PATCHes an “atomic unit” and makes unsupported “never” and “no contradictory rows persist” claims.

### Previous finding status

| Area | Status |
|---|---|
| Canonical integer IDs and 422 validation | **Resolved** |
| Legacy numeric-string coercion | **Resolved** |
| Ordinary reconcile-to-reconcile locking | **Resolved in one process** |
| Wrong-lock retry exhaustion | **Resolved** |
| Failed post-lock revalidation | **Resolved** |
| Marker fresh-read failure fallback | **Resolved** |
| Ordinary opposing global saves | **Resolved in healthy, same-process path** |
| Initial sweep settings-fetch reporting | **Resolved** |
| No-account monitor sweep | **Resolved** |
| Cancellation and counter invariants | **Resolved** |
| Durable global clear | **Unresolved** |
| Pipeline failure-path coordination | **Unresolved** |
| Cross-process coordination | **Unresolved** |
| Normalization freshness/truthfulness | **Unresolved** |

### Decisions required

- **Cross-process coordination:** recommended approach is one process as sole mutation/background-work owner with TLS proxied to it; alternative is a cross-process lock.
- **Durable global desired state:** recommended approach is a versioned group-level desired selection including explicit empty; alternative is a durable retry/tombstone record tied to the latest intent.

### Positive observations

- UI and backend now use canonical integer IDs, with invalid inputs rejected instead of silently becoming clears.
- Cross-layer and legacy-input tests materially improve contract coverage.
- Sorted same-process lock acquisition avoids lock-order deadlocks.
- Reconciliation fails closed on unstable keys and failed revalidation.
- Marker writes skip mutation when their fresh read fails.
- Monitor, cancellation, and task-counter behavior improved.
- Global-scope disclosure, stale-ID handling, and listbox keyboard behavior improved.
- No new authentication, authorization, injection, secret, or network-boundary vulnerability was found.
- All visible CI checks pass and the diff is whitespace-clean.

**Team conclusion:** keep PR #724 and bead `enhancedchannelmanager-y3m6o.3` in progress until these integrity blockers are resolved.

MotWakorb:
## Round-8 findings addressed — new head `0524a1cc`

Head `8794e027` review: several findings resolved further, plus five that are **cross-process / distributed-durability** in nature. Per a deliberate scope decision (below), this round ships the tier-appropriate correctness + honesty and **defers the distributed-systems work to a tracked follow-up bead** rather than building cross-process locks and tombstone state into a single-operator home-lab tool.

### Scope decision: right-size, don't over-build
Finding 1 established that the per-effective-group lock registry is in-process, and TLS mode runs a second app subprocess — so an in-process lock can't serialize across processes. Rather than build cross-process distributed locking + versioned tombstone desired-state for a single-operator tier, we:
- **Confirmed** background/monitor/reconcile work runs **only in the main process** (gated on the HTTPS-subprocess env), so no double-monitor contention exists.
- Made the concurrency **guarantees honest**: downgraded "atomic / never / no contradictory rows persist" docstrings to the real **same-process, single-operator** guarantee; the sibling cascade is documented as **sequential remote PATCHes, not an atomic transaction**.
- **Deferred** cross-process coordination, versioned/tombstone durable desired-state, and deep normalize/save TOCTOU to a dedicated hardening bead — to be designed upfront **if/when the tier changes** (single-operator/single-process is the current, documented assumption). Code comments at the lock registry, save, executor, and normalize sites point to it.

### Fixed this round (tier-appropriate correctness + honesty)
- **Partial "Stop managing" clear** no longer silently succeeds: an incomplete clear surfaces a **named-account error + recovery action**, no false success toast. (The durable auto-heal of a partially-failed clear is the deferred tombstone work.)
- **Fail-closed** on unresolvable lock keys — both the pipeline executor (channel has a group but lock key can't resolve) and the save route (selection change + lock-key fetch failure) now fail closed and retry, **without** blocking healthy saves or field-only edits (verified write-free + non-over-blocking).
- **No phantom audit entry:** a fail-closed no-write save no longer emits an "Updated group settings" journal entry for a change that was never applied (the journal only records applied changes now).
- **Honesty findings:** normalize account-list fetch failure is counted (not silently zero); monitor change-capture exceptions counted; post-refresh success alert gated on the reconcile outcome; propagation failures name the affected accounts.
- **Truthful counters:** `total_items ≥ success_count/failed_count` invariant holds in every case incl. the zero-accounts + errored-sweep double-fault (regression tests added).
- a11y (picker accessible name includes selection + `aria-controls`), coalescing schedules a trailing pass instead of dropping a non-equivalent follower.

### Gates (green on head `0524a1cc`)
- Backend full suite: **7833 passed, 13 skipped**
- Frontend: **2175 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at `0.17.6-0153`

The remaining round-8 items that are cross-process / durable-desired-state are **intentionally deferred** to the follow-up hardening bead, not pending here. Version steady at `0.17.6-0153`; no closing keyword — GH #720 gets a status update and closes when this merges. Ready for re-review of head `0524a1cc` against the single-operator tier.

MotWakorb:
## Team Re-Review — Full (head `0524a1cc`)

**Verdict: Changes requested — not ready to merge**  
**Effective deployment tier:** home-lab (`ecm-core`, `data-store`, `docs`, `dev-tooling`)  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

The new head contains real improvements: grouped pipeline assignments and selection-changing saves now fail closed when lock keys cannot be resolved; phantom journal entries are suppressed; normalization/capture/post-refresh reporting is more truthful; the picker accessibility fix landed; and coalescing now attempts a trailing pass. The team nevertheless found three immediate contract blockers plus a current-topology decision that must be resolved before merge.

### Must fix before merge

1. **A no-write save still returns HTTP 200 and the UI reports “Saved”** — PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing  
   On lock-key fetch failure for a genuine selection change, the route deliberately performs no Dispatcharr write (`backend/routers/m3u.py:1411-1428`) but returns the ordinary successful response (`:1547-1555`). The regression test explicitly expects 200 while asserting no update and no journal (`backend/tests/routers/test_profile_reconcile_hooks.py:483-511`). The frontend treats the resolved request as saved, may continue linked saves, starts refreshes, calls `onSaved`, closes the modal, and displays “Saved, but…” (`frontend/src/components/M3UGroupsModal.tsx:238-327`; `frontend/src/services/api.ts:950-967`).  
   **Action:** return a retryable non-2xx response (preferred), or add a mandatory top-level `save_applied:false` contract handled by every caller. The UI must keep retry state, skip linked saves/refresh, and say “Not saved.” Add an end-to-end component regression.

2. **Coalesced followers still receive false terminal success and lose their request state** — PM, Engineering, Code Review, Database, SRE, QA, Technical Writing  
   A follower only sets `_sweep_pending=True` and immediately receives an all-zero `{coalesced:true}` result (`backend/services/profile_reconcile.py:957-983`). Its `client`, `all_settings`, and `cancel_check` are discarded; the leader’s trailing loop reuses the leader’s arguments (`:984-995`). A cancelled leader or stale supplied settings can therefore make the trailing pass useless. Post-refresh interprets the zero counters as complete and may announce success before queued work runs or fails (`backend/routers/m3u.py:222-249,301-345`).  
   **Action:** retain pending request state or use a shared future/result that followers await; alternatively return a distinct nonterminal `queued` outcome that no caller maps to success. Add cancelled-leader/distinct-follower/failing-trailing-pass and post-refresh result tests.

3. **The claimed bounded trailing loop is unbounded** — Security, Engineering, Code Review, SRE, QA, Technical Writing  
   `while _sweep_pending` can continue indefinitely if a follower arrives during each pass, despite comments promising one bounded trailing pass.  
   **Action:** cap the trailing work to a defined pass count or implement an explicit worker/generation handoff, with deterministic boundedness coverage.

4. **TLS concurrency is a current supported topology, not a future tier-change condition** — Architecture, PM, Engineering, Code Review, Database, SRE, QA, Technical Writing; Security rates Medium/acceptable only with explicit risk acceptance  
   Locks/coalescing are process-local (`profile_reconcile.py:80-125`), while TLS starts another request-serving `uvicorn main:app` process (`backend/tls/https_server.py:52-64,100-118`). Only background services are skipped in that process (`backend/main.py:1129-1133`). A single operator’s HTTPS save can therefore overlap the main-process monitor today. `enhancedchannelmanager-nq3ed` exists and captures the technical hardening, but it has no dependency on `y3m6o.3` and incorrectly activates only on a future tier change.  
   **Action / PO decision:** choose one currently supported contract: (A) make TLS a single-writer/proxy path now; (B) implement cross-process coordination; or (C) enforce that profile-mutating saves are unsupported through the TLS subprocess. Update bead dependencies and public docs accordingly. Merely documenting “single operator” does not prevent the automatic monitor from being a concurrent writer.

### Data-integrity conflict requiring explicit disposition

Security considers the following Medium residuals acceptable at home-lab if explicitly accepted and tracked. Database and Architecture consider them merge-blocking under the current exact/global acceptance contract:

- **Partial global CLEAR can still resurrect the old selection.** Empty desired state is not durable; a failed sibling’s non-empty row can win normalization and be restored into cleared rows (`profile_reconcile.py:867-895`; `dispatcharr_client.py:692-710`).
- **Pre-lock full-row snapshots can overwrite unrelated fields.** Save and normalize assemble remote row updates before acquiring locks (`routers/m3u.py:1217-1229`; `profile_reconcile.py:917-943`).

A full tombstone system is not the only home-lab option. Database proposed a bounded clear fix: update siblings first and the authoritative primary last, aborting before primary clear on sibling failure, plus fresh-under-lock reads. Otherwise the public contract must be explicitly narrowed to best-effort and the operator must receive actionable recovery details.

### Important findings

- Backend incomplete-clear errors name affected accounts and recovery, but the frontend discards `outcome.error` and shows generic “check the logs.”
- Post-refresh warnings universally promise next-sync retry, which is false for a partially failed clear.
- CHANGELOG/API still claim “durable guarantee,” “no contradictory rows persist,” and “never” concurrent stranding despite documented TLS/clear limitations.
- `docs/api.md` treats a rejected primary write as downstream reconciliation failure and documents only a successful 200 envelope.
- `ProfileApplyOutcome` cannot distinguish primary-not-written, sibling-partial, membership-incomplete, and queued/nonterminal outcomes.
- Cancellation can discard already observed failure counts/details.
- Capture failure counters do not identify affected accounts in task details.
- Accessibility implementation now includes current value and `aria-controls`, but tests do not assert those exact behaviors.
- Some concurrency tests still use timing sleeps instead of Events/barriers.
- Several inline comments/log messages describe behavior superseded by the fail-closed implementation.

### Confirmed resolved/improved

- Canonical integer profile IDs, strict validation, legacy coercion, and cross-layer coverage.
- Grouped pipeline assignment fails closed on unresolved topology; genuinely groupless channels remain allowed.
- Selection-changing save fails closed; field-only edits are not overblocked.
- Phantom journal entry on no-write path is suppressed.
- Unstable lock-key and post-lock revalidation paths fail closed.
- Marker fresh-read failures skip mutation.
- Normalize account-list and monitor capture failures now affect warning counters.
- Non-coalesced post-refresh incomplete outcomes gate success alerts.
- No-account monitor sweep, cancellation checkpoints, and counter invariants improved.
- Picker current-value accessible naming and trigger/listbox relationship landed.
- No new authentication, authorization, injection, secret, or network-boundary issue was found.

### CI status at posting

Backend, frontend, MCP, CodeQL, Semgrep, Fake-Test Guard, visual regression, benchmark, version consistency, and Wait for Test Suite are green at immutable head `0524a1cc`. Multi-arch ECM/MCP image builds are still in progress, so CI is not yet fully terminal.

**Team conclusion:** keep PR #724 and `enhancedchannelmanager-y3m6o.3` in progress. Resolve the no-write response and coalesced-result blockers, make the TLS topology decision explicit, and either satisfy or formally narrow the remaining exact/global data contract before merge.

MotWakorb:
## Review Scope Clarification — Frozen Merge Gate

The repeated review rounds have started to move the goalposts. Some later findings were valid regressions introduced while fixing earlier issues, but the team also repeatedly re-opened hardening risks that the PO had already chosen to defer for the current home-lab tier. That is not a fair or useful merge process.

This comment narrows and freezes the remaining merge gate for PR #724.

### Required before merge

1. **Truthful no-write save contract**  
   If a selection-changing PATCH fails closed and performs no primary write, it must not appear as an ordinary successful save or trigger linked saves/refresh. Use a retryable non-2xx response or an explicit `save_applied:false` contract handled by every caller. The UI must say the change was not saved and preserve a retry path.

2. **Truthful coalesced-sweep outcome**  
   A queued/coalesced follower must not receive a terminal-looking zero-failure result that callers map to success. Preserve/execute appropriate follower state or expose a distinct nonterminal outcome; post-refresh must not announce successful profile reconciliation before queued work completes. Bound the trailing-pass behavior and cover it with deterministic tests.

3. **Contract and recovery documentation**  
   Public API, changelog, UI copy, and alerts must match the behavior actually supported in this release. Remove categorical `atomic`, `never`, `durable guarantee`, and `no contradictory rows persist` wording where the accepted limitations apply. Surface actionable recovery for a primary no-write and partial sibling propagation.

4. **Terminal green CI**  
   All required checks must complete successfully on the reviewed head.

### Explicitly accepted and deferred for this PR

The following are **not merge blockers for PR #724** under the current PO-approved home-lab scope. They remain tracked in `enhancedchannelmanager-nq3ed`:

- TLS/cross-process coordination or single-writer redesign
- Durable explicit-empty/tombstone desired state for crash-hard partial-clear recovery
- Deep normalize/save TOCTOU and optimistic-concurrency hardening
- Broader stale full-row protection against non-cooperating external writers

These risks should be documented accurately, and the follow-up bead should state that TLS already creates the relevant multi-process topology. They must not be reclassified as blockers in another review unless a new change materially worsens them or new concrete evidence invalidates the PO’s accepted premise.

### Next review protocol

The next review is **delta-only and binary**:

- Verify the two behavioral fixes above.
- Verify documentation/recovery accuracy.
- Check for regressions introduced by that delta.
- Confirm terminal green CI.
- Approve if those gates pass.

No new architectural hardening requirements will be added to this PR merely because another review round occurs. This frozen gate supersedes the broader blocker classification in the immediately preceding team-review comment.

MotWakorb:
## Round-9 findings addressed — new head `32754990`

Three of the four round-9 blockers were incompleteness in the round-8 fixes (not new scope); one was a fair correction (TLS is a current topology). All resolved the tier-appropriate way — **build less**, take the cheap correct fixes, no distributed machinery.

### Blockers
- **1 — no-write save read as "Saved":** the fail-closed selection-change path now returns **HTTP 503** (retryable) naming the affected account(s) + recovery — never 200, no write, no journal entry. Frontend on 503: modal **stays open**, no refresh, no linked saves, `onSaved` not called, "NOT saved" shown. (Field-only + healthy saves stay 200.)
- **2 & 3 — coalescing false-success + unbounded loop:** the trailing-pass mechanism was **removed** (it was over-built). A coalesced follower returns a non-terminal `queued` that no caller maps to success; the idempotent scheduled sweep converges the deferred work.
- **4 — TLS is a current supported topology (main = sole writer):** the HTTPS subprocess (a second request-serving process) now **proxies all profile-mutating / reconcile-triggering routes to the main process**, so channel-profile writes + reconcile execute only in main where the in-process lock is authoritative. Forward allowlist (audited for completeness): group-settings PATCH, `m3u/refresh/{id}`, `channel-pipeline/run`, `channel-pipeline/rules/{id}/run`, `tasks/{id}/run`. SSRF-safe (subprocess-gated, hard-coded localhost target, anchored allowlist, auth preserved, hop-by-hop stripped, main-unreachable → 502).

### Data-integrity — partial CLEAR (bounded fix, no tombstones)
The clear now clears **siblings first** (fresh read under the lock), the **authoritative primary last**, and **aborts before the primary clear if any sibling clear fails** (503, named account) — so a failed sibling can no longer be re-elected to resurrect the old selection. Closes the resurrect without versioned/tombstone desired-state.

### Honesty + cheap
Named-account errors now surface in the UI (not "check the logs"); post-refresh warnings are outcome-accurate (no false "retry next sync" for a failed clear); CHANGELOG/`docs/api.md` downgraded to the honest same-process guarantee and document the 503 fail-closed envelope + clear-ordering.

### Deliberately NOT built (tracked, not gating)
Four-way `ProfileApplyOutcome` discriminated type (a boolean `applied` + `error` + `queued` suffices); the general pre-lock full-row-snapshot TOCTOU and two pre-existing unlocked membership paths (`/api/profiles/…bulk`, DBAS-restore task-run) are recorded on the follow-up hardening bead — out of scope at this tier.

### Gates (green on head `32754990`)
- Backend full suite: **7840 passed, 13 skipped**
- Frontend: **2177 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at `0.17.6-0153`

Version steady at `0.17.6-0153`; no closing keyword — GH #720 gets a status update and closes when this merges. Ready for re-review of head `32754990` against the single-operator tier.

MotWakorb:
## Frozen-Gate Team Re-Review — head `32754990`

**Binary verdict: Changes requested**  
**Scope:** delta `0524a1cc..32754990` only  
**Effective tier:** home-lab  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

This review followed the frozen merge gate. It did **not** reopen TLS/distributed hardening, durable tombstones, deep TOCTOU, or external-writer follow-ups. The findings below are concrete omissions or false outcomes introduced by the new delta itself.

### Frozen gates that passed

- A lock-key fail-closed selection change now returns retryable HTTP 503 instead of 200.
- The frontend keeps the modal open, performs no refresh/linked save, does not call `onSaved`, and displays “NOT saved.”
- The unbounded trailing-sweep loop was removed.
- Post-refresh recognizes `queued` as nonterminal rather than clean success.
- Named backend recovery details now reach frontend warnings.
- Documentation is materially more honest about 503 behavior, sequential clear ordering, rollback boundaries, and same-process locking.
- Required CI is terminal green at immutable head `327549908701e1177850570e237738b515345e5c`, including backend, frontend, MCP, CodeQL, Semgrep, visual regression, benchmark, version checks, and AMD64/ARM64 ECM/MCP builds.

### Remaining delta blockers

1. **TLS sole-writer proxy coverage is incomplete** — Security, Engineering, PM, Code Review, Database, QA, Technical Writing  
   The new proxy forwards canonical pipeline-run routes but not the supported deprecated aliases mounted from the same router:
   - `POST /api/auto-creation/run`
   - `POST /api/auto-creation/rules/{id}/run`

   It also omits ECM’s direct membership writers:
   - `PATCH /api/channel-profiles/{profile_id}/channels/bulk-update`
   - `PATCH /api/channel-profiles/{profile_id}/channels/{channel_id}`

   Evidence: `backend/tls/subprocess_proxy.py:42-48`, dual router mount at `backend/main.py:699-709`, membership routes at `backend/routers/profiles.py:135-171`. These first-party routes still execute in the HTTPS subprocess, contradicting the delta’s main-process-only writer boundary.  
   **Action:** add exact anchored proxy entries and parameterized canonical/deprecated/membership route tests.

2. **TLS task run/cancel lifecycle is split, and long runs gain a new 60-second failure ceiling** — Architecture, PM, Code Review, Database, SRE, QA, Technical Writing  
   `POST /api/tasks/{id}/run` is forwarded to main, but `/cancel` remains in the HTTPS subprocess and consults a different process-local task engine. The proxy also applies `httpx.AsyncClient(timeout=60.0)` to task runs that intentionally await terminal completion and are exempt from main’s ordinary request timeout. A long task can therefore return proxy 502 while continuing in main, and the HTTPS cancel path cannot stop it.  
   Evidence: `backend/tls/subprocess_proxy.py:42-48,82-86`, `backend/routers/tasks.py:578-639`, `backend/main.py:421-456`.  
   **Action:** forward run and cancel coherently to the same engine and preserve the endpoint’s long-running contract—or change run to enqueue/202 with polling. Add run→cancel affinity and timeout-contract tests.

3. **A nonterminal queued reconcile is still recorded green by the scheduled monitor** — Security, PM, Code Review, Database, SRE, UX, QA, Technical Writing  
   `_queued_result()` correctly returns only `{status:'queued', coalesced:true}` and explicitly says callers must not map it to success. Post-refresh honors that. `M3UChangeMonitorTask`, however, ignores `status`, treats missing counters as zero, and can return `success=True` / “no external changes detected.”  
   Evidence: `backend/services/profile_reconcile.py:957-987`, `backend/tasks/m3u_change_monitor.py:200-229,261-309`.  
   **Action:** map queued to a truthful deferred/warning task-history outcome and add a monitor-level queued regression.

4. **Multi-sibling CLEAR can partially mutate while returning a false no-write 503** — Architecture, PM, UX, Code Review, Database, SRE, QA, Technical Writing  
   Sibling PATCHes are sequential. With siblings A and B, A can clear successfully and B can fail. The primary remains unchanged, but the response says the prior selection was preserved, `applied=False`, and no journal entry is written—even though A changed. Existing coverage models only a single sibling failure before any successful sibling mutation.  
   Evidence: `backend/routers/m3u.py:1150-1194,1242-1267,1400-1411`; current test around `backend/tests/routers/test_profile_reconcile_hooks.py:540-580`.  
   **Action:** compensate previously changed siblings before returning the frozen no-write response, or expose and journal a truthful partial-application result. Add a three-account earlier-success/later-failure regression. The current “prior selection preserved” wording must match the chosen behavior.

5. **Malformed account enumeration allows an unsafe primary CLEAR** — Engineering, PM, Code Review, Database, QA, Technical Writing  
   A non-list `get_m3u_accounts()` result is silently converted/ignored rather than treated as an error. The sibling list becomes empty and the primary can be cleared even though ECM could not verify or update siblings.  
   Evidence: `backend/routers/m3u.py:1230-1251,1252-1267`.  
   **Action:** require a valid fresh account list for CLEAR; return 503 with zero writes/journal on malformed or failed enumeration. Add `None`, object/non-list, and fresh-read-exception tests.

### Documentation corrections tied to these fixes

After implementation, update the claims together:

- Main-process-only writer language must match the complete proxy allowlist.
- Task API documentation must describe coherent run/cancel and timeout behavior.
- “Prior selection preserved” / “NOT written” must be true after a later sibling failure.
- `docs/api.md` must distinguish downstream reconcile failure (200 + apply outcome) from pre-write/cascade safety failure (503).
- Monitor task history must describe queued work as deferred, not successful.

### Scope discipline

No accepted follow-up was promoted back into this PR. Durable tombstone state, generalized deep TOCTOU protection, and non-cooperating external-writer hardening remain outside this frozen gate. The items above are all direct behavior or coverage gaps in the new proxy/503/queued/CLEAR implementation.

**Team conclusion:** keep PR #724 in progress. Fix these five narrow delta issues, rerun terminal CI, and perform one final delta-only verification against this unchanged gate.

MotWakorb:
## Round-10 delta findings addressed — new head `7befad16`

The frozen-gate review found five narrow defects in round-9's own delta (proxy/503/queued/CLEAR) — all fixed, and an internal failure-mode confirm caught one further CLEAR edge, now closed too.

### Frozen-gate blockers
- **1a — proxy alias coverage:** `channel_pipeline_router` is dual-mounted (`/api/channel-pipeline` + deprecated `/api/auto-creation`); the alias run/rules-run now forward to main too. Full third-pass allowlist audit confirms no other reconcile/assign route (rollback/restore/event-sync are not membership/reconcile paths) is served by the subprocess.
- **1b — narrowed the claim (not forwarded):** the direct membership endpoints (`/api/channel-profiles/{id}/channels/…`) never took the reconcile lock, so forwarding them would be theater; the "main is the sole writer" language is scoped to the auto-sync reconcile + `assign_channel_profile` lock-participating paths (the pre-existing direct-membership endpoints remain tracked in the follow-up hardening bead).
- **2 — task run/cancel coherence + timeout:** task `cancel` now forwards alongside `run` (same main engine, so a forwarded run is cancellable), and the proxy uses `timeout=None` so a legitimately long task run is not falsely cut to a 502 (main's request-timeout middleware, which exempts task routes, is authoritative).
- **3 — monitor queued=deferred:** a coalesced (`queued`) sweep is now recorded as a deferred/warning task outcome (never green), truthful that the reconcile was deferred to the next sweep.
- **4 — truthful partial CLEAR:** a multi-sibling clear where an earlier sibling succeeded and a later one failed now returns a 503 that NAMES the changed + failed accounts and JOURNALS the sibling(s) that actually changed — no more "no write / preserved" when a sibling was written. The authoritative primary is never cleared while any sibling failed (no resurrection).
- **5 — fail-closed enumeration:** a CLEAR requires a valid fresh account list that CONTAINS the account being edited; None / non-list / raised / empty / primary-absent → 503, zero writes, no journal, primary never cleared.

### Docs
Sole-writer language matches the (reconcile-complete) allowlist + narrowed claim; task API describes coherent run/cancel + the long-running timeout contract; the partial-clear wording matches the truthful behavior; docs/api.md distinguishes downstream-reconcile-failure (200 + apply outcome) from pre-write/cascade safety failure (503); monitor history describes queued work as deferred.

### Gates (green on head `7befad16`, independently re-verified)
- Backend full suite: **7852 passed, 13 skipped**
- Frontend: **2177 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at `0.17.6-0153`

Deferred (tracked, not gating): the direct membership endpoints, the general pre-lock full-row-snapshot TOCTOU, and durable tombstone state remain in the follow-up hardening bead — out of scope at the single-operator tier. Version steady at `0.17.6-0153`; no closing keyword — GH #720 gets a status update and closes on merge. Ready for a delta-only re-verification of head `7befad16`.

MotWakorb:
## Frozen-Gate Team Re-Review — head `7befad16`

**Binary verdict: Changes requested**  
**Scope:** delta `32754990..7befad16` only  
**Effective tier:** home-lab  
**Participants:** Security, Architecture, PM, Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

The review remained inside the unchanged frozen gate. Accepted direct-membership, durable-tombstone, deep-TOCTOU, and external-writer hardening were not reopened.

### Five posted fixes: PASS

1. **TLS proxy route/lifecycle coverage — PASS**  
Both deprecated `/api/auto-creation` run aliases now forward to main. Task `run` and `cancel` forward to the same main-process engine, and `timeout=None` removes the false 60-second ceiling. Focused proxy tests cover aliases, cancel routing, and timeout construction.

2. **Queued monitor reporting — backend/history PASS**  
A queued/coalesced reconcile now increments the warning/failure unit, preserves counter invariants, and adds “profile reconcile deferred.” Task history and backend notifications render the result as completed-with-warnings rather than green.

3. **Truthful multi-sibling CLEAR — PASS**  
If an earlier sibling clears and a later sibling fails, the primary remains selected, the 503 names changed and failed accounts, and the actual sibling mutation is journaled. The new three-account regression covers this exact sequence.

4. **Fail-closed account enumeration — PASS**  
Raised, `None`, non-list, empty, and primary-absent fresh account results return 503 before any clear write or journal entry. Parameterized regressions cover the variants.

5. **Public documentation/recovery — PASS**  
CHANGELOG and API docs accurately narrow proxy scope, distinguish zero-write safety 503 from truthful partial-clear 503, document run/cancel timeout ownership, and describe queued monitor history as deferred.

### One frozen-gate caller defect remains

**Manual Run Now still renders deferred/warning work as green success.**  
The monitor returns the established warning envelope: `success=true`, `failed_count>0`, with `message` containing “profile reconcile deferred.” History and backend notifications correctly interpret that as amber/completed-with-warnings. Two direct callers still branch only on `result.success` and call `notifications.success(..., "Task Completed")`:

- `frontend/src/components/ScheduledTasksSection.tsx:444-460`
- `frontend/src/components/TaskEditorModal.tsx:352-375`

A manual run can therefore display green “Task Completed” for a deferred reconcile—the exact frozen rule that no caller may map queued work to success.

**Required correction:** in both callers, use warning/amber notification whenever `failed_count > 0`, preferably surface `result.message`, and add focused regression tests. No new backend status type is required; the existing completed-with-warnings envelope is adequate when consumers honor it.

### One stale internal comment

`backend/routers/m3u.py:1213-1220` says every abort leaves the group in its prior consistent state. The tested partial-clear behavior intentionally allows earlier siblings to remain cleared while the primary remains selected; those siblings are reported and journaled. Correct the comment in the same patch. Public docs and runtime recovery text are already accurate.

### CI status

At posting, head is exactly `7befad160f04935d3b8781d8f72af309c762624a`. All completed checks are green or expected-skipped, but full **Backend Tests** and **Wait for Test Suite** remain `IN_PROGRESS`; GitHub merge state is therefore `BLOCKED`.

### Final gate remains unchanged

After the two direct caller warning fixes and one-line comment correction:

- Run terminal CI.
- Perform one final delta-only check of those exact edits.
- Approve if green.

No additional architectural or hardening requirements will be added to PR #724.

MotWakorb:
## Final delta item addressed — new head `efac4c3e`

Both items from the last frozen-gate review are fixed. Per that review's stated contract ("approve if green" after these edits; no further architectural/hardening requirements), this is the last delta.

- **Manual "Run Now" callers now honor the completed-with-warnings envelope.** `ScheduledTasksSection.tsx` (`executeRunNow`) and `TaskEditorModal.tsx` (`handleRunSchedule`) branched only on `result.success` and showed green "Task Completed" for a deferred (queued) reconcile. Both now emit an **amber warning** surfacing `result.message` whenever `result.failed_count > 0` (regardless of `success`), keeping the plain success path only for `failed_count === 0` — matching the Task History panel and backend notifications. General-correct: any task returning success + failures reads amber in these callers. One focused regression per caller drives the real Run button and asserts warning-not-success on `failed_count:1` / success on `failed_count:0`.
- **Stale internal comment corrected** (`routers/m3u.py` CLEAR-ordering docstring): now describes the truthful partial-clear (an earlier sibling can remain cleared while the primary stays selected; reported + journaled; self-heals on the next normalize sweep; primary never cleared on failure). Comment-only — public docs/runtime text were already accurate.

### Gates (green on head `efac4c3e`, independently re-verified)
- Frontend: **2181 passed** (154 files); `tsc --noEmit`, `vite build`, `eslint --max-warnings 0` all clean
- Focused backend (reconcile-hooks + monitor + subprocess-proxy): **51 passed**
- Backend full suite unchanged (comment-only backend edit; last full run **7852 passed**)
- Version consistency: all three touchpoints at `0.17.6-0153`

Version steady at `0.17.6-0153`; no closing keyword — GH #720 gets a status update and closes on merge. Ready for the final delta-only check of `efac4c3e`.

PR #733, opened 2026-07-24
Auto-purge run_on_refresh_skipped + task start/complete journal noise (gjb01, v0.17.6-0160)

MotWakorb:
## Auto-purge the last two journal noise buckets, configurably (bead gjb01)

Implements the PO decision (2026-07-19): auto-purge `auto_creation/run_on_refresh_skipped` (~8k rows) and scheduled task start/complete (~8.5k rows, ~285/day), with the policy configurable in settings — extending the existing `JournalNoisePurgeTask` (bead uliyr) rather than adding a new mechanism.

### Predicates (allowlist-scoped, verified against the actual writers)
- **`run_on_refresh_skipped`**: `category=auto_creation AND action_type=run_on_refresh_skipped AND user_initiated=false` — sole writer verified (`tasks/channel_pipeline.py`, always non-user-initiated).
- **Task start/complete**: `category=task AND action_type IN (start, complete) AND user_initiated=false`. The `task` category's five action_types were enumerated from `task_engine.py`; **cancel / fail / error are outside the allowlist and survive regardless of age** (anomaly forensics), and **manually-triggered runs are kept** (`user_initiated` guard — operator audit trail). Only routine scheduled start/complete rows purge.

Both toggles default **on** at the established 3-day retention, surfaced on the same TaskEditorModal settings surface as the existing categories.

### Per-category retention: evaluated, not needed
The forensic-value asymmetry the bead flagged is handled **structurally by the predicate** — everything with forensic value (cancel/fail/error, manual runs) is excluded from the purge entirely, and routine scheduled rows are redundant with the `task_executions` table (richer record, kept 30 days). A per-category days field would add settings surface without protecting anything; rationale documented inline.

### Also in this pass (per the bead)
- **Rule-duplicate audit-trail fix**: duplicating a channel-pipeline rule now journals a `create` entry (previously delete-only pairs — an audit asymmetry).

### Tests
Per-bucket purge + extensive **survival tests**: cancel/fail/error at 400 days survive; manual rows survive while identical scheduled rows purge; unknown future action_types survive; `start`/`complete` in *other* categories survive (proves the category+action_type conjunction); toggles off → untouched; existing survival tests hardened to prove survival by allowlist alone. Plus the duplicate-journal test and frontend toggle render/save tests.

### Gates
- Backend full suite: **7877 passed, 13 skipped**
- Frontend: **2205 passed**; `tsc`, `vite build`, `eslint --max-warnings 0` clean
- Version consistency: all three touchpoints at **0.17.6-0160**

### Ops note (found during gating, already healed)
The 2 recurring e2e failures (`test_channels.py::TestChannelGroups`) were **container deployment drift**: a prior partial backend deploy updated `dispatcharr_client.py` but not `services/profile_reconcile.py`, so the live container 500'd the channel-groups endpoints. Redeploying the one stale file + restart fixed it; the full suite then went green. Takeaway: partial backend deploys that skip `services/` can silently break the container.

Version bumped `0.17.6-0159 → 0.17.6-0160`. CHANGELOG entry added.


MotWakorb (commented):
# Team Review — PR #733

**Verdict: CHANGES REQUIRED** (posted as a comment because GitHub does not allow the PR author to submit a formal Request Changes review on their own PR)

**Depth:** Quick  
**Effective tier:** small-team (strictest-wins across home-lab components plus `ci-release`)  
**Participants:** Security, IT Architecture, Project Management, Project Engineering, UX, Code Review, Database, SRE, QA, Technical Writing

## Block — the new destructive purge toggles do not survive restart

The bead and PR explicitly require both new purge policies to be configurable in settings. The UI save updates only the live task instance:

- `frontend/src/components/TaskEditorModal.tsx:829-843` sends the new toggle values.
- `backend/task_registry.py:510-512` applies `task_config` in memory, then `:566-567` calls `sync_to_database()`.
- `backend/task_registry.py:236-286` never writes `ScheduledTask.config`.
- `backend/task_registry.py:152-168` reconstructs tasks on startup without loading/applying `db_task.config`.
- `backend/tasks/journal_noise_purge.py:94-108` therefore restores both new toggles to `True` after restart.

Concrete failure path: operator disables either bucket → save succeeds and the running process obeys → container/app restarts → the setting silently returns to enabled → the next scheduled purge can delete rows the operator explicitly chose to retain.

This is a pre-existing registry infrastructure gap, but it is directly in scope because it prevents this PR from satisfying the approved “configurable in settings” acceptance criterion. SRE, QA, and Technical Writing independently flagged it; QA and Technical Writing reproduced the restart round trip in disposable worktrees, observing persisted `ScheduledTask.config == None` and fresh-instance defaults re-enabled.

**Required correction:** persist task-specific config to `ScheduledTask.config`, hydrate it during `sync_from_database`, and add a restart-roundtrip regression proving both new toggles remain disabled and are passed to the purge call after reconstruction.

## Debate resolved

Architecture raised a Warn that `user_initiated=False` could theoretically include the internal `triggered_by="api"` value. Code Review and DBA traced the production paths: scheduler runs use `scheduled`, while the public UI/MCP/API “Run Now” path funnels through `run_task(... triggered_by="manual")`. With no reachable non-manual operator path shown, the team did **not** retain this as a finding. Expanding provenance semantics would be goalpost moving for this PR.

## Positive observations

- The new delete predicates are narrow category + action-type + provenance allowlists.
- Tests preserve manual task runs, cancel/fail/error, unknown future actions, fresh rows, and same-named actions in other categories.
- All four deletes share one transaction with rollback on failure.
- The duplicate-rule audit entry has router/database coverage.
- UI labels and retention explanations accurately communicate the intended policy while the process remains running.

## Verification

Independent disposable-worktree checks passed:

- 213 relevant backend tests
- 7 focused frontend tests
- frontend typecheck and lint
- `git diff --check`
- exact-head GitHub checks are green, including backend/frontend/MCP tests, Semgrep, CodeQL, version consistency, visual regression, and image builds

No other in-scope Block/Warn/Nit findings. No backlog bead was created: this blocker belongs to the current PR’s acceptance criteria, and filing unrelated follow-up work would move the goalposts.

MotWakorb:
## Blocker addressed — new head `124b5423`

**Task config now survives restart** — persisted generically at the registry level, not purge-specific:

- **Persist:** `_persist_task_config()` snapshots the task's `get_config()` as JSON into `ScheduledTask.config` on every save (both the update and create branches of `_save_task_to_db`). A snapshot failure logs and keeps the prior stored value — never aborts the save.
- **Hydrate:** `sync_from_database()` applies a non-null stored config onto the reconstructed instance via the **same** `update_config` path the live PATCH uses — one application code path. Malformed JSON is logged and ignored; startup never breaks.
- **Back-compat:** merge-over-defaults falls out of the existing `update_config` contract (implementations only touch keys present in the dict), so a stored config missing a newer key keeps the code default; `config == None` (every pre-existing row) skips hydration → exactly today's behavior. **No migration.**

### Blast radius — surveyed all 17 tasks
Three must NOT persist and are opted out via a new `TaskScheduler.persist_config = False` flag, each with an inline rationale:
- `dbas_restore` / `dbas_sync` — `confirm_apply` arms a destructive apply and `artifact_path` is a deleted temp upload; the fail-safe restart direction is disarmed (today's behavior, now deliberate + tested).
- `m3u_digest` — its config is backed by its own settings table also edited directly by `/api/m3u` endpoints; a registry snapshot would go stale and clobber newer settings.

All other tasks hold durable operator settings — for them, today's restart-reset was itself latent data loss (e.g. CleanupTask's retention days also reverted on every restart; this fix heals those too).

One flagged nuance (benign, no destructive task affected): a manual run's per-run parameters are applied via the same `update_config` path and synced post-run, so they now persist across restart instead of resetting — matching what the live instance/modal already displayed. If save-path-only persistence is preferred, it's a small follow-up.

### The required regression (plus 8 more)
`test_disabled_toggles_survive_restart_and_gate_the_purge` — both new toggles set False via the operator save path → fresh registry reconstructed from the DB → toggles **still False** on the distinct new instance, **and** an `execute()` on it leaves the disabled buckets' old rows untouched while the still-enabled bucket purges (proves the config gates the delete). Plus: persist-on-save, persist-on-create, `config=None` legacy defaults, missing-key merge, malformed-JSON tolerance, and 3 opt-out tests (including a hand-planted stored `confirm_apply: true` that is NOT rehydrated — re-disarms on restart).

### Gates (green on head `124b5423`)
- Backend full suite: **7886 passed, 13 skipped**
- Focused (config-persist + journal-noise-purge + task_registry): **53 passed**
- Version consistency: all three touchpoints at **0.17.6-0160** (no re-bump; frontend untouched)

Ready for the delta re-check.

MotWakorb:
## Team delta re-review — CHANGES REQUIRED

Reviewed only `d4c3f59b..124b5423` against the blocker in the prior review. The journal-noise-purge restart requirement itself is now satisfied: explicit task settings are persisted, startup rehydrates them through `update_config`, and the new regression proves both disabled buckets remain disabled and gate deletion after reconstruction.

### Block — per-schedule parameters are now persisted as global restart defaults

The generic persistence path has a wider regression:

- `backend/task_engine.py:847-852` applies the triggering `TaskSchedule.parameters` directly to the shared singleton task instance.
- `backend/task_engine.py:944-945` then calls `registry.sync_to_database(task_id)` after execution.
- New code in `backend/task_registry.py:314-333` snapshots that mutated instance via `get_config()` into `ScheduledTask.config`.
- `backend/task_registry.py:163-180` rehydrates that snapshot on restart.

This changes the scope of schedule-owned inputs. A scheduled `stream_probe` run with `channel_groups=["sports"]` and `timeout=17` becomes the task-wide restart default; later manual/no-parameter runs inherit that scope and timeout. QA reproduced the same leakage for all three affected task types:

- `stream_probe`: `channel_groups` / `timeout`
- `black_screen_scan`: `channel_groups` / `sample_duration`
- `failed_stream_reprobe`: `timeout` / `max_concurrent`

The PR comment says all non-opted-out tasks hold durable operator settings, but these fields are explicitly per-schedule parameters. This blocker is introduced by the chosen generic fix and is therefore inside the original persistence correction, not a new goalpost.

**Required correction:** separate explicit operator/PATCH persistence from execution-time schedule overlays, or opt out every schedule-parameter task and prove the boundary. Add a regression covering scheduled execution → post-run sync → fresh registry, demonstrating that schedule parameters do not populate or rehydrate as global defaults.

### Team debate

Security and SRE also noted that malformed stored JSON falls back to constructor defaults, which re-enables destructive purge buckets. Database and QA judged that unsupported/manual corruption path acceptable at the small-team tier because valid saves serialize JSON, startup logs the problem, and the behavior is explicitly tested. The team is not retaining it as a PR finding.

### Verification

- Required config-persistence suite: **9 passed**
- Persistence + journal-noise-purge suites: **44 passed**
- QA disposable repro: **3 expected failures**, one for each affected schedule-parameter task
- `git diff --check`: clean
- Current GitHub checks are green except Backend Tests and the aggregate Wait for Test Suite, which are still pending

No bead was created: this regression belongs to the current PR correction. No other in-scope findings.

MotWakorb:
## Delta-review block addressed — new head `3990b6eb`

**Per-schedule parameters can no longer become global restart defaults.** Persistence is now gated to the explicit operator save path only:

- `sync_to_database(..., persist_config: bool = False)` — default OFF; gates `_persist_task_config` in **both** the update and create branches of `_save_task_to_db`.
- The **only** runtime caller passing `persist_config=True` is `update_task_config` (the router PATCH path). The post-run sync (`task_engine.py:945`) is unchanged and therefore persist-off: run bookkeeping updates, the stored config snapshot stays untouched.
- Create-branch analysis: the startup fresh-row path persists **pure code defaults** (instance constructed one line above — no overlay can exist); a hypothetical post-run save against a missing row creates it with `config=None` → restart yields clean defaults, never the overlay (tested).
- **Invariant, stated in the docstrings:** the stored snapshot is always "the operator's last explicit save" (or clean defaults for a fresh row); execution-time schedule overlays mutate the live instance for the run but never reach the stored config.

### Bonus real bug found while proving the invariant against the real `StreamProbeTask`
The prior snapshot format had a hydration hazard: `get_config()` reports `channel_groups: None` for "not configured (probe **all**)", but `update_config` coerces an explicit `None` to `[]` ("probe **nothing**") — replaying a full snapshot at restart would silently flip probe-all into probe-nothing. Fixed: None-valued keys are treated as "unset" and **omitted** from the snapshot, so hydration expresses unset by key absence (merge-over-defaults); an operator's explicit `[]` still roundtrips as `[]` (tested).

### Required regression (plus three more)
- `test_engine_run_with_schedule_parameters_does_not_persist_overlay` — the **full engine path**: real `TaskEngine._execute_task(parameters=…)` after an operator save; asserts the overlay reached the live instance for the run, the stored snapshot is exactly the operator save, and a **fresh registry** reconstructs the operator config — overlay gone.
- `test_post_run_sync_keeps_operator_saved_config` (real `StreamProbeTask`, exact engine call sequence), `test_post_run_sync_on_missing_row_does_not_capture_overlay` (create-branch guard), `test_explicit_empty_channel_groups_roundtrips` (None-vs-`[]` pin).
- All 9 prior persistence tests green unchanged; the malformed-JSON fallback untouched per the resolved team debate.

### Gates (green on head `3990b6eb`)
- Backend full suite: **7890 passed, 13 skipped**
- Focused (config-persist + purge + registry + engine files): **61 passed**
- Version consistency: all three touchpoints at **0.17.6-0160** (no re-bump; frontend untouched)

Ready for the final delta check.

MotWakorb:
## Team final-delta review — CHANGES REQUIRED

Reviewed only `124b5423..3990b6eb` against the blocker in the immediately prior team comment. The normal post-run sync is now correctly persistence-off, the missing-row branch does not capture an overlay, and the supplied 13-test persistence suite passes. Two exact persistence-boundary regressions remain.

### Block 1 — a later enabled-only PATCH still persists the schedule overlay

`task_engine.py:847-854` applies schedule parameters to the shared singleton and leaves them live. The new post-run call correctly keeps `persist_config=False`, but `update_task_config()` later calls `sync_to_database(..., persist_config=True)` unconditionally at `task_registry.py:664-667`, even when the PATCH contains no `task_config`.

The standard UI enabled toggle sends only `{ enabled }`. Multiple reviewers reproduced this sequence with the real Stream Probe configuration:

1. Operator explicitly saves `timeout=30`.
2. A scheduled run overlays `channel_groups=["sports"]`, `timeout=17`.
3. Post-run sync correctly leaves the stored operator snapshot unchanged.
4. Operator toggles only `enabled`.
5. Stored config becomes the live overlay, including `channel_groups=["sports"]` and `timeout=17`.

Restart then restores run-scoped values as global defaults. This directly contradicts the delta docstring and PR claim that the snapshot is always the operator’s last explicit config save.

**Required correction:** make persistence source-aware. An enabled/alert/bookkeeping PATCH must not snapshot the mutable live instance; persist only explicit task-config intent or maintain/restore a separate durable operator snapshot. Add the exact overlay → post-run sync → enabled-only PATCH → restart regression.

### Block 2 — snapshots written by the preceding head can hydrate `null` as `[]`

The new write path omits `None` values, but `task_registry.py:176-180` still passes existing stored JSON fields directly to `update_config`. Head `124b5423` wrote complete snapshots containing null-valued keys, so this is real upgrade data from the immediately preceding reviewed implementation, not a hypothetical legacy format.

Reviewers reproduced a stored Stream Probe snapshot of `{"channel_groups": null, "timeout": 30}` hydrating with `channel_groups == []`. For Stream Probe, `None` means probe all/unset while `[]` means probe nothing. The write-side filter cannot repair already-stored rows.

**Required correction:** normalize/omit field-level nulls during hydration or migrate the stored snapshots, and add a regression seeded with JSON containing `"channel_groups": null`—not merely SQL `config = NULL`. Preserve an explicit empty list as `[]`.

### Team agreement

Architecture, Security, UX, Code Review, Database, SRE, QA, and Technical Writing independently identified the PATCH boundary; six personas reproduced it. Security, UX, Code Review, Database, SRE, QA, and Technical Writing independently identified the field-null upgrade issue; six reproduced it. Engineering and PM found the supplied paths sound but did not exercise these sequences. The reproduction evidence resolves the disagreement in favor of retaining both blocks.

The previously resolved provenance and malformed-JSON questions remain closed and were not reopened.

### Verification

- Supplied focused suite: **13 passed**
- Independent enabled-only PATCH repro: failed as described
- Independent stored field-null hydration repro: failed as described
- Current GitHub checks: completed checks pass; Backend Tests and aggregate Wait for Test Suite remain pending

No bead was created: both findings are part of the current PR persistence correction. No other in-scope findings.

MotWakorb:
## Both final-delta blocks addressed — new head `34412d8b`

The fix is structural this time: **the write side never reads the live instance at all.**

### Block 1 — persistence is now intent-driven
- `update_task_config` threads the PATCH's explicit `task_config` dict as `config_intent`; `_persist_task_config` computes `new_stored = strip_nulls({**previous_stored, **explicit_intent})` — **`instance.get_config()` is never called for persistence** (the parameter survives only for the opt-out check).
- An enabled/alert/bookkeeping-only PATCH (`task_config` absent or `{}`) → intent `None` → stored snapshot untouched.
- A partial config PATCH updates only the PATCHed keys — a live overlay cannot leak regardless of PATCH shape, structurally.
- Fresh startup rows now store **`config = NULL`** (defaults by absence — the same contract every legacy row already has), removing the last `get_config()` read from the write side.
- Invariant (docstring): *the stored snapshot is the ACCUMULATION of explicit operator config saves — the merge of PATCHed keys over the previous snapshot — never a reading of the live instance.*
- Router/frontend untouched: `routers/tasks.py` already passes `config.config` as `task_config` (enabled-only PATCH arrives as `None`).

### Block 2 — hydration-side null stripping
- Shared module-level `_strip_unset_config_keys` ("None means unset ⇒ express by absence") used on **both** sides: the write merge (an explicit `None` in a PATCH unsets the stored key) and **hydration**, which strips field-level nulls from loaded JSON before `update_config` — repairing `124b5423`-era snapshots at read time. Explicit `[]` preserved.

### Required regressions (18-test persistence suite)
- **The review's exact Block-1 sequence** on real `StreamProbeTask`: operator save → engine overlay left live → post-run sync → **enabled-only PATCH via the real update path** → stored is exactly the operator save; fresh registry rehydrates probe-all. Also appended to the full real-`TaskEngine` path test.
- Partial-PATCH-while-overlay-live → only the PATCHed key stored, no leak (restart confirmed).
- **Block-2 seed**: a row whose config JSON is literally `{"channel_groups": null, "timeout": 30}` (field-level null, not SQL NULL) → hydrates as `channel_groups=None` (probe-all), `timeout=30`; explicit-`[]` roundtrip stays green.
- Plus: saves accumulate across PATCHes; explicit-None-unsets; fresh-row `config IS NULL`; all prior overlay/post-run/create-branch/opt-out/back-compat tests green. Provenance + malformed-JSON untouched per the closed debates.

### Gates (green on head `34412d8b`)
- Backend full suite: **7895 passed, 13 skipped**
- Focused (persist + purge + registry + engine + router files): **244 passed**
- Version consistency: all three touchpoints at **0.17.6-0160**

Ready for the final delta check.

MotWakorb:
## Code Review: Auto-purge run_on_refresh_skipped + task start/complete journal noise (gjb01, v0.17.6-0160)

### Verdict: Approved

### Summary
Re-reviewed current head `34412d8b` against the complete prior review history and the two blockers retained in the immediately preceding round. Both are resolved without reopening previously closed provenance or malformed-JSON questions, and I found no new in-scope blocking issue in the complete current diff.

### Test Quality
- **TDD Adherence**: Yes for the requested corrections; the regression suite pins the exact previously failing sequences.
- **Coverage Assessment**: The tests now cover operator-save → schedule overlay → post-run sync → enabled-only PATCH → restart, partial config intent while an overlay is live, field-level `null` hydration, explicit `[]` preservation, config accumulation/unset behavior, opt-outs, and destructive purge gating after restart.
- **Findings**: None blocking or non-blocking.

### Correctness
- **Prior Block 1 — resolved**: `ScheduledTask.config` is now updated only from explicit `task_config` intent. Persistence never reads the mutable live task instance, so schedule overlays cannot leak through post-run, enabled-only, alert-only, or partial-config saves.
- **Prior Block 2 — resolved**: hydration strips field-level `null` values before `update_config`, preserving unset/default semantics; explicit empty lists remain persisted and rehydrated as empty lists.
- The original restart-survival requirement remains satisfied: disabled purge buckets stay disabled after reconstruction and gate the delete path.

### Security
- **Scan Results**: Current-head CodeQL and Semgrep checks are green. The automated mixed-import CodeQL comment is informational and the CodeQL check completed successfully.
- **Logic-Level Findings**: PASS for the changed purge allowlists/provenance guards and config persistence boundary. No auth/access-control surface changed in the final delta.

### API Design
- **Contract Compliance**: The existing task PATCH contract is preserved; only explicit `config` keys contribute to durable task config.
- **Findings**: None.

### Maintainability
The intent-driven merge creates a clear persistence invariant, and `persist_config = False` documents the three task surfaces that must not use registry persistence.

### Performance
No material regression identified; persistence adds one small JSON merge only on explicit operator config saves, and purge operations remain bounded SQL deletes in one transaction.

### Style
- **Guide Compliance**: Follows existing Python/React patterns.
- **Findings**: None.

### Review Items

| # | File | Line | Severity | Category | Finding |
|---|------|------|----------|----------|---------|
| — | — | — | — | — | No remaining findings |

### Positive Observations
- The fix addresses the persistence boundary structurally instead of adding caller-specific exceptions.
- The exact prior failure sequences are retained as regressions.
- Previously resolved questions were kept closed; no non-blocking item was promoted into a blocker.
- Current GitHub checks are green, including backend/frontend/MCP tests, CodeQL, Semgrep, version consistency, visual regression, performance benchmark, and image builds.

### Verification
- Focused backend review suite: **231 passed** (`test_task_registry_config_persist.py`, `test_journal_noise_purge.py`, and `test_channel_pipeline.py`, run with coverage disabled because a focused subset cannot satisfy the repository-wide coverage threshold).
- `git diff --check origin/dev...34412d8b`: clean.
- No disposable worktree or reproduction artifact was created; all checks were read-only against the exact current head.

### Style Guide Updates Needed
None.

GitHub does not permit the PR author account to submit a formal APPROVE review on its own PR, so this approval is posted as the structured PR response rather than an `APPROVED` review event.

PR #737, opened 2026-07-25
Fix flaky EPG preview-token tamper test — base64 padding-bit no-op (zfp2z, v0.17.6-0164)

MotWakorb:
## Summary
- Fixes the CI flake in `test_epg.py::TestGuideMigration::test_apply_rejects_invalid_preview_tokens_at_acceptance[tampered]` (202 vs 409) seen on dev push `17075ce3` (Tests run 30066263282).
- Root cause (reproduced deterministically): the test tampered the token by flipping the last base64 char of the HMAC signature. When that char is `A` (~1/16 of tokens), replacing it with `B` only changes padding bits that `urlsafe_b64decode` ignores — the decoded signature bytes are identical, so the 'tampered' token is genuinely valid and the 202 is CORRECT. Measured 6.09% flake rate over 20k tokens, matching 1/16.
- **Not a product validation gap** — `verify_preview_token` is unchanged and correct. The Dispatcharr 'missing http:// protocol' errors in the failing run were downstream symptoms of the correctly-accepted apply running against the unconfigured test client.
- Fix: new `_tamper_preview_signature` helper decodes the signature, flips a byte (`^= 0xFF`), re-encodes — guaranteed-different HMAC bytes.

## Testing
- Fixed test 20/20 in a repeat loop (pre-fix: failure deterministically reproduced by forcing last-char `A`)
- Full CI-equivalent backend suite: 7824 passed, 12 skipped (pre-existing env skips), 2 deselected
- `scripts/check_version_consistency.py` passes at 0.17.6-0164

Bead: enhancedchannelmanager-zfp2z

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -142,7 +142,7 @@ handle authentication automatically when accessed through the web UI.
 Login endpoints are rate-limited to 5 requests per minute per IP address.
     """,
 
-    version="0.17.6-0162",
+    version="0.17.6-0164",
     openapi_tags=tags_metadata,
     docs_url="/api/docs",
     redoc_url="/api/redoc",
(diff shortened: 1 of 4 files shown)


MotWakorb (commented):
## Team review — PASS

Quick review at current head `eeb9fa09` by all ten team personas; calibrated to the home-lab tier.

No blocking findings. The helper mutates decoded HMAC bytes before re-encoding, making the tampered signature deterministically byte-different and eliminating the Base64 padding-bit no-op. Production validation is unchanged and current CI is green.

**Verdict:** ready from team-review scope.

PR #740, opened 2026-07-25
M3U change digest: detect pure renames when group count is stable (wawwh, v0.17.6-0167)

MotWakorb:
## Summary
- `M3UChangeDetector.detect_changes` only recorded stream diffs on group COUNT change; a pure rename (old name out, new name in, count stable — classic PPV slot provider behavior) recorded nothing in M3UChangeLog.
- The count gate was not a perf guard: per-group name sets were already built unconditionally from data fetched every refresh and then discarded for count-stable groups. Fix adds the missing `elif` branch — reuses the already-built sets, records `streams_added`/`streams_removed` pairs only when membership actually changed.
- Fails open like the existing branches: only fires when both snapshots captured names for the group (disabled groups / 500-name cap can't fabricate renames). Marginal cost: one extra set-difference on in-memory data; no new network calls.

## Testing
- 5 new tests: pure rename (count stable), same-names no-op, multi-swap, fail-open with missing names, count-change path unchanged; 10/10 repeat-loop clean
- Full CI-equivalent backend suite: 7829 passed, 12 skipped (pre-existing env skips)
- `check_version_consistency.py` passes at 0.17.6-0167; CHANGELOG entry added

Bead: enhancedchannelmanager-wawwh (found in jqwfq spike)


MotWakorb (commented):
## Team review — CHANGES REQUIRED

Quick review at current head `6552904` by all ten team personas; calibrated to home-lab with the required data-integrity lens.

The basic rename case is good, but three scope-local issues remain:

1. Sets collapse duplicate names. `["Event A","Event A"] → ["Event A","Event B"]` records an addition without the replaced occurrence. Use multiset semantics or stable stream identity and test duplicate-name persistence.
2. The first-500 capture is nonempty but incomplete and has no stable-order/completeness signal. An unchanged 501-stream group whose returned window shifts from `s0…s499` to `s1…s500` is durably recorded as a rename. Only compare demonstrably complete captures (or stable identities), and add a reordered 501+ regression. The current fail-open test covers empty names, not incomplete samples.
3. The changelog incorrectly says this affects jqwfq’s previous-day stale-dateless guard; that guard reads `M3USnapshot.groups_data`, not `M3UChangeLog`. Limit the claim to the change log/tab/digest and state the cap limitation accurately.

Unlike the existing count-change path, the incomplete sample is the only evidence of change here.

**Verdict:** address before merge.

MotWakorb (commented):
## Team re-review — CHANGES REQUIRED

Quick re-review at current head `533b1b8` by all ten team personas, calibrated to the existing small-team effective tier. Scope was frozen to the three findings from the prior review; no new requirements were introduced.

Two findings are fully resolved:

1. **Duplicate names:** the count-stable branch now uses `Counter` multiset subtraction, with focused coverage for duplicate replacement, duplicate-count change, and multiset-identical no-op.
2. **Incomplete capped captures:** the branch now requires both true group counts to be strictly below `MAX_STREAM_NAMES`; tests cover 499 detection, exact-500 silence, and reordered 501-name windows remaining silent.

The jqwfq scope correction is directionally right—jqwfq reads `M3USnapshot.groups_data`, not `M3UChangeLog`, so this pure-rename logging gap did not affect it—but one sentence still overclaims the snapshot behavior:

> `create_snapshot()` writes that column in full on every refresh

That is not accurate. Name capture is enabled-group-only and capped at 500, and the scheduled path can skip the capture sweep when the upstream timestamp is unchanged. Please replace that justification with wording such as:

> The jqwfq stale-dateless guard is a separate code path and was never affected by this `M3UChangeLog` gap: it reads `M3USnapshot.groups_data` directly via `previous_day_names()`, and when a capture sweep runs, `create_snapshot()` stores the independently captured names for enabled groups (subject to the 500-name cap) regardless of whether `detect_changes()` records change-log rows.

That is the only remaining ask.

Verification: multiple reviewers independently ran `backend/tests/unit/test_m3u_change_detector.py` at the reviewed SHA: **43 passed**. Current CI is green and the PR is mergeable.

**Verdict:** correct that one changelog sentence; no further code or test changes requested.

PR #743, opened 2026-07-26
v0.17.6-0170: DBAS closeout — WebDAV enablement, startup key-integrity probe, logo-miss channel links, 12-category producer completion (v0.18 scope, plugins excluded per ADR-012 D10)

MotWakorb:
Four-bead DBAS closeout batch on one branch; single version-bump commit (0.17.6-0170), followed by the PR #743 team-review fixes (four frozen items — all addressed).

## 0i2vt.8 — WebDAV cloud targets enabled end-to-end

**What:** Adds `webdav` to the `provider_type` Literal on CloudTargetCreateRequest/UpdateRequest/TestRequest (and the CloudStorageTarget column comment); frontend gains webdav in ProviderType/PROVIDER_OPTIONS/PROVIDER_FIELDS (base_url/username/password) with icon+label mappings; OneDrive/Dropbox are greyed out as "not yet supported" in the create dropdown (existing rows still render and edit); MCP cloud-target docstrings name the supported set.
**Why:** The WebDAV adapter was already built, SSRF-hardened, and in SUPPORTED_PROVIDERS — only the CRUD validation layer blocked it, so operators could not create WebDAV targets at all. Deferred-provider greying is the PO decision of 2026-07-25.
**Review item 2 (fixed):** TLS policy is now single-source. A validated top-level `insecure: bool` flows through create/update/test requests and persists to `CloudStorageTarget.insecure`; `credentials.insecure` is reserved and rejected (422) at every request surface; the saved-test path strips it from legacy stored rows and overrides from the column — the same override scheduled uploads apply, so test behavior equals upload behavior. Verification skips on test paths write a `cloud_test_insecure_tls` audit row mirroring the upload path's row. The editor exposes the flag as an advanced WebDAV-only checkbox with an explicit interception warning.
**Tests:** WebDAV end-to-end integration test (API create with real credential encryption → saved-target test-connection through the real adapter + SSRF chokepoint → real-adapter streamed upload), Literal parity tests, MCP forwarding test, editor affordance tests, plus the TLS-policy suite: default verification on, explicit opt-in + audit (saved and inline), `credentials.insecure` rejection on all three surfaces, persistence round-trip, legacy-credentials override parity.

## m40pn — Fernet key integrity probed at container startup

**What:** `verify_key_integrity_at_startup()` in `cloud_storage/crypto.py`, invoked from the startup event in `backend/main.py`.
**Why:** Key file mode/ownership violations previously surfaced only at the first scheduled backup. They now log an unmissable ERROR at every container start. Posture is log-loudly-but-boot: the probe never blocks startup. At use time, a decrypt failure reports the credentials as unavailable and the credential-consuming backup operation aborts before any outbound use (review item 4 wording).

## cm9bi — logo-miss details list affected channels

**What:** `LogoMissDetail` gains `channels: list[LogoMissChannel]` (destination Dispatcharr channel id where known + operator-facing name), resolved from archived `logo_id` FKs through the CHANNEL remap namespace (channels import before logos). LogoMissBanner renders nested per-channel rows with Dispatcharr links.
**Why:** Restores the original 0i2vt.19 scope (PO decision 2026-07-25) — a miss should tell the operator which channels are affected. Dispatcharr's SPA has no per-channel URL (channel editing is a modal on /channels — verified against the live bundle), so per-channel links open the /channels list page. Additive optional fields, no CONTRACT_VERSION bump; dry-runs never emit provisional remap ids.
**Review item 1 (fixed):** real artifacts previously lost the source logo id (producer inventoried logo binaries by filename/size only; decoder emitted id-less rows), so the affected-channel list was always empty on genuine backups. The builder now joins each archived logo file to its SOURCE Dispatcharr logo record by URL basename (lowest-id tie-break, mirroring the importer's file match) and carries `id` + display `name` in `binary/metadata.json`; the decoder re-attaches them. A genuine-artifact round-trip regression proves a missed logo reports its archived affected channel — name-only on dry-run, destination id resolved through the CHANNEL remap (5 → 505) on confirmed apply.
**Review item 3 (fixed):** `logo_misses` counts missed LOGOS, so the banner aggregate is now logo-oriented ("1 logo is missing after this restore" / "N logos are…") with the shared-logo regression: one missed logo affecting two channels renders "1 logo" plus two nested channel rows.

## lc6zu — backup artifact producers for user_agents / core_settings / comskip / dvr_rules

**What:** `build_backup_artifact()` now emits the four settings/agents sections (`RESTORABLE_SECTIONS` entries are `dispatcharr + artifact_only`, so legacy YAML export/validate surfaces are unchanged), and the artifact decoder maps them into the restore plan: `user_agents`/`dvr_rules` as entity categories, `core_settings`/`comskip` into the single SETTINGS plan category as `{section, values}` records — the exact contract the settings importer consumes. The restore summary UI now labels the Settings category.
**Why:** The restore importers for these categories were fully wired by PR #742, but no producer wrote their sections — restoring them silently no-opped against a real backup. This completes round-trip coverage of all 12 categories in the v0.18 scope; plugins remain excluded per ADR-012 D10 (review item 4 wording). Comskip finding: Dispatcharr has no separate comskip endpoint; comskip keys live in the core-settings namespace (the importer PATCHes them per-key there), so the producer fetches once and splits by the `comskip` prefix. Redaction: setting values whose key names match the importer's conservative denylist (`is_safe_setting_key`, imported so the lists cannot drift) carry the REDACTED sentinel — always, even on credential-carrying migration exports, since the importer unconditionally skips those keys by name.
**Tests:** TDD red-first: producer tests incl. a core-setting secret sentinel in the whole-artifact byte scans, decoder tests, and the real-apply round-trip feeds these categories from a genuinely produced artifact (the augmented-plan shim for them is removed), including the DVR rule channel-FK remap.

## Gates (post-review-fix)

- Backend: 7968 passed, 4 skipped (full suite).
- Frontend: vitest 2250 passed / 157 files; tsc clean; eslint clean; production build clean.
- `scripts/check_version_consistency.py`: all 3 touchpoints at 0.17.6-0170.


MotWakorb (commented):
## Team review — CHANGES REQUIRED

Quick initial review at `a428aa5` by all ten team personas, calibrated to the effective **small-team** tier. Scope was limited to the four advertised closeout beads and their direct tests/docs/version changes.

### Frozen acceptance matrix

| # | Finding | User impact | Frozen expected behavior / required evidence | Disposition |
|---|---|---|---|---|
| 1 | Real DBAS artifacts lose the source logo ID needed by the new affected-channel lookup | Logo-miss details from actual backups always contain `channels: []`, so the advertised drill-down is nonfunctional | Preserve a stable source logo ID/correlation through build → decode → importer; add a genuine artifact round-trip regression proving a missed logo reports its archived affected channel(s), including destination remap where known | **Fix now — `cm9bi`** |
| 2 | WebDAV TLS policy has two conflicting paths: documented first-class `target.insecure` is absent from create/update/UI, while arbitrary `credentials.insecure` can disable verification for saved/inline tests without the target-level audit policy | Self-signed WebDAV cannot be configured through the documented public flow; an accidental credential key can silently weaken test TLS, and test behavior can differ from scheduled upload | Add validated top-level `insecure` through API/types/editor and derive adapter TLS policy only from it; reject/reserve `insecure` inside credentials; test default verification, explicit opt-in, saved/inline tests, persistence, and scheduled-upload parity/auditing | **Fix now — `0i2vt.8`** |
| 3 | `logo_misses` counts missed logos, but the banner labels the aggregate as channels | One shared missed logo affecting two channels renders “1 channel…”, contradicting the nested rows | Use logo-oriented aggregate copy (`1 logo…` / `N logos…`) while retaining affected channels as nested details, with the shared-logo regression | **Fix now — `cm9bi`** |
| 4 | Release language claims a “13-category round-trip,” but ADR-012 D10 excludes plugins; the shipped v0.18 scope is 12 categories. The crypto wording also overstates where fail-closed behavior lives | Changelog/PR contract is inaccurate | Say all **12 categories in the v0.18 scope** are covered and plugins remain excluded by ADR-012 D10. Clarify that decrypt returns unavailable on failure and credential-consuming backup operations abort before outbound use | **Fix now — `lc6zu` / `m40pn` wording** |

### Evidence

- The artifact producer inventories logo binaries by filename/size only; the decoder explicitly emits logo rows without `id`; `_affected_channels()` returns empty unless that source ID is an integer. QA reproduced this with a built ZIP → decode → restore flow.
- WebDAV request schemas accept an untyped credential dictionary but no top-level `insecure`; saved/inline tests pass that dictionary directly to the adapter, which converts `credentials.insecure` into `verify=False`. The scheduled backup path instead overwrites it from `target.insecure`, confirming the policy mismatch.
- `logo_misses` increments once per missed logo, while `LogoMissBanner` renders the number as channels.
- ADR-012’s original 13 includes plugins, while D10 explicitly removes plugins from v0.18; no plugin producer/importer is shipped here.

Focused reviewers independently passed the relevant existing suites (including 98 code-review tests, 96 DBAS tests, 203 security tests, and QA’s targeted decoder/logo/WebDAV/lifecycle set). Current CI is green and the PR is mergeable; these defects sit at untested contract seams rather than failing existing gates.

### Explicitly cleared / not merge blockers

- Startup key probe: non-fatal as advertised and does not regenerate an existing key.
- `user_agents`, `dvr_rules`, `core_settings`, and `comskip`: producer → decoder → importer shapes and redaction/FK-remap order are symmetric.
- Generic Dispatcharr `/channels` links: accepted disclosed constraint; no per-channel route exists.
- Beads remaining `in_progress` before merge: normal landing workflow, not a code defect.
- No new requirement for enterprise controls, deeper Dispatcharr routing, or unrelated DBAS hardening.

**Verdict:** address the four frozen items above before merge. Subsequent review is delta-only against this matrix plus concrete regressions introduced by the fixes.

MotWakorb (commented):
## Team remediation re-review — ONE FROZEN ITEM REMAINS

Delta-only verification at `d50813a` against the four-item acceptance matrix from the initial review. No previously cleared area was reopened and no new requirement was added.

### Matrix status

1. **PASS — real artifact logo correlation.** The producer now persists source logo `id`/`name`, the decoder restores them, and a genuine build → decode → dry-run/apply regression proves affected-channel reporting plus destination remap (`5 → 505`).
2. **PASS — WebDAV TLS policy.** Top-level `insecure` is validated/persisted/surfaced; `credentials.insecure` is rejected on public request surfaces and safely stripped from legacy saved rows; saved, inline, and scheduled paths share the first-class policy; insecure test paths audit the verification skip; default TLS remains verified.
3. **PASS — logo aggregate copy.** The banner now counts logos, while shared-logo channel impact remains in the nested rows; the one-logo/two-channel regression passes.
4. **PARTIAL — release/crypto wording.** The changelog, PR body, implementation comments, and crypto wording correctly say 12 v0.18 categories with plugins excluded and accurately describe decrypt-unavailable/consumer-abort behavior. However, the current PR title still says:

   > `13-category producer completion`

   That directly conflicts with the frozen expected release wording and is the only remaining change: update the title to say **12-category** (or equivalent wording that does not claim 13).

### Verification

- Code review: focused backend suites passed; 28 additional upload/crypto tests passed; targeted frontend suites passed **29/29**.
- QA: focused artifact/logo/WebDAV suites passed **65/65**; frontend suites passed **29/29**; TypeScript and `git diff --check` passed.
- Security: fresh request-surface/TLS reproduction passed; focused WebDAV/upload policy suites passed **20/20**; no remaining security finding.
- Current CI is green and the PR is mergeable at the reviewed SHA.

**Verdict:** CHANGES REQUIRED only for the stale `13-category` PR title. All code/test findings are closed; no further code changes requested.

PR #752, opened 2026-07-27
v0.18.1-0003: per-target sync concurrency — one task per SyncTarget (bead enhancedchannelmanager-7ipq2.3)

MotWakorb:
## [enhancedchannelmanager-7ipq2.3] Per-target sync concurrency (ADR-013 S6)

> **STACKED on #751** (`feat/7ipq2-live-validation`, which is stacked on #750). Merge #750 → #751 → this, in order. **HOLD — do not merge without PO sign-off** (overnight run; PRs held for morning review).

### What

Each `SyncTarget` now owns its own registered task `dbas_sync_<target_id>` (dynamically registered), replacing v1's single parameterized `dbas_sync` id. This implements ADR-013 S6 verbatim: *"One task_id per SyncTarget (distinct targets run concurrently; the ALREADY_RUNNING guard excludes a second run of the same target)."*

- **Different targets sync concurrently** — a slow/unreachable B no longer starves the others. (v1 also silently *swallowed* the second target's run when two schedules came due in the same tick: the engine groups due schedules by task_id, runs the first schedule's parameters, and advances `next_run_at` for all of them.)
- **Same target never runs twice at once** — the engine's per-task_id guard is now a true per-target lock; refusal is a non-silent `ALREADY_RUNNING` TaskResult (manual) / retry-next-tick (scheduled).
- **Bounded**: `ECM_SYNC_MAX_CONCURRENT` (default 3, floor 1) caps simultaneous sync runs across targets; excess runs queue, never dropped.
- **One-shot arming (7ipq2.2) preserved under concurrency**: per-target registry instances remove cross-target parameter leakage structurally; the disarm resets to the *bound* target id; a run parameter naming a different target hard-fails (`BOUND_TARGET_MISMATCH`) without reaching `run_sync`.
- **Lifecycle**: startup reconcile registers all targets, migrates legacy `dbas_sync` schedule rows to their per-target id (rows referencing a deleted target are disabled with a WARN), and prunes stale per-target rows; SyncTarget create/rename/delete hook the registry best-effort.
- **Authz**: `is_privileged_task_id()` prefix-matches `dbas_sync_`, so per-target ids inherit the O8TBV-1 admin gate (endpoint-tested).
[...]

PR #753, opened 2026-07-27
v0.18.1-0004: pin entrypoint interpreter to /opt/venv + self-diagnosing preflight + config-mount warning (beads enhancedchannelmanager-0oi96, enhancedchannelmanager-ebl4n)

MotWakorb:
## v0.18.1-0004 — entrypoint interpreter pinning + config-persistence warning

Two commits, one field case (2026-07-27, operator down on 0.18.0). Read them apart:

1. `eb29b24` — bead `enhancedchannelmanager-0oi96` (P1): pin the interpreter, make preflight self-diagnosing.
2. `e51e0b6` — bead `enhancedchannelmanager-ebl4n`: warn when `CONFIG_DIR` is not a mount.

### The operator-visible symptom

```
✓ Python 3.12.13 found
✗ Required Python packages missing
...
ModuleNotFoundError: No module named 'fastapi'
```

Container never starts. Nothing in the message says which Python that was or why it lacked the packages.

### Repro (against the published image, no build needed)

```bash
docker run --rm -e PATH=/usr/local/bin:/usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin \
    ghcr.io/motwakorb/enhancedchannelmanager:latest
```

Reproduces byte-for-byte on `dev`. To see the fix without rebuilding, bind-mount this branch's entrypoint over the image's:

```bash
docker run --rm -e PATH=/usr/local/bin:/usr/local/sbin:/usr/sbin:/usr/bin:/sbin:/bin \
    -v "$PWD/backend/entrypoint.sh:/app/entrypoint.sh:ro" \
    ghcr.io/motwakorb/enhancedchannelmanager:latest
```

### Root cause

The image is sound. `backend/entrypoint.sh` resolved `python3` and `uvicorn` purely through `PATH`, and ECM's dependencies live in the image virtualenv at `/opt/venv`. Any container whose `PATH` no longer contains `/opt/venv/bin` silently ran the Debian system interpreter — which reports a reassuring `Python 3.12.13` while having none of ECM's packages.

The real-world trigger is not a hand-edited compose file. An "update/recreate" in Portainer, Unraid or Watchtower rebuilds the container from the new image while **copying the old container's environment onto it**. Any container created before 0.16.0 (when `/opt/venv` was introduced) therefore inherits a pre-venv `PATH` and breaks on its next update. The reporting operator's container had `ECM_VERSION=0.8.4-0007` and a pre-venv `PATH` on a current image filesystem; their own check confirmed `/opt/venv/bin/python3` imports `fastapi` fine.

### Commit 1 — `0oi96`

- `ECM_PYTHON` / `ECM_UVICORN` resolved **once** near the top of `entrypoint.sh`, pinned to `/opt/venv/bin/{python,uvicorn}`. Both overridable; both fall back to a `PATH` lookup when the pinned path is not executable, so images without the venv are unaffected.
- `"$ECM_PYTHON"` at all 7 former bare-`python3` sites: `check_python` version + import probes, `check_application` import + traceback re-run, `check_tls_config`'s two JSON reads.
- `exec gosu appuser "$ECM_UVICORN"` — `uvicorn` exists **only** in `/opt/venv/bin`, so fixing preflight alone would have moved the same failure one line down.
- Dockerfile `HEALTHCHECK` interpreter pinned the same way.
- `check_python` is now self-diagnosing. Success names the interpreter; either failure branch prints:

```
✗ Required Python packages missing from /usr/local/bin/python3
→ Interpreter diagnostics:
→   resolved python : /usr/local/bin/python3
→   resolved uvicorn: /opt/venv/bin/uvicorn (NOT EXECUTABLE — the app launch would fail here too)
→   sys.prefix      : /usr/local
→   PATH            : /opt/venv/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
→   /opt/venv/bin/python: MISSING — the image virtualenv is not where ECM expects it
→ ECM installs its Python packages into the image virtualenv at
→ /opt/venv. If you override PATH or bind-mount over /opt/venv,
→ that virtualenv is bypassed. Remove the override, or point
→ ECM_PYTHON/ECM_UVICORN at the interpreter that has ECM's deps.
```

That block alone would have closed the field case from the first log paste instead of a round trip.

### Commit 2 — `ebl4n`

While diagnosing the above, `docker inspect --format '{{json .Mounts}}'` on the operator's container returned `[]` — no bind mount, no named volume. `/config` (settings, `journal.db`, logos, TLS material) lived only in the container's writable layer, while preflight said `✓ Config directory exists` / `✓ Config directory is writable`. True, and dangerously reassuring: the recreate that fixes their startup problem would also have wiped their data.

New `check_config_persistence()` reads `/proc/self/mountinfo` (field 5 = mount point) — the one place both bind mounts and named volumes appear from inside the container — with an exact match plus an ancestor-prefix rule, so `/config-backup` does not satisfy the check for `/config` while a volume mounted at a parent of `CONFIG_DIR` correctly does. Unreadable mountinfo stays silent rather than raising a false alarm. Advisory by construction: called outside the `FAILED` chain, always returns 0.

### Verification

- **Container level, against the published image** with this branch's entrypoint bind-mounted in (no local image build was needed, and none was done):
  - repro `PATH` → preflight passes, `✓ Python 3.12.13 found at /opt/venv/bin/python`, app reaches `Uvicorn running on http://0.0.0.0:6100` and `Application startup complete`.
  - `/opt/venv` shadowed by an empty mount → the diagnostics block above prints and names the cause.
  - no mount → the full `DATA IS NOT PERSISTENT` warning prints, startup continues.
  - `-v <volume>:/config` → single `✓ Config directory /config is on a mounted volume` line, no warning. Test volume removed afterwards.
- **Not verified at container level:** the Dockerfile `HEALTHCHECK` change (it requires an image build; the change is a literal path substitution to `/opt/venv/bin/python` and is pinned by a text test).
- **Tests:** 17 new tests across `backend/tests/unit/test_entrypoint_interpreter.py` (8) and `backend/tests/unit/test_entrypoint_config_persistence.py` (9). Both execute the *real* code out of `entrypoint.sh` in a POSIX shell — the same extract-and-execute pattern as the existing `test_event_loop_selection.py`, whose launch-tail harness is updated for the resolved-uvicorn exec line. Every one of the 17 was proven red against the pre-fix script; the headline interpreter test reproduces the operator's log verbatim on failure. The persistence tests run against `mountinfo` files **recorded from real containers** of the published image (no-mount / named volume / bind mount / mounted parent), not hand-written shapes.
- **Gates:** full backend suite green in the worktree (exit 0, coverage 76.24% vs the 56% floor); `sh -n` clean on `entrypoint.sh`; version consistency + advance checks pass at `0.18.1-0004`. No frontend source touched (only the `package.json` version stamp), so vitest was not run.

### Deployment

- Rollback: revert the two commits — no schema, no API, no data migration.
- Operators already broken by this should still recreate their container from compose rather than using their update tool's in-place update — but read the new mount warning first and take a backup if their `/config` is unmounted.


MotWakorb:
## Team Review — PR #753 (10 personas, quick depth)

**Effective tier: small-team.** `ecm-core`/`dev-tooling` are home-lab, `ci-release` (Dockerfile + published ghcr.io image) is small-team — strictest-wins per `COMPONENTS.md`.

**Verdict: one mechanical blocker. The fix itself is correct and independently proven.**

The diagnosis in this PR is genuinely good and it holds up under adversarial checking. The project engineer reproduced the field failure byte-for-byte inside the running image and then proved the new entrypoint survives the same conditions — including container-level verification the PR body honestly disclaims:

```
=== OLD (dev) preflight, REAL image, STALE PATH ===
✓ Python 3.12.13 found
✗ Required Python packages missing          <-- field report, reproduced exactly

=== NEW preflight, REAL image, STALE PATH ===
✓ Python 3.12.13 found at /opt/venv/bin/python
✓ FastAPI and Uvicorn available
WOULD-EXEC: /opt/venv/bin/uvicorn
```

Also confirmed independently: under the stale PATH, `uvicorn` is not on PATH **at all** — so the PR's claim that pinning only the checks would move the failure one line down is true, not rhetorical. A straggler grep found no remaining bare `python`/`uvicorn`/`pip` in the image's runtime path (`Dockerfile.dev` is correctly untouched — it installs `--system` and never runs `entrypoint.sh`). Gates re-run against the branch head, not taken on trust: `sh -n`/`dash -n`/`bash -n` clean, 46 entrypoint tests pass, 44 version-consistency tests pass, and `test_event_loop_selection.py` passes 29/29 on **both** branch and `dev` — the harness edit is a faithful accommodation, not a weakening.

---

### 🔴 BLOCKER — the branch is stale and cannot merge

`gh pr view 753 --json mergeable` → `CONFLICTING` / `DIRTY`. Confirmed with `git merge-tree`:

```
CONFLICT (content): Merge conflict in backend/main.py
CONFLICT (content): Merge conflict in backend/routers/backup.py
CONFLICT (content): Merge conflict in frontend/package.json
```

The branch forked before #752 landed and bumped `dev` to `0.18.1-0003` (commit `12736f22`). This PR's version literals say `-0002 → -0004`, colliding on the identical lines. **`-0003` was not skipped or burned** — it is a real merged dev build, so after rebasing the correct bump is `-0003 → -0004` and no number is wasted.

⚠️ All required checks show green, but they ran against the stale merge-base. **Green checks here are not evidence of mergeability.** Rebase, resolve the three version literals, reconcile the CHANGELOG insertion point, re-push, and re-verify `mergeable: MERGEABLE` before merging.

While you're in the CHANGELOG: `[Unreleased]` now carries **two separate `### Fixed` headings** (lines 9 and 17 at the branch head) with an `### Added` between them. A reader who scans by section reasonably stops at the first one. Worth merging into one block per Keep a Changelog, which the file's own preamble claims to follow.

---

### Recommended to land during that rebase — you're touching the branch anyway

**1. Preflight can go fully green and the container still dies at `exec`.** *(security-engineer, code-reviewer, and orchestrator all reproduced this independently)*

`check_python` probes `$ECM_PYTHON --version` and `import fastapi, uvicorn` — the *module*. Nothing asserts `[ -x "$ECM_UVICORN" ]`:

```
✓ Python 3.12.3 found at /usr/bin/python3
✓ FastAPI and Uvicorn available
PREFLIGHT_RC=0
would exec: [/opt/venv/bin/uvicorn]
>>> uvicorn NOT EXECUTABLE — exec would fail here <<<
```

This is the PR's own thesis with a door left open one line down. The realistic trigger is the operator this PR is written for: they follow the new diagnostics, set only `ECM_PYTHON`, and get an all-green preflight followed by a bare not-found and zero diagnostics. `print_python_diagnostics` already computes the answer — the string `"(NOT EXECUTABLE — the app launch would fail here too)"` at `:89` can never fire, because the function is reachable only from failure branches. Four lines in `check_python` closes it.

**2. An explicit `ECM_PYTHON` override that's wrong is silently discarded.** *(code-reviewer, architect, engineer — proven live in `ecm-ecm-1`)*

`[ ! -x "$ECM_PYTHON" ]` can't distinguish "the default pin is absent" from "the operator typed this and typo'd it." `ECM_PYTHON=/opt/venv/bin/pyhton` silently resolves something else with no log line. Self-inflicted loop: the diagnostics tell them to set `ECM_PYTHON`, they fat-finger it, and get back the identical message that sent them there, naming a path they never typed. Capture `${ECM_PYTHON:+1}` before the default expansion and warn when discarding.

Both are tracked in **`enhancedchannelmanager-wpzo6`** if you'd rather ship the hotfix and follow up — that's a PO call, not a reviewer condition.

---

### Filed as beads — NOT merge conditions

These were dispositioned as follow-ups by the reviewing personas. Recording them here so the disposition is explicit and doesn't resurface as a late blocker.

| Bead | Finding | Flagged by |
|---|---|---|
| `6gwju` | `HEALTHCHECK` hard-pins `/opt/venv/bin/python` with no fallback or `ECM_PYTHON` honoring, while the entrypoint has both. An `ECM_PYTHON` user gets a working container reported `unhealthy` forever — and `docker-compose.mcp.yml:16-18` already has `condition: service_healthy`, so the MCP sidecar would silently never start. Unverified at container level (CI builds the image but never runs the healthcheck). | SRE, security, code-reviewer, engineer, architect (5/10) |
| `bvic4` | `config_dir_is_persistent` ignores filesystem type, so `--tmpfs /config` prints `✓ data persists across recreates` — false, and the *reassuring* direction, which is the exact failure shape `ebl4n` exists to close. Separately, the kernel octal-escapes whitespace in field 5, so a mounted `/my config` gets the full false alarm; the code comment names that hazard and then doesn't handle it. | SRE, security, code-reviewer, QA, engineer + orchestrator |
| `mos0l` | QA injected `$ECM_PYTON` (a typo expanding to empty) at the `check_tls_config` site and **all 46 tests still passed** — `check_application` and `check_tls_config` sit outside the harness cut point and have no behavioral coverage. The `-c\|--version`-only regex guard also passes `python3 -m …` and even a hardcoded `/usr/local/bin/python3 -c`, i.e. the bug itself. | QA (by injection), code-reviewer, engineer |
| `nz8q4` | `CONFIG_DIR` is honored in 1 of 4 places — `check_filesystem` and `check_tls_config` still hardcode `/config`. The new check now *endorses* `CONFIG_DIR=/data` (`✓ … data persists`) for a config the entrypoint never chowns, so the app fails after preflight said fine. Also: `ECM_MOUNTINFO` is a test seam promoted to permanent public API on a `:latest` image, and it fails open — setting it silently disables real detection. | architect |
| `v7n0d` | The stale-inherited-environment *class* is untouched. `ECM_VERSION=0.8.4-0007` is being reported from `/api/version`, `/api/health` and the build-info metric by affected containers **right now**. Plus the one-line `export PATH` repair, which also covers the `ffmpeg`/`ffprobe` lookups the pin doesn't reach. | architect, engineer |
| `mevqr` | Three new operator-facing env vars documented nowhere — while `print_python_diagnostics` prints two of them at the operator. No runbook for the failure mode this PR exists to fix; no recovery procedure for the data-loss scenario the new warning announces. | tech-writer, architect, engineer |
| `4ysp0` | The 12-line warning prints on every start indefinitely for a legitimately-unmounted container (alarm fatigue on the one message that means irreversible data loss); diagnostics use the routine-info channel for the actionable fix. | UX, SRE |

---

### Cleared on inspection — closed, please don't reopen these

- **Data-durability claim is accurate and complete.** Every category the warning names — `settings.json`, `journal.db` (+ its `-wal`/`-shm` sidecars), logos, m3u uploads, TLS material — verified rooted under `CONFIG_DIR`. Nothing ECM persists lives outside it.
- **"Back up first (Settings → Backup & Restore)" is safe, not a trap.** The UI button calls `GET /backup/create`, which streams the zip straight to the HTTP response and lands it in the operator's own Downloads folder — never the doomed writable layer. (`POST /backup/save` *does* write into `CONFIG_DIR`, but it's wired to no button and no scheduled task.)
- **No new code-execution or privilege exposure.** `ECM_PYTHON`/`ECM_UVICORN`/`ECM_MOUNTINFO` are settable by the same principal who can already set `--entrypoint`, `--user 0`, or `-v /:/host`. `"$ECM_UVICORN"` is quoted and lands as `argv[0]` of `gosu` — no field-splitting, no flag injection, and it does not need the whitelist treatment `ECM_UVICORN_LOOP` gets. Net, pinning absolute paths is a small **hardening**: it removes the pre-existing case where a writable directory earlier on `PATH` could redirect a root-context interpreter invocation during preflight.
- **No secret leakage.** `print_python_diagnostics` emits paths, `sys.prefix` and `PATH` — nothing secret-bearing. `check_config_persistence` prints only fixed text plus `${CONFIG_DIR}`; mountinfo content never reaches the log. The committed fixtures carry overlay digests, container IDs and a volume name — judged non-sensitive at this tier, no scrub requested.
- **Advisory-never-fatal is the right call and the security engineer defends it.** This is a heuristic over a kernel interface whose shape varies across runtimes; making it fatal would make availability depend on a check that can't be proven correct. Fail-open is correct. (`bvic4` is the one case fail-open reasoning doesn't cover.)
- **Bundling `ebl4n` with `0oi96` was right.** `ebl4n`'s own description pre-declares it: *"Pairs with 0oi96 (same file, same field case) — consider shipping together if 0oi96 has not merged yet."* Pre-declared at bead creation, not scope drift found in review.
- **Shell/test-host parity is fine.** Host `/bin/sh` and container `/bin/sh` are both dash (`python:3.12-slim`). Not a hidden gap.
- **Entrypoint pinning is the right architectural layer.** `ENV ECM_PYTHON=…` in the Dockerfile was considered and rejected on merit: it would fix today's cohort while manufacturing the identical bug for the next venv move, since the recreate flow would then copy a stale `ECM_PYTHON` forward exactly as it copies stale `PATH` today.

---

### Credit

The root-cause analysis is the strongest part of this PR — the Portainer/Watchtower env-copy mechanism is the kind of thing that normally goes undiagnosed, and identifying that `uvicorn` exists *only* in the venv (so pinning the checks alone would relocate the failure) is the insight that makes the fix complete rather than cosmetic. The tests execute the *real* shell out of `entrypoint.sh` rather than reimplementing it, the mountinfo fixtures are recorded from actual containers instead of hand-drawn, the no-pipe fix in `check_python` correctly stops a missing interpreter from reading as success, and the two text guards make the duplicated `/opt/venv` literal *safe* by failing CI if either copy moves alone. The PR body's disclosure of what was **not** verified is exactly the discipline that makes a review like this cheap to run.

---

**Merge boundary, frozen:** the rebase is the only blocker. Findings 1 and 2 above are recommended-while-you're-there; the seven beads are follow-ups and will not be re-raised as merge conditions on a later pass.

🤖 Team review via [Claude Code](https://claude.com/claude-code) — personas: security, architect, PM, engineer, UX, code-reviewer, DBA, SRE, QA, tech-writer

PR #765, opened 2026-08-03
[enhancedchannelmanager-q6xjl] Fix DBAS restore settings PATCH 404 and User-schema YAML parse

MotWakorb:
## Summary

Fixes the two ECM-side Dispatcharr API contract bugs that make a same-instance DBAS restore apply fail 7/7 and roll back (bead `enhancedchannelmanager-q6xjl`, found in doc-test run 4):

- **A — settings PATCH used the wrong lookup identifier.** Dispatcharr 0.28.2's `CoreSettingsViewSet` uses the default integer-`pk` detail route; ECM PATCHed `/api/core/settings/{key}/` (404). Now: one `GET /api/core/settings/` per apply run builds a key→id map (run-scoped lazy `CoreSettingIdResolver`, shared by `core_settings` and `comskip`), PATCHes `/api/core/settings/{id}/`, and a key absent on the destination fails that key explicitly (`DEPENDENCY_UNRESOLVED`) instead of via a raw 404.
- **B — `/api/schema/` is YAML by default** (drf-spectacular); ECM called `.json()` on it, so the users capability check raised `UsersCapabilityError` on every restore. Now requests `?format=json`.

Also bumps version to 0.18.1-0014 (CI build-advance gate) + CHANGELOG entry.

## Verification

- TDD: 11 tests red before the fix, green after. Fixtures are **recorded slices of the live Dispatcharr 0.28.2 responses** (OpenAPI document + settings list shape), per the real-fixtures rule.
- Full backend suite: 8131 passed, 4 skipped; coverage 76.39% (threshold 56%). Re-run independently by the orchestrator: green.
- Read-only in-container proof on the test instance (`ecm-doctest` → live Dispatcharr 0.28.2): all 7 core_settings keys resolve from a single list GET; key-string detail route 404s while integer-id route 200s; users capability check now passes. No write requests issued.
- **Pending (PO-authorized, post-review):** one live restore apply from `ecm-doctest` as the end-to-end round-trip proof.

Follow-on beads (out of scope here): `lsa0s` (dvr_rules remap), `zt3kf` (degraded backup severity), `y6zg6` (dry-run WOULD-FAIL), `d20jf` (upload-path confirm dialog).


MotWakorb:
**Code review (AI team code-reviewer, relayed by orchestrator): Approved with notes — no blocking findings.**

Verified: contract pins are real (reverting either fix breaks specific tests — `?format=json` param assertion; integer-id PATCH set assertions); one-GET-per-apply-run enforced by construction and pinned at orchestrator-integration level; denylist gates before id resolution and key names never reach log sinks; version bump/CHANGELOG touchpoints correct; scope clean vs follow-on beads `lsa0s`/`zt3kf`/`y6zg6`/`d20jf`.

Non-blocking findings (dispositions tracked with the PO):
1. **Warn** — `get_core_setting_id_map` reads only page 1 of a DRF-paginated envelope; `next` never followed (unreachable on 0.28.2's 7-row bare list; fail-safe direction).
2. **Warn** — no dedicated test for PATCH-failure-after-successful-resolve on the settings categories (pre-existing gap; helper support exists).
3. **Nit** — duplicate destination keys last-write-wins silently.
4. **Note** — single-key settings failure still rolls back the whole restore (unchanged behavior; `zt3kf` territory for cross-instance drift).
5. **Nit** — `DEPENDENCY_UNRESOLVED` reused for "key absent on destination".

**Live end-to-end proof (PO-authorized, post-review, reviewed build byte-verified in container):** fresh same-instance DBAS backup → upload → preview (Settings 7 WILL UPDATE) → apply: **"Restore complete", Settings 7 UPDATED / 0 FAILED, `outcome=success`, `created 0, failed 0 across 12 categories`** — the exact run-4 failure signature (7×404 → full rollback) is gone.

PR #768, opened 2026-08-03
[enhancedchannelmanager-lsa0s] Remap dvr_rules to Dispatcharr's recurring-recording-rules resource

MotWakorb:
## Summary

Fixes the last known-bad Dispatcharr path (bead `enhancedchannelmanager-lsa0s`): `/api/dvr/rules/` doesn't exist on 0.28.2 (SPA catch-all → JSONDecodeError → every backup silently exported a `_warning` stub for this category).

**Premise correction found during implementation**: the bead's named target `/api/channels/recordings/` was also wrong — recordings are per-instance state (absolute timestamps, source-disk media references), not rules. The real rules resource is **`/api/channels/recurring-rules/`** (`RecurringRecordingRule`: optional name, one channel FK, weekday/time/date window), which matches the category's existing contract exactly — so `EntityType.DVR_RULE`, the `dvr_rules` artifact key, and the frontend label are all unchanged: no artifact migration. Series rules deliberately excluded (already carried inside `dvr_settings` via core_settings — double-apply hazard); recordings deliberately excluded (instance state).

Also: unnamed-rule identity (schedule-fallback key prevents repeat-restore duplication), dry-run/apply parity for unresolvable channel FKs (same `DEPENDENCY_UNRESOLVED` reason+message both modes), paginated-envelope normalization, corrected the stale comskip comment (endpoint exists but serves `{path, exists}` + multipart upload — not a backup source).

Version 0.18.1-0017 + CHANGELOG.

## Verification

- Red-first at the only layer a wrong path is provable: real `DispatcharrClient` over `httpx.MockTransport` reproducing 0.28.2 routing incl. the SPA catch-all — pre-fix reproduces the exact `_warning` stub symptom. Recorded live fixtures (rules list, dead-path response, OpenAPI slice), operator-data-free.
- Backward compat pinned: pre-fix artifacts with `_warning` stubs decode to an empty category, no traceback.
- Full suite 8157 passed (engineer + independent orchestrator re-run, both exit 0); fake-test guard, version consistency (0017 ×3), build-advance all green.
- Live read-only proof in the test container: `get_dvr_rules() → []` against real Dispatcharr 0.28.2.


MotWakorb:
**Review record (AI code-reviewer, relayed): Approved.** Round 1 (f1d20d86): approved w/ 2 Warns — 'enabled' absent from unnamed-rule identity key (silent collapse of distinct rules); no populated live capture for days_of_week wire shape. Premise correction independently re-derived from Dispatcharr model source (recurring-rules is the real target; series-rules double-apply and recordings instance-state exclusions verified); provisional-remap parity claim traced real; importer ordering structurally shared; /api/dvr/rules/ confirmed never-existent across tags v0.10.0→0.28.2. Round 2 (e45bdcc3) delta-verified: enabled-in-key closed with red-proven test (worktree revert), would-be-state normalization endorsed; fixture gap documented precisely rather than laundered into a fake capture. CI green both heads.

PR #771, opened 2026-08-03
[enhancedchannelmanager-d20jf] Gate upload-path DBAS restore apply behind a type-to-confirm dialog

MotWakorb:
## Summary

Closes the safety asymmetry doc-test run 4 hit live (bead `enhancedchannelmanager-d20jf`): the saved-backups restore path gates "Apply these changes" behind a `TypeToConfirmDialog`, while the upload path applied immediately on one click. The guardrail was never present there (not removed — `git log -S` on the component is empty), so `verify-a-backup.md`'s documented confirmation step was accurate for one of two paths.

- Reuses `TypeToConfirmDialog` as-is (no fork, no prop changes; the saved modal and its tests are untouched).
- Confirmation token is the uploaded `File.name` — the upload-path analogue of the saved path's artifact filename, and the identifier an operator would double-check before overwriting config.
- **Preview (dry run) stays one-click** — it makes no changes.
- **Second unguarded path also gated**: the configure-step "Apply" (which let an operator skip the dry run entirely with one click, behind only a static warning banner) now routes through the same dialog. Leaving it would have left the identical defect class live in the component being fixed. Its trigger reads "Apply restore…" so two buttons don't both read "Apply restore" while the dialog is open; the dialog's own confirm button matches the saved modal exactly.

Version 0.18.1-0020 + CHANGELOG.

## Verification

Red-first (2 of 9 new/existing tests failed against the unmodified component with the expected shape). Gates run by the engineer and independently re-run by the orchestrator: vitest (2508 tests, 0 failed), ESLint `--max-warnings 0` clean, `tsc --noEmit` clean, version consistency (0020 ×3) + build-advance green. Built and deployed to the test container only (`ecm-doctest`, never production) — served bundle contains the new dialog markup, container healthy.


MotWakorb:
**Review record (AI code-reviewer, relayed): Approved, no blockers.** Guardrail completeness verified by full path enumeration — both apply buttons, the dialog's own confirm, Enter-key submit, Escape/Cancel/backdrop dismissal, and stale-file-swap all traced; `start(true)` is reachable only via the gated confirm. Token comparison confirmed identical to the saved path (same shared component, strict equality); preview/dry-run proven still ungated by an explicit negative assertion; `TypeToConfirmDialog` byte-identical to origin/dev and `DbasRestoreSavedModal` untouched. Scope expansion to the second unguarded configure-step apply endorsed as necessary. Two Nits carried to backlog (one test-symmetry assertion; `role=dialog`/aria attrs missing on the shared dialog — pre-existing since build 0019, deserves a house-wide modal audit rather than a rider). CI green, CLEAN.

PR #772, opened 2026-08-03
docs: correct stale Dispatcharr API docs and add restore-runbook maintenance section

MotWakorb:
## Summary

Two beads, one PR — closing the documentation half of the DBAS defect sweep.

**`enhancedchannelmanager-r9oqx` — the developer doc whose staleness seeded three bugs.** `docs/dispatcharr_api.md` claimed the schema lived at `/swagger.json`; it's `GET /api/schema/` (drf-spectacular, **YAML by default** — `?format=json` required, which was fix B of q6xjl). Now also documents what the incident proved: `/api/core/settings/{id}/` is keyed by **integer pk**, not the setting key string (the 404 root cause), plus ECM's list→key→id resolution pattern, bare-list vs paginated-envelope handling, the 20-page cap, and the drop-don't-guess rule; and that `/api/dvr/rules/` has no route on any Dispatcharr version (real resource: `/api/channels/recurring-rules/`; series rules live in the `dvr_settings` blob; comskip-config exists but isn't a backup source). Adds a "how to verify a path against the live schema" note citing the three recorded fixtures as in-repo source of truth. `docs/architecture.md` repeated the stale reference — corrected.

**`enhancedchannelmanager-nvhg7` — the unenforced commitment.** `COMPONENTS.md` calls dbas-backup "highest-care: restore must stay periodically exercised" with nothing prompting it (flagged independently by PM, SRE, and QA in the team review). `docs/runbooks/disaster-recovery-restore.md` gains a **Recurring maintenance** section: monthly dry-run preview, what to check in the report (including the new y6zg6/zt3kf signals), file a bead on surprise. Home-lab tier — documentation, not enforced infrastructure — with the future CI-automation path noted, not committed.

Version 0.18.1-0021 + CHANGELOG.

## Verification

Every claim checked against the recorded fixtures and current client code before writing. `verify-a-backup.md` deliberately left as-is per the PO decision recorded on `d20jf`. Gates run by the writer and independently re-run: `mkdocs build --strict` clean, `npm run docs:check` (fragment/anchor validator) passed, version consistency (0021 ×3) + build-advance green.


MotWakorb:
**Review record (AI code-reviewer, relayed): Approved.** Round 1 (1a94be29): the r9oqx half verified byte-for-byte against `dispatcharr_client.py`, `settings_agents.py`, `routers/backup.py` and all three recorded fixtures — every claim (schema `?format=json`, integer-pk detail route, key→id resolution with one memoized GET per run, 20-page cap, drop-don't-guess, `/api/dvr/rules/` non-existence, series-rules-in-dvr_settings, comskip response shape) confirmed with no overclaiming; zero `/swagger.json` references remain repo-wide. **Block** on the nvhg7 runbook: the FAILED-row example cited a denylisted key, which lands in `cat.skipped` (EXCLUDED_BY_OPERATOR) — a different counter from `cat.failed` — so it could have led an operator to dismiss a genuine failure, and its cross-reference pointed at a doc with no denylist content. Plus a nav-label Warn.

Round 2 (d4863be3) delta-verified: counter semantics re-derived independently from current code; the example now cites the real DEPENDENCY_UNRESOLVED case and points at the actual denylist location; nav path matches the sibling doc and UI. The optional rewrite to the direct per-row restore action was verified performable AND accurate — the saved path schedules the same DbasRestoreTask through the same `validate_artifact_manifest` chain and opens dry-run-first. CI green both heads.

PR #773, opened 2026-08-04
[enhancedchannelmanager-ax0kf] ADR-014, Dispatcharr contract sweep, and untested-version advisory

MotWakorb:
## Summary

The systemic answer to the guessed-endpoint bug class that produced three shipped defects (q6xjl root causes A and B, lsa0s). PO-approved strategy: keep deep recorded fixtures **and** add a contract sweep, with the version advisory riding this implementation.

**ADR-014** (`docs/adr/ADR-014-dispatcharr-api-drift-strategy.md`, Accepted) — the option table (recorded-fixture-per-path / contract sweep / runtime version gate / schema-driven generation) with verdicts, the sweep's non-negotiable properties, its honest limitation (existence + method + path-param type only — **not** response shape or semantics; that's what the deep fixtures are for), and the event-triggered rule: a Dispatcharr major-version bump triggers a restore exercise before trust.

**The sweep** — an AST-derived extractor walks `DispatcharrClient` and checks every `(method, URL template)` against a recorded 224-path manifest of the live 0.28.2 OpenAPI document. Fails loudly on any unresolvable path expression (never skips silently), carries a canary so a broken extractor can't pass by finding nothing, and its failure message names the offending tuples plus both possible causes (stale manifest vs real client bug). Ships with `scripts/record_dispatcharr_openapi_manifest.py` and 10 negative tests proving each check can actually fail.

**Version advisory** — `GET /api/core/version/` on the existing connection test returns a non-blocking `warning` when the connected Dispatcharr isn't in `TESTED_DISPATCHARR_SERIES` (MAJOR.MINOR series, since Dispatcharr is 0.x). Warn only, never blocks; an undeterminable version stays silent. Surfaced in the settings dialog without touching verified state.

## Sweep result on first run: GREEN

97 call sites, 92 distinct (method, template) pairs, **58 distinct URL templates — zero mismatches**, matching the design-time baseline exactly. Three subtleties handled honestly rather than papered over:
- `GET /api/schema/` is absent from the schema's own paths (drf-spectacular doesn't document its own route) — the single exemption, and it **self-invalidates**: a test asserts the exemption set is exactly that entry *and* that the path is still absent, so it fails the day upstream documents it.
- Path-param type checking is asymmetric by necessity: a naive int-vs-non-integer rule produced 12 false positives on untyped Django converters. The shipped rule fails on a declared `format` (uuid — the shape an int→UUID pk migration produces; the document already carries 5) and passes bare `string`.
- One helper reaches httpx directly; its call sites are swept, and a registry test fails on any *new* raw httpx call site until a human classifies it.

Version 0.18.1-0022 + CHANGELOG.

## Verification

TDD on the advisory (helper/client/router observed red first). Full backend suite 8214 passed, 13 skipped — engineer run + independent orchestrator re-run, both exit 0; sweep test file 18 passed standalone; fake-test guard, version scripts, `mkdocs build --strict`, `npm run docs:check`, frontend `tsc --noEmit` + ESLint all clean. Advisory verified live read-only against the real Dispatcharr from the test container (no warning under the shipped constant; warning + server log when the constant was temporarily pointed elsewhere), test container restored afterward.


MotWakorb:
**Review record (AI code-reviewer, three rounds, relayed): Clear-to-merge.**

R1 (4314dd6b): **Block** — 22 advisory tests written, run locally, never committed; the constant they guard turns `in` into a substring match on a missing trailing comma. Plus W1 (`get_version()` had zero callers; the probe hand-wrote a URL the sweep can't see, while ADR+commit claimed it was swept), W2 (four ordinary refactors left the sweep green while dropping calls), W4 (stale counts), N2 (unbounded upstream string to UI+log).

R2 (3e7b4943): fixes verified — but the W1 remedy **introduced a regression**: routing through the client inherited the 401-retry branch, so a 401 on the probe triggered a second credentialed login against a 3/min rate limit, while three artifacts asserted that was impossible. New Block. Also: 'coverage cannot silently shrink' was false for a *cross-module* base class (31 of 93 pairs lost, canary floors untripped) — the same defect class as W1, twice on one artifact.

R3 (19bcd79b): all seven items closed, plus a written invariant pass over every ADR claim that found **three more** false ones (incl. the ADR *underselling* the sweep — the dead schema path would in fact have been caught).

Final ceilinged verification, by construction on the real module and real endpoint: wire trace shows exactly 1 login POST under a 401 (2 without the fix — the opt-out is load-bearing); a physical cross-module mixin turns the sweep RED with actionable guidance while the in-module case stays lossless; 6/6 prior seam evasions plus 13 reviewer-invented shapes all correctly classified; false-positive guard holds; invariant-pass claims independently measured (12 bare-string params, 5 uuid params, path_count 224 == len(paths)). Counts measured at 98 calls / 93 pairs / 59 templates, 0 unresolved, 5 reviewed seams, nothing allowlisted. CI 22 success / 0 failing, CLEAN.

Two Warns + two Nits filed rather than fixed, per the declared last-round ceiling.

PR #781, opened 2026-08-06
[enhancedchannelmanager-ixdaw][enhancedchannelmanager-oebpv] Resolve case-differing stream names, and audit playability for every restored channel

MotWakorb:
## The result, first

Drill run 4 answered *"did a full backup, total destruction and clean restore reproduce the instance?"* with **no** — and one channel was the entire reason. With this change, on a live disposable stack:

```
# Inventory diff — variant: encrypted
**Did the restore reproduce the instance? YES**
## Findings (0)
- (none)
```

The first zero-finding round trip in this drill's history.

| | ARCHIVED | BEFORE this PR | AFTER |
|-|-|-|-|
| slot 1 | `TX \| DALLAS \| PBS KERA` (102) | `TX \| Dallas \| PBS KERA` (101) — wrong member | **102** ✅ |
| slot 2 | `TX \| Austin \| PBS KLRU` (98) | 98 ✅ | **98** ✅ |
| slot 3 | `TX \| Dallas \| PBS KERA` (101) | placeholder, **no url** | **101** ✅ |

Playback **fetched, not inferred**: the channel returns 200 / `video/mp2t` / 262144 bytes on **both** artifact variants. On the standard redacted path it previously returned **HTTP 500**.

### `ixdaw` — the matcher collapsed a case-differing pair onto one stream

`_normalized_name()` case-folds, so Tiers 2/3 saw one match where the destination had two, and the lowest-id tie-break handed **both** archived slots id 101. Id 102 — a byte-identical name match — sat unused.

- Encrypted path: the channel silently lost a stream.
- Redacted path: the duplicate made Dispatcharr reject the whole channel PATCH — `psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "unique_channel_stream"` — and the channel kept all three placeholders and would not play.

**Fix:** within each exact-name tier, among the candidates that tier's predicate already admits, prefer any whose **raw `name` is byte-identical**; the lowest-id tie-break then runs inside whichever set was selected.

**Cannot regress.** With no raw-name match the selected set is unchanged and the result is bit-identical to before. Tier integers unchanged (a raw hit in Tier 2 is still Tier 2). Still pure, still order-independent. `claimed_ids` in `placeholder_rebind.py` is **byte-identical** — it is now the backstop for a *genuine* identical-name collision rather than the primary handling.

### `oebpv` — the playability audit was blind on a repeat restore

The verdict was nested inside "did *this run's* placeholder survive?", so a channel stranded by an **earlier** restore escaped it. Run 4 measured a repeat restore reporting `channels_with_no_playable_stream: 0` and empty `notes[]` for a channel answering HTTP 500.

**Fix:** the verdict is taken for **every** restored channel, from `final_ids` (what it is actually left holding) against the real URL-bearing destination streams. Bad slots are named from the destination's own stream list rather than surfacing as `<unknown>`. The early return on an empty placeholder set is gone — it was exactly what made those channels invisible.

**Rebinding and deletion stay strictly ledger-scoped**: nothing this run did not create is ever rebound or deleted. Only the *audit* widened.

**No over-trigger:** a channel holding a leftover placeholder beside a real stream still reports `has_playable_stream: true` and still yields `outcome: success` — which `daziw` requires and run 4 verified as already correct.

## Verification

Live, on a disposable drill stack (never production; both modules patched in and both markers confirmed present at every rollout):

- encrypted round trip → **zero-finding diff**; rebind log `14 of 14 slots rebound` (was 13), `0 channels still hold a slot that cannot play`
- standard redacted round trip → **zero channel findings**, all 12 channels on real streams, **4/4 playback**
- second restore of the same artifact → **genuine no-op**; before this PR it silently dropped a slot and reordered the channel
- `daziw` unregressed → a redacted restore still reports `completed_with_failures` with a `warning` / "Task Completed with Warnings" notification
- regression checks all still pass: uploaded-logo sha256 identical, 12/12 EPG links, credential fingerprint `0ea118b1` unchanged, enabled groups 1/375, custom stream profile resolves to `DrillAgent4` **by name** on a target where id 4 is a decoy

Gates, re-run independently rather than taken from the authoring agent:

- `8472 passed, 18 skipped`, **0 failures**
- **Both fixes red-proofed by me**: reverting each in isolation turns the new tests red (4 matcher, 3 rebind), and the restored files were verified byte-identical to the fixed versions
- `py_compile` clean; `check_version_consistency.py` passes

## Honest limits

- The `oebpv` **detection** path is proven by unit test and by the audit now running unconditionally in the live log. My live stranding experiment ended up exercising the **repair** path instead — the restore rebound the channel and it played, so the reported `0` was truthful. I could not construct a channel that stayed stranded live. I am not claiming end-to-end live proof of detection.
- Not fixed here and still open: `asf3n` (a degraded restore emits both a `warning` and an `error` notification), `dgnms` (preview health counters under-predict on a fresh target), `w2jky`, `sdpzy`.
- Pre-existing and unchanged by this PR: the redacted path leaves orphaned placeholder stream objects (14 in both run 4 and run 5), bound to no channel; and the synthetic `ECM Custom Streams (DBAS restore)` account survives even when empty.

Full report and evidence: `~/ecm/backup-restore-runs/2026-08-05-run5-fixverify/` (out of repo by PO decision).


MotWakorb:
**Correction to this PR's description — `w2jky` is RETRACTED, it was my error.**

The description lists `enhancedchannelmanager-w2jky` ("saved-backup modal claims an unencrypted artifact is encrypted") among the still-open defects. That finding was wrong and the bead is now closed as my own error.

`This backup is encrypted` is a **checkbox label**, not an assertion the UI makes. In `frontend/src/components/DbasRestoreSavedModal.tsx:41` it is `useState(false)` — it renders unchecked, offering the operator a way to say "mine is encrypted" and reveal the passphrase field. My own drill capture proves it was unchecked: the passphrase field is gated behind `{isEncrypted && (...)}` and never appeared in the innerText I recorded.

Root cause of the mistake: I read control labels out of `document.innerText` and treated them as UI assertions without reading control **state** back out of the DOM — the exact trap the drill procedure warns about.

Nothing in this PR depends on that finding; the diff is unaffected. The documentation paragraph added on the back of it (merged in #780) is being reverted separately.

The other three remain genuinely open and are being fixed: `asf3n`, `dgnms`, `sdpzy`.

PR #815, opened 2026-08-10
fix: exempt documentation-only changes from the version-advance ship guard

MotWakorb:
Bead: `enhancedchannelmanager-uf2gh`. Sequenced deliberately behind PR #814, which merged first.

## The defect

`.claude/hooks/version-advance-guard.sh` blocks `gh pr create` when the build number does not advance. It performed no path classification, so it fired on documentation-only changes that CI deliberately exempts (`test.yml` gates the advance step on `docs_only != 'true'`, commented *"Documentation-only PRs are exempt: they carry no build to advance"*).

**Not an oversight — a half-finished bead.** `w9irb`, which created this hook, already specified *"SCOPE using the EXISTING docs-only classification."* Only the CI half implemented it. The hook's own header calls it *"the agent-side half of the build-advance gate"* and states *"a false block here is a friction bug, not a safety gap"*, so exceeding CI was never the intent.

Measured cost: a two-line `.beads/issues.jsonl` PII redaction was forced to bump three version touchpoints across `frontend/package.json`, `backend/main.py` and `backend/routers/backup.py` — files it had no business touching — and consumed a build number another in-flight branch already held. More broadly, 15 of the last 40 merged PRs are "documentation plus a version bump and nothing else".

## The fix

The hook shells out to `scripts/classify_changed_paths.py` in its already-resolved `PROJECT_DIR`, so there is **one** definition of docs-only rather than a bash reimplementation. Classification uses `git diff --name-only origin/<base>...HEAD` — three-dot, matching the file set the PR will actually contain rather than the working tree.

`docs_only` is extracted **by key**, not by whole-output equality. That matters: an earlier revision compared the classifier's entire stdout against two literals, which #814's second output line (`docs_site_affected`) would have broken — every invocation would have taken the fail-open path, producing a permanently degraded guard that warns on each run while enforcing nothing. Caught and fixed before shipping.

Fails **open** on classifier failure, matching the hook's existing missing-checker behaviour and `check_version_advances.py`'s own soft-pass on an unavailable baseline. A missing or unrecognised *value* for `docs_only` is still "could not classify" and is never read as `false`; only the presence of other keys stopped being fatal. Duplicate `docs_only` lines fail open rather than letting the first win, and `\r` is stripped so a CRLF checkout cannot degrade the guard the same way. Every fail-open path warns loudly naming the reason.

## Verification

Live exercises against the **real** classifier on current `dev`, unmodified:

| Exercise | Result |
|---|---|
| docs + beads only, no bump | `SKIPPED the build-advance check` → exit 0 |
| **code change, no bump** | `FAIL: build 60 equals baseline 60` → exit 2 |
| this branch (code change *with* bump) | `OK: build advances` → exit 0 |

The second row is the one that matters: under the pre-fix parsing that case passed with a warning phrased as though the guard were working.

Also verified that **every** stdout path in the real `main()` emits both keys, including the unreadable-input path, so the by-key extraction has no blind spot against the actual script.

29 hook cases in a new 424-line fixture self-test, closing a gap bead `yxtuz` recorded as a backlog candidate. Five `TestVerdictIsReadByKey` cases confirmed red against the pre-fix revision. Reverting the hook entirely fails 12 of 29, so the deny-side and injection cases are not vacuous. `shellcheck -S warning` clean.

## Known, and flagged rather than discovered

The docs-only definition is still `**.md` OR `.beads/**`. Bead `gnz19` records that `**.md` is too broad because `.md` files are live test fixtures — `backend/tests/unit/test_check_em_dashes.py:42` pins `docs/security/threat_model_dbas_import.md`, and the classifier returns `docs_only=true` for it. So a PR editing that fixture now skips the version check agent-side. Inherited from the shared classifier, not introduced here, and `gnz19` is queued — but this change is what first makes it reachable.

**Fail-open residual:** while the classifier is unusable the agent-side check is gone entirely. It warns loudly, but a warning is not a block. CI still enforces the same rule server-side against the PR head.

**Contract gap:** nothing enforces that the classifier keeps emitting `docs_only`. The hook's tests pin the hook's side; the classifier's obligation is unpinned. Filed as `m1ug0`.


MotWakorb:
## Rework after review — both blockers closed, plus a defect the fix itself exposed

Head is now `f6109f76`. Version re-bumped `0061` → **`0.18.1-0062`**, because bead `9ogyd`'s CI half (PR #816) takes 0061 and merges first.

### B1 — chose real visibility, not the minimal closure

The hooks contract confirmed independently: on exit 0, stderr reaches the debug log only and Claude never sees it. So the "fail open, but loudly" bargain had no loud half. One extra clause makes the fix safe — *"Exit 2 means a blocking error. Claude Code ignores stdout and any JSON in it"* — so adding stdout JSON to the exit-0 paths cannot disturb the block path.

An `emit_notice` helper now writes to stderr **and** emits `hookSpecificOutput.additionalContext` (reaches Claude) plus `systemMessage` (reaches the operator). Called only on paths that then exit 0; the block path is explicitly stderr-only and emits nothing on stdout, pinned by a test. The guard announces, it does not vote — no `permissionDecision` is set.

The self-test now asserts on parsed JSON rather than process stderr, which was the gap: the old assertion proved the script wrote bytes to fd 2, not that anyone received them. Mutant making `emit_notice` stderr-only fails 16 tests.

### B2 — the exemption is now reachable

`docs/shipping.md` §3 is split. §3a comes first and tells the agent to classify **before editing any version literal**, with the command to run. §3b is the existing lockstep content, unchanged.

A callout names the circularity that caused this: bump first, and the diff now contains the three version touchpoints, so the verdict flips to `docs_only=false` and the bump justifies itself. That is how 15 of 40 consecutive merged PRs carried a version bump for a change that was nothing but Markdown.

`docs/versioning.md` gets the matching note: **lockstep governs how you bump, not whether you bump.**

### An unplanned finding: the guard fired on its own commit message

Making exit-0 output visible immediately exposed a pre-existing defect. Stage 1 treated every `;` as a command separator, so prose inside a heredoc — *"…ran at step 3; gh pr create fires at step 6"* — matched, and the new base parser then read `--base main` out of the surrounding text. It fired twice, each time emitting a confident release-cut notice about a command that was not a ship.

Silent before, misinformation after. Detection is now two-stage: stage 1 is the same cheap grep and remains the entire cost for every non-ship command; stage 2 runs only on what stage 1 admits, drops heredoc bodies, tokenises the rest, and requires `gh` `pr` `create` to survive as three consecutive tokens at a command position. Stage 2 only ever narrows stage 1.

### Also included: the hook half of bead `9ogyd`

`--no-renames` on the classifier's `git diff`. PR #816 fixes the CI half; this closes the same hole agent-side, and the two halves are one defect.

The interaction is worth stating: once §3a stops agents bumping on documentation PRs, the classifier's verdict becomes load-bearing for the first time rather than academic. That is exactly when a rename blind spot starts to matter.

Live proof, real repo and real hook — `git mv` a test suite into `docs/`:

| | Result |
|---|---|
| without `--no-renames` | exit 0, **SKIPPED** — the guard stands down on a branch that deleted a test suite |
| with `--no-renames` | exit **2**, BLOCKED |

### Warns addressed

W1 and W2 were unpinned load-bearing behaviours — the `^` anchor and the three-dot diff. Both now have a test that fails on exactly that mutation and nothing else. W5 went further than asked: the base is read by `shlex`-tokenising the invocation rather than by regex, so `--base mainline-experiment` is no longer exempt and a `--base dev` ship whose body quotes `--base main` is no longer exempt. W6 and W8 corrected.

### Corrected figures for the PR body

The original body's numbers were stale, not merely wrong. Self-test is **849 lines, 45 cases**; reverting the hook alone fails **29 of 45**.

### Verification

45 hook cases, 72 classifier, 30 advance-checker, all passing. Seven live exercises against the real hook including a real docs-only branch, a real code branch, a release cut, and the rename hazard both ways.

### Backlog, not actioned

`-B main` (gh's short form) is still not read, so a release cut written that way false-blocks — pre-existing, and narrowing was the ask rather than widening. `docs/shipping.md` §6 says "7 passing status checks" while `CLAUDE.md` says "the 5 required checks"; §6 is what a ship agent reads, so that is worth reconciling.

MotWakorb:
## Round 3 — both blockers closed

Head . Merged  (PR #816) rather than rebasing, per shipping.md's own instruction for a published branch; only the three version touchpoints conflicted.

**B-1.** Stage 2 no longer re-implements the decision. It now deletes heredoc bodies and quoted spans, then runs **stage 1's own regex** on the remainder — no tokenising, no second opinion about what a ship looks like. All four shapes restored:

| shape | pre-branch | round 2 | now |
|---|---|---|---|
| `git push …` ⏎ ship `--base dev` | BLOCK | **ALLOW, SILENT** | BLOCK |
| `POST_PR_URL=$(ship --base dev …)` | BLOCK | **ALLOW, SILENT** | BLOCK |
| `(ship --base dev)` | BLOCK | **ALLOW, SILENT** | BLOCK |
| `git push;ship --base dev` | BLOCK | **ALLOW, SILENT** | BLOCK |

The header now records the lesson rather than the invariant that hid the bug: *"stage 2 only ever NARROWS stage 1" was true, and useless. For a guard, over-narrowing IS the failure mode.* The rule is now that stage 2 may delete data from the text and may not re-implement the decision.

**B-2.** 3a decides from what the agent edited, as a table with no command, and states outright that the arbiter cannot run yet — *"an empty changed-file set is classified as code. That answer looks like a verdict and is not one."* New **6a** runs after the commit, where the classifier has real input, uses `--no-renames`, and routes `docs_only=false` back to 3b via `git commit --amend`. Step 6's staging now separates the three touchpoints as CODE CHANGE ONLY.

**Also closed:** N-1 through N-4, and both security findings — base scanning stops at the separator ending the invocation and takes the **last** `--base` (matching gh), and value-taking flags have their values skipped, so `--title "--base=main"` is a title.

10 new fixtures, all red against `f6109f76`. Six targeted mutants, each killed by exactly its own case. Self-test now 1041 lines / 59 cases; reverting the hook alone fails 34 of 59.

Gates: 59 + 81 + 30 passed, version consistency and advance both green at `0.18.1-0062`.

MotWakorb:
## Correction to the previous comment

The previous comment was posted with shell-mangled formatting. Corrected record below.

## Round 3 — both blockers closed

Head `1db62370`. Merged `origin/dev` (PR #816) rather than rebasing, per shipping.md's own instruction for a published branch; only the three version touchpoints conflicted.

**B-1.** Stage 2 no longer re-implements the decision. It deletes heredoc bodies and quoted spans, then runs **stage 1's own regex** on the remainder — no tokenising, no second opinion about what a ship looks like. `command_position` is gone entirely. All four shapes restored:

| shape | pre-branch | round 2 | now |
|---|---|---|---|
| `git push …` then ship `--base dev` on the next line | BLOCK | **ALLOW, SILENT** | BLOCK |
| `POST_PR_URL=$(gh pr create --base dev …)` | BLOCK | **ALLOW, SILENT** | BLOCK |
| `(gh pr create --base dev)` | BLOCK | **ALLOW, SILENT** | BLOCK |
| `git push;gh pr create --base dev` | BLOCK | **ALLOW, SILENT** | BLOCK |

The header now records the lesson rather than the invariant that hid the bug: *"stage 2 only ever NARROWS stage 1" was true, and useless. For a guard, over-narrowing IS the failure mode.* The rule is now that stage 2 may delete data from the text and may not re-implement the decision.

**B-2.** Step 3a decides from what the agent edited, as a table with no command, and states outright that the arbiter cannot run yet — *"an empty changed-file set is classified as code. That answer looks like a verdict and is not one."* New step **6a** runs after the commit, where the classifier finally has real input, uses `--no-renames`, and routes a `docs_only=false` result back to 3b via `git commit --amend`. Step 6 staging now separates the three version touchpoints as CODE CHANGE ONLY.

**Also closed:** N-1 through N-4, plus both security findings. Base scanning now stops at the separator ending the invocation's own arguments and takes the **last** `--base`, matching gh's semantics, so `--base main --base dev` resolves to `dev` and blocks. Value-taking flags have their values skipped, so `--title "--base=main"` is a title and blocks.

10 new fixtures, all red against `f6109f76`. Six targeted mutants, each killed by exactly its own case. Self-test now 1041 lines / 59 cases; reverting the hook alone fails 34 of 59.

Gates: 59 + 81 + 30 passed; version consistency and advance both green at `0.18.1-0062`.

Still open: `-B main` (gh's short form) is deliberately not read as a base — recorded backlog, not a regression.

MotWakorb:
## Round 4 code review — BLOCK

Reviewed exact head `1db623703fefc25aa3604ab8965f46b2e5657b75`. One blocking finding remains.

### B-1 — stage 2 deletes contexts where Bash still executes command substitutions

The new invariant says quoted spans and heredoc bodies are data that can be deleted before re-running the stage-1 ship regex. That is not true for two Bash contexts:

```bash
PR_URL="$(gh pr create --base dev --title t --body b)"
```

and:

```bash
cat <<EOF
$(gh pr create --base dev --title t --body b)
EOF
```

Bash executes command substitutions inside double quotes and inside an unquoted heredoc body. The hook deletes both regions, decides `ship=false`, and exits 0 with empty stdout/stderr on an unbumped code change. A plain `gh pr create --base dev` control correctly exits 2. Single-quoted equivalents and quoted/backslash-quoted heredocs correctly remain inert.

This directly falsifies the header claim that deleting those regions cannot hide a real ship. Existing coverage pins only the unquoted assignment form and either quoted heredocs or commands outside heredoc bodies.

Because this is the second harmful over-narrowing regression in the same parser, implementation is paused until the next patch records an explicit execution-context invariant/failure-mode model distinguishing inert data from expansion-capable contexts. Required red-first fixtures: both executable forms above plus their inert single-quoted / quoted-delimiter controls, followed by targeted mutation proof.

Other reviewed surfaces had no additional blockers: `--no-renames`, keyed classifier parsing, base scoping/repetition, exit-0 JSON vs exit-2 stderr, docs/version consistency, and all eight changed files.

Security assurance separately passed the previously documented matrix, but did not include these newly identified expansion-capable contexts; this code-review finding therefore controls the merge decision.

MotWakorb:
## Final redesign and independent review — PASS

Head `3a248c2ac24ff15057c26d4d1516901ff5108196`.

After the repeated over-narrowing regressions, implementation stopped for a focused mechanism spike. The conclusion was that PreToolUse receives raw Bash source, not an execution trace, so a locally maintained quote/heredoc/substitution parser cannot reliably distinguish execution from inert data.

The final design removes that parser entirely:

- broad Bash hook routing remains;
- one monotonic raw detector admits any whitespace-separated literal `gh pr create` followed by a non-word boundary;
- admitted candidates are never narrowed by quote, heredoc, token, substitution, command-position, or base parsing;
- literal prose/data may therefore run the repository-state check, an explicit visible-friction tradeoff rather than a silent allow;
- diagnostics say **possible PR creation command**, not that execution was proven;
- hook-input `cwd` plus `git rev-parse` resolves the repository, and the docs explicitly require PR creation from the target checkout rather than an inline `cd` the pre-execution hook cannot forecast;
- CI remains the authoritative merge gate.

The final boundary also covers valid zero-argument invocations followed immediately by `)`, command-substitution close, `;`, `&&`, pipe, or redirection, while rejecting adjacent identifiers such as `createfoo`, `create_foo`, and `create2`.

Verification: 35 hook-state cases, 146 targeted hook/classifier/version-check tests, full backend suite at 90.95% coverage on the final design, syntax/JSON/diff/version/build gates, plus mutation proofs for deleting or over-anchoring the detector.

Independent code review: **PASS** on the complete PR at this head. Independent security re-review: **APPROVE**, with no blocking security finding in the stated conservative raw-literal contract.

PR #855, opened 2026-08-13
fix(auth): close the remaining admin-gate inconsistencies across five routers

MotWakorb:
Closes `9kwzp.10` and `9kwzp.12`. One branch because both edit the gate inventory and the version touchpoints.

**⚠️ This PR changes the behaviour of seven shipped MCP tools. See the decision at the bottom — I am holding the merge for it.**

## The finding that matters most: a gate went stale when its capability grew

`POST /api/backup/restore-dbas` and `/restore-dbas-saved` sat on the plain admin tier while `/restore` and `/restore-saved` were human-admin. That reads as an oversight. It wasn't.

Bead `6n76m` **deliberately** left the DBAS restore on the plain tier, and its reasoning was correct at the time: it was "the guardrailed dry-run-by-default DBAS restore, which applies denylist-filtered settings to the Dispatcharr upstream rather than ECM's own settings file."

Then bead `dfkbn` added `backend/dbas/importers/ecm_settings.py`, and the restore began writing **ECM's own settings blob** — excluding only the live Dispatcharr connection, `INTERNAL_KEYS` and redaction sentinels. So it now writes `emby_base_url` / `plex_base_url` / `jellyfin_base_url`, the notification credentials, the #473 safety caps, and `ssrf_outbound_mode`.

The gate was right when written and became wrong when the capability underneath it grew. **A route inventory structurally cannot catch that** — the route and its dependency never changed. Recorded as its own class in the inventory.

## Verdicts

| Item | Verdict |
|---|---|
| `PATCH /settings/security` | **Denied.** New `RequireHumanAdminForOutboundPolicy`. Gating eleven outbound sinks while their policy stayed writable by the same principal was a partial control. |
| DBAS restore pair | **Denied.** Reused `RequireHumanAdminIfEnabled`; stale-gate reasoning above. |
| Sync targets | **Writes denied, reads admitted.** |
| Cloud targets | **Writes denied, reads admitted.** Shared `RequireHumanAdminForOutboundDestination`. |
| Alert methods | **Denied including reads.** New `RequireHumanAdminForNotificationCredential`. |
| `reset-stats` | **Denied.** New `RequireHumanAdminForStatisticsReset`. |

**Sync/cloud targets went against the bead's own framing.** The bead suggested these might be working-as-designed since `jcj0f` deliberately exposed them as MCP tools. Adversarial review pushed back and I agree: a shipped tool establishes product *intent*, not least privilege. A write names a host, stores the credentials a scheduled job authenticates with, sets `insecure`, and registers a `dbas_sync_<id>` task; `update` repoints a push already configured. The router's own docstring says the authoritative SSRF check runs at execute time. Encryption at rest and masked responses bound *disclosure*, not redirection.

**Alert methods are denied on reads too**, because `GET /api/alert-methods` and `/{id}` return `config` verbatim — Discord webhook URL, Telegram bot token, SMTP password. `GET /types` stays plain admin as a static catalogue.

**`reset-stats` does not inherit `restart-services`' reasoning.** That one kept MCP access because rebuilding services from saved settings is work a settings write triggers anyway. Nothing re-derives those seven tables: no compensating write, no rollback ledger. Contrasted in-code with the pipeline rollback routes, which *are* admitted because they are the pipeline's own reversal mechanism.

## First-run

No lockout. One genuine first-run caller exists — `SecurityFirstRunModal.tsx` PATCHes `/api/settings/security` — and is safe because the factory returns before both the admin check and the MCP rejection. That route already carried an admin gate; only its MCP half changed. Every route pinned in both `setup_complete=False` and `require_auth=False` postures.

Every gate no-ops when auth is disabled or setup is incomplete. The CHANGELOG and docstrings state that condition rather than claiming routes categorically cannot be reached.

## Tests

**114 new** across 19 routes × 4 postures, witness-asserted so a refusal is proved to be the gate answering. Plus `test_every_human_admin_gate_names_its_own_surface` — the eight denial bodies must be pairwise distinct, so a copy-pasted body fails rather than leaving every behavioural test green while misdirecting triage.

Red-without-fix proven by reverting sync-targets to `RequireAdminIfEnabled`: 8 failures across both files.

`9632 passed, 9 skipped` (baseline 9517, +115). Frontend `2792`. All guards green at `0.18.1-0095`.

---

## DECISION NEEDED before merge

**Seven shipped MCP tools now return 403 on auth-enabled installs:** `list_alert_methods`, `create/update/delete_cloud_target`, `create/update/delete_sync_target`.

Their docstrings now say so and point at the UI, so it degrades honestly rather than mysteriously. But it is a real capability loss for anyone automating through the sidecar, and `jcj0f` added several of them on purpose. Each is a one-line dependency swap plus a set move if you want it kept.

Two smaller items, neither actioned:
- The alert-method `config` leak is **gated but not redacted** — a human admin still receives the SMTP password in clear. Filed as `9kwzp.13`.
- `GET /api/cloud-targets/{id}` is documented in `docs/api.md` but does not exist in the router. Pre-existing.


MotWakorb:
## Update: adversarial review found a regression; fixed in `8b40bfe0`

An independent adversarial review with full repo access (read-only) flagged this as not merge-ready despite all 29 checks being green. Two blocking items, both fixed. A **second regression was found while fixing the first.**

### Blocking 1 — non-admin Settings was broken, and auth-disabled installs were worse

`SettingsTab.tsx:1183` called `listAlertMethods()` with no admin guard, hitting the route this PR made human-admin-only. The original fix touched `AlertMethodsSection` and missed `SettingsTab` entirely.

Verifying that surfaced a second, subtler break introduced by this PR. `useAuth` returns early **without fetching the user** when the backend reports `require_auth=false` or `setup_complete=false` (`hooks/useAuth.tsx:91`), leaving `user` null. A `user?.is_admin ?? false` predicate reads null as "not admin" — so on an **auth-disabled single-operator install**, where every gate here no-ops server-side, the operator would have been locked out of their own Alert Methods section and Reset Statistics button.

`BackupRestoreSection` already had the correct predicate. Now one derived `isAdminUser = !user || user.is_admin` (`:413`), used at all five sites. Six new cases in `SettingsTab.alertMethodGate.test.tsx`, red-without-fix proven twice: once with the guard removed, once with `?? false` restored.

### Blocking 2 — the DBAS gate denied read-only dry runs

Both endpoints gated before anything read `confirm_apply`, so the zero-mutation preview was denied along with apply — contradicting the re-tiering rationale, which was about settings **writes**.

Resolved as a split, because the two routes are **not symmetric**:

- **`/restore-dbas`** takes a caller-supplied **upload**, streamed to the config partition and decoded before `confirm_apply` is ever read. Even a preview hands the principal a parser and a disk-consumption surface with an artifact of its choosing. No MCP tool exists for it (no file upload over MCP). Blanket `RequireHumanAdminIfEnabled`, dry run included.
- **`/restore-dbas-saved`** names an artifact **already on disk**, which only an admin could have saved. Keeps `RequireAdminIfEnabled` and refuses only the apply, in the handler, following the `kgz3k` precedent — a dependency cannot see a body field without consuming the body.

The zero-mutation claim was verified rather than taken from the docstring: `restore_orchestrator.py:666` forces `is_dry_run = True` and is documented as the single choke point a caller cannot opt out of.

Worth noting: the carve-out predicate was first written in `backup.py`, and a test caught that it read a **different `get_auth_settings` binding** than the gate stacked above it. That is real drift risk, not a test artifact, so it moved to `auth/dependencies.py` as `ResolveIsMcpServicePrincipalIfEnabled`, guaranteeing the carve-out's setup-mode early return is literally the same one the gate makes.

### Caller sweep re-done for every gated route

Alert methods (`AlertMethodsSection` + **`SettingsTab` ×3**), `reset-stats`, cloud-target CRUD, sync-target CRUD, `/restore-dbas*`, `PATCH /settings/security`. Nothing else found.

### Prose corrected

- **"Masked means harmless" is gone.** The inventory and CHANGELOG now state what admitted reads still disclose: `base_url`/`upload_path`, the `insecure` flag, credential key names and last-4, `credential_version`, revocation state, past sync outcomes. Topology plus credential fingerprints, handed to the principal declared untrusted to manage those destinations. What masking buys is stated narrowly: a read alone cannot authenticate to the destination.
- **Coarseness is owned explicitly.** Rename-only updates, severity-filter changes and *tightening* `ssrf_outbound_mode` are all denied like their dangerous siblings. Coarse is defended on its merits — these routes don't mix admin config with per-user preferences the way `POST /api/settings` does, so it costs the credential nothing legitimate, fails safe, and is far cheaper to verify.
- **"No first-run lockout" is no longer categorical.**

### Verification

`9637 passed, 9 skipped` backend (was 9632), `208 files / 2798 tests` frontend (was 2792). Both beat baseline. All guards green at `0.18.1-0095`.

Also confirmed by the review, with commit-level evidence now in the bead and CHANGELOG: the stale-gate narrative holds. `/restore-dbas` landed `21f93e683` (2026-06-19) with the plain gate; `6n76m` (`e83d31b1`, 2026-07-08) human-gated the legacy paths while correctly leaving DBAS alone; `fd63235d` (2026-08-04) added `ecm_settings.py` with a generic `setattr` loop plus `save_settings`, which is what made the distinction stale.

### Still filed, not fixed here

- `SecurityFirstRunModal.tsx:64` writes its localStorage "seen" flag **before** awaiting the PATCH and swallows failures, so a 403 permanently retires the prompt with nothing persisted. Pre-existing.
- `docs/api.md` documents `GET /api/cloud-targets/{id}`, which the router does not implement. Pre-existing.

**Merge still held** pending a maintainer decision on the MCP tool capability changes.

PR #890, opened 2026-08-18
[enhancedchannelmanager-04c0u.11] Give the governance audit a credential it can read with

MotWakorb:
Completes bead `enhancedchannelmanager-04c0u.11` (child of epic `enhancedchannelmanager-04c0u`).

## Why

The Security Governance Audit activated by #888 failed on its first real run (Tests `32167876126`, job `95811655391`) in about two seconds:

```
gh: Resource not accessible by integration (HTTP 403)
```

Four of the six controls it reads are repository **Administration** surfaces, and the built-in `GITHUB_TOKEN` can never reach them — the workflow `permissions:` key has no `administration` scope at all, so no setting would have fixed it. GitHub separately documents that `security-events` covers code scanning only and that secret-scanning alerts require a GitHub App or a PAT.

## What this does

The elevated reads now run on `GOVERNANCE_AUDIT_TOKEN`, a repository secret holding a read-only fine-grained PAT scoped to this repository alone (Administration, Contents, Secret scanning alerts, Dependabot alerts — no write permissions).

- The secret is passed with an **explicit single-secret mapping**; `secrets: inherit` is used nowhere in `.github/`, and a test asserts the mapping is exactly `{GOVERNANCE_AUDIT_TOKEN}`, so widening is caught rather than only inheritance.
- The one call the built-in token can serve — reading a branch ref, which needs only `contents: read` — deliberately stays on `github.token`, and a test keeps it there. The behavioural harness proves it by logging which token the stubbed `gh` actually received.
- `security-events: read` was granted and used by nothing, so it is dropped. The audit's permissions are now exactly `contents: read`.
- No fallback between the two credentials in either direction, and no swallowed exit code. **Four** distinguishable outcomes: secret absent or empty (named by name), 401 rejected credential, 403 under-scoped (naming the missing permission), and a genuine control violation.

## Cadence arming

`schedule:` and `workflow_dispatch:` on this file are currently **inert**: GitHub registers those triggers only for workflows present on the default branch, and this file lives on `dev` (`gh api .../actions/workflows/security-governance-audit.yml` answers 404). They arm by themselves at the next release cut. Rather than assume it, the audit probes `main` for its own file on every run and reports `ARMED` or `PENDING` — and the summary states only the check that actually ran ("this file is present on `main`"), explicitly noting that the cron expression itself is not validated. Reading the workflows endpoint would have produced a stronger-sounding claim at the cost of an `actions: read` grant nothing else needs, so it was deliberately not done; a test asserts that grant was not added.

## Review round — what changed after three independent reviews

The first revision's guards were largely vacuous, and the reviews proved it rather than suspected it. Recording that plainly, because the original PR body claimed "16/16 mutants killed" and an independent reviewer then ran 15 mutants of its own and **12 survived**:

- `test_governance_audit_swallows_no_exit_codes` used `\|\|\s*(true|:|echo)\b`. The `\b` after `:` can never match, so `\|\| :` — the most idiomatic spelling of the exact defect class this repo has been burned by — passed, along with `\|\| status=0`, `\|\| /bin/true`, `set +e` and `\|\| printf ''`. Rewritten as an invariant: the only permitted `\|\|` continuation is the exact `\|\| status=$?`, and `set +` is banned outright.
- `test_under_scoped_credential_is_reported_separately_from_a_violation` was satisfied by the file's own **header comment**. Replacing the entire elevated step body with the naive pre-PR version left it passing. It now asserts line-anchored call sites against comment-stripped step bodies, plus a structural invariant that every `gh api` in an elevated step has 401, 403 and 404 branches.
- The empty-secret guard test was satisfied by its own error message: changing `${GOVERNANCE_AUDIT_TOKEN:-}` to `${GOVERNANCE_AUDIT_TOKEN:-x}`, a one-character mutation that makes the guard permanently false while the audit runs unauthenticated, survived the suite. It now asserts the condition itself.
- `set -euo pipefail` was the sole enforcer of four branch-protection assertions and nothing pinned it. Removing that one line let a repository with `enforce_admins: false`, force-pushes enabled and zero required contexts pass. Those assertions now re-raise their own status explicitly, and a test pins strict mode as the first executable line of every step.
- A **paused** Dependabot reported as enabled. GitHub auto-pauses security updates on repositories whose update PRs go unactioned, answering `{"enabled": true, "paused": true}`; the check tested only `.enabled`, so a non-functioning control audited green.
- An expired PAT returns 401, which had no branch — the most likely future failure of this control would have read as a branch-protection violation. The expiry hint was also sitting on the 403 branch, which expiry never reaches.

**Both harnesses are now committed**, because evidence that lives only in a transcript is not evidence: `backend/tests/unit/test_governance_audit_behavior.py` (CI-enforced, 30 scenarios, runs each step's real `run:` body under `bash` against a stubbed `gh`) and `scripts/governance_audit_mutants.py` (developer tool, 34 mutants, refuses a dirty checkout and restores after every mutant). `scripts/README-governance-audit-harness.md` explains both. Current state: **34/34 mutants killed, 30/30 scenarios pass.**

`docs/security/release-security-governance.md` was also corrected — its bootstrap section listed four items as outstanding that are in fact already in place. All six controls were re-verified live against the repository, and the table is now labelled as a snapshot with the audit named as the authority, plus two facts marked explicitly as asserted-not-checked.

## Verification

Build advances `0.18.1-0125` (the base) to `0.18.1-0127`; `0126` is reserved for the in-flight `.8` branch, and the advance check requires only strictly-greater.

**Note:** secret values are unreadable locally, so the credential's actual scoping is proven only by the first post-merge run on `dev`. If it is under-scoped, the audit now names the missing permission instead of failing opaquely. This bead should close on a green post-merge run, not on the merge.


MotWakorb:
Closing: overtaken by the CI gate reduction, not by drift.

This PR exists to give the security governance audit a credential it can read with. Commit `3404d2d5` (2026-08-19, "cut the gate surface to four required checks", by PO decision) **deleted `.github/workflows/security-governance-audit.yml`** along with `version-consistency`, `semgrep-lint`, `trivy-scan`, `dast-scan`, `operator-docs` and their supporting scripts and tests. That landed one day after this PR was opened.

So the job this PR configures no longer exists on `dev`. Rebasing it would mean resurrecting a workflow that was deliberately removed, and its remaining substantive files (`test.yml`, `test_release_gate_policy.py`, `docs/security/release-security-governance.md`) only describe that same removed gate.

Two further notes for anyone reopening this:

- Its build number (`0.18.1-0127`) is not merely stale, it is a **duplicate** — `0127` is already assigned to nine shipped CHANGELOG entries on `dev` from PR #904. Any revival needs a fresh number.
- It was ~64 commits behind `dev` and `CONFLICTING`, including a modify/delete conflict on the deleted workflow itself.

PO decision 2026-08-23. If the governance audit is ever restored, reopen this against the restored workflow rather than rebasing this branch as-is.

Issue #13, opened 2026-01-23
[Feature]: Add channel collections to allow for flexible and easy custom channel order options


burntoc:
### What would you like this feature to do?

I would like to have easy channel re-ordering for Dispatcharr and client guides, and the opportunity to have multiple variations.  Optimally this would support drag and drop.

The desired soltion mimics Channels DVR's concept of Channel Collections. You have have multiple, and each represents an effective sort order for the guide.   I can create one with my normal channel listing just as I like it. On game day I have another collection with fewer channels selected and ordered different so I can more easily find and move between those sports channels (including locals that show sports potentially as well as more traditional all-sports channels).

It doesn't change their groups or anything like that. In a way it is similar to profiles, but it's not just setting which channels are enabled or disabled by profile - it allows you to sort them and view them in the guide differently. More information is here:

https://getchannels.com/docs/channels-dvr-server/how-to/channel-collections/

This would allow for more flexibility, and it reduce reliance on CDVR. Channels DVR is great, but there is some cost as well as a general disregard for iptv sources and hiccups associated with "double buffering" and DVR recording weirdness when Dispatcharr and CDVR are combined.

### How would you suggest implementing?

Channel collections are like Dispatcharr profiles, but in addition to allowing you to have a different subset of channels enabled, they can be arranged in different order. If ECM would allow for another tab with "channel collections" which would allow easy enable/disable and drag and drop sorting of channel order (while leaving group membership of each intact), then provide a link to that new .m3u for each collection, I think we'd be good to go.

This would greatly improve flexibility and ease personalized channel management. This is how my family actually has personal channel guides they can easily load via Channels DVR. Using Dispatcharr would be even better to eliminate the timing issues we see in some cases when both Dispatcharr and CDVR are in play.

### How critical do you think this is?

High - Would significantly improve my experience

Issue #12, opened 2026-01-23
[Feature]: Stats on data pulled from provider


Deekerman:
### What would you like this feature to do?

Currently the stats are great, they show how much data has gone out from Dispatcharr. 

What would be neat is to know how much is coming in from the provider. 

Since it could be 1 stream in to 2 clients it isnt as simple as doubling the out. 

### How would you suggest implementing?

It would be cool if it was on the same page as the stats. I dont know if having them on the same chart would be good or having it on its own.. 

### How critical do you think this is?

Low - Minor improvement, not urgent

MotWakorb:
This was added in an earlier release.

PR #927, opened 2026-08-26
fix(sync): revoke stale Apply and update operator walkthrough

MotWakorb:
## Summary
- reconcile the cross-instance sync guide with the current preview/apply and Scheduled Tasks UI
- prevent a failed preview from leaving stale Apply authorization from an earlier successful preview
- add a live-validated operator flow with state-accurate sanitized screenshots
- narrow the Dispatcharr 0.29.0 acceptance record to claims supported by retained evidence

## Review remediation
Two team reviews requested changes:
- https://github.com/MotWakorb/enhancedchannelmanager/pull/927#issuecomment-5428425841
- https://github.com/MotWakorb/enhancedchannelmanager/pull/927#issuecomment-5428450457

Commit `ff8130a3` addresses the authorized in-scope findings: stale Apply authorization, screenshot policy and state accuracy, explicit source-wins warning, scheduled-task provisioning and troubleshooting, in-flight disable semantics, B-side programme verification, and acceptance-evidence scope.

Commit `90ed9e2b` completes the delta-review shipping requirements: build `0.18.1-0149`, an `[Unreleased]` changelog entry, and log guidance that matches both task-registration failure paths. The separate pre-existing edit-path authorization defect is tracked as `enhancedchannelmanager-bt4r9`; it is not folded into this PR.

The walkthrough remains in the existing Cross-Instance Sync reference under Backup & Restore so setup, semantics, verification, troubleshooting, and safety warnings stay in one operator-facing page.

## Verification
- `npm test -- --run`: 253 files, 3,571 tests passed on the behavior-fix head
- targeted `SyncTargetsCard.test.tsx`: 33 tests passed
- version touchpoint subset: 24 tests passed
- all three canonical version literals are `0.18.1-0149`
- ESLint, TypeScript, Vite build, strict MkDocs, and diff checks passed
- path classifier reports `code_paths_changed=true` and `docs_site_affected=true`
- PNGs visually inspected and audited for metadata and chunk structure; byte-oriented secret scanning is not presented as evidence for image pixels

## Tracking
- `enhancedchannelmanager-x4eoi` and `enhancedchannelmanager-s675b` remain in progress until this PR lands
- `enhancedchannelmanager-bt4r9` separately tracks preview invalidation after editing a target
- `enhancedchannelmanager-v7d37` was not exercised by the credential-free EPG fixture and receives no pass/fail conclusion
- `enhancedchannelmanager-ydmu3` remains separate and open


MotWakorb:
# Team review — scoped to this PR's diff

**Target:** `git diff origin/dev...HEAD`, merge base `8e9fb19b` — 5 files, +98/−20, docs-only.
**Tier:** home-lab (`docs`, `dev-tooling`) per `COMPONENTS.md`.
**Method:** eight domain reviewers, each fenced to this diff; every load-bearing claim below was then re-verified independently against source rather than taken from a reviewer's report.

Findings anchored in unchanged prose were demoted unless this PR's changes made that prose newly wrong — see *Scope notes* at the end for what was deliberately excluded.

## Verdict

The core work is correct and worth landing. This PR fixes a guide that documented six controls that do not exist, and corrects a security statement that was wrong in the dangerous direction. **All ten factual claims it adds were verified accurate against source.** The issues are concentrated in the two screenshots and in the acceptance record's numbers — not in the guide's facts.

## Findings

| # | Finding | Severity |
|---|---|---|
| 1 | Acceptance record counts do not reconcile and no baseline is given | Major |
| 2 | Screenshots captured against an environment `.gitignore` forbids for this directory | Major |
| 3 | Both screenshots depict states contradicting the prose they illustrate | Major |
| 4 | Step 2 dropped the inline "A overwrites B" warning text | Major |
| 5 | Acceptance record discloses `ydmu3` but is silent on `v7d37` | Major |
| 6 | Step 3 has no fallback if the scheduled task is not found | Minor |
| 7 | Three-gate model not reflected in Troubleshooting; `Enable` vs `Enabled` drift | Minor |
| 8 | Kill switch: "immediately" dropped; in-flight behavior unstated | Minor |
| 9 | `x4eoi`'s explicit "decide and state where the guide lives" never addressed | Minor |
| 10 | Both beads closed pre-merge, ~15s after commit | Minor |
| 11 | "15 destination collections" never enumerated; normalization recipe not reproducible | Minor |
| 12 | "independent documentation review: approved with no findings" is unverifiable from tracking | Minor |
| 13 | `detect-secrets-hook` claim covers the 3 `.md` files, not the 2 PNGs | Nit |

---

### 1. Acceptance record counts do not reconcile — `tests/dbas-test-env/README.md`

Fresh run reports preview `create 1,416` then apply `create 951`, with `failed 0` — a ~465-object shortfall under a clean bill of health. Repeat run reports preview `update 6` then apply `update 322`.

Neither gap is explained, and the record states no pre-sync baseline for B, so a reader cannot tell whether B ended up equivalent to A's plan. The section header says "Fresh-destination result" while the opening says the run "used the retained disposable environment" — these point opposite ways on whether B started empty.

This matters because **the immediately preceding acceptance record in this same file sets a higher bar** and shows exactly how to close the gap:

- an explicit pre-sync baseline read from B's own Postgres (`0 channels / streams / groups / profiles / memberships / logos / EPG sources`)
- an **"Every counter the run reported, checked against B"** table — including an `entities_blocked_by_dependency` row, which is precisely the field that would explain a create shortfall
- a **"Side-by-side, both read from Postgres"** A-vs-B table per category

The new record has none of these while carrying an **ACCEPTED** verdict. The run may well be sound; the defect is that it is published in a form no reader can check.

Related, lower severity (**#11**): the idempotency proof cites "15 destination collections" without enumerating them or the normalization recipe. Prior records in this file used 9 endpoints and 11 collections, also unenumerated — the count keeps changing and is never defined.

### 2. Screenshots do not meet this directory's documented capture bar

`.gitignore` carries an explicit, committed policy for exactly this path:

> Exception (bd-2mluw): `docs/images/user_guide/` is tracked. Those screenshots are captured against the **sanitized documentation environment** — placeholder hosts (`dispatcharr.example.local`), the `e2e_test` operator, pseudonymized provider names, empty credential fields … **Anything added there later must clear the same bar.**

These two were captured against the DBAS disposable environment; image 1 renders the disposable service host rather than the mandated placeholder. `git ls-files` confirms they are the **first two files ever tracked** under `docs/images/user_guide/backup-restore/`, so they set that directory's precedent.

**To be precise about what this is not:** every pixel of text in both frames was transcribed, and the PNG chunk structure was walked — no `tEXt`/`iTXt`/`zTXt`/`eXIf` chunks, all CRCs valid, zero trailing bytes after `IEND`. **No credential, provider host, stream URL, or PII is disclosed.** The crop discipline is genuinely tight. This is a control deviation with no disclosure — but the bypassed control is the only one this directory has, since this project's own `wfz8z` record documents an API key reaching a committed image precisely because byte scanners cannot read pixels.

### 3. Both screenshots depict states contradicting the prose they illustrate

- **Image 1** sits under "The target appears in the list and starts **Enabled**" — describing a target just created, before any run. The image shows `Success` and a `Last synced` timestamp, which only exist after a completed sync. An operator following the steps will see neither.
- **Image 2** sits immediately above "ECM now applies the sync automatically on that schedule," directly after the 7 steps that enable it. The image shows `Disabled`, `Schedule: Manual only`, `Next Run: Disabled` — the pre-configuration state. The alt text is honest about this, which makes the *placement* the problem rather than the caption.

Both read as the "before" screenshot doing duty as the "after" illustration. A caveat caption would resolve #3 without new captures, though it would not resolve #2.

### 4. Step 2 dropped the inline overwrite warning — `cross-instance-sync.md`

The removed text stated the consequence inline:

> A confirmation dialog warns you that **A's configuration overwrites B**.

The replacement points at it instead:

> Accept the browser's native confirmation only after you have read the source-wins warning. This confirmation is browser chrome and is therefore intentionally not pictured.

The doc no longer says where that warning appears or what it says. An operator who jumps straight to Step 2 can click through a generic native `confirm()` without having read a source-wins warning anywhere. Not pictured is fine; not stated is a regression on a destructive action.

### 5. `v7d37` is not addressed in the acceptance record

The record carves out `ydmu3` explicitly and correctly ("does not accept, waive, or close the UI bulk-creation defect"), but says nothing about **`v7d37`** — P1, in progress, "Cross-instance sync strips the XC guide URL entirely, so 53 of 59 channels land on the replica with no EPG link."

That bead's description says: *"Observed on Dispatcharr 0.28.2 — RE-CONFIRM ON 0.29.0 before fixing."* **This acceptance run was against 0.29.0 and reports EPG-source counts on B**, so it was the confirmation opportunity. Either it reproduced and should be disclosed alongside `ydmu3`, or it did not and that is a valuable result worth recording on the bead.

### 6–13, briefly

- **#6** — Step 3 sends the operator to Settings → Scheduled Tasks to find `Cross-Instance Sync: <target>`, with no fallback if absent and no statement of whether it is auto-provisioned on target creation. This is a new hand-off between two settings pages, so the dead end is new.
- **#7** — The new three-gate model is stated once; the Troubleshooting section still checks only the target toggle, and calls it **Enable** where the new text calls it **Enabled**. Partly anchored in unchanged prose, included only because the new model made that section newly incomplete.
- **#8** — The old kill-switch text said runs stop "immediately"; the new text is purely preventative and never says what happens to a run already in flight. Neither version documented it, so this is *less actionable*, not less accurate.
- **#9** — `x4eoi` instructed: *"DECIDE AND STATE where the guide lives … Propose with reasoning rather than assuming."* The walkthrough was placed inside the existing reference page with no stated rationale in the diff, PR body, or bead.
- **#10** — Both beads were closed seconds after the commit, while this PR is still open. Merging resolves it, but per `CLAUDE.md`, an open PR is not a shipped change.
- **#12** — The PR body's "independent documentation review: approved with no findings" cannot be reconciled with tracking data. Not evidence the review did not happen; it does mean the claim is uncheckable.
- **#13** — `detect-secrets-hook` was run on the three `.md` files, not the two PNGs. Per this repo's own `wfz8z` finding it would have proven nothing on binaries either way — the effective control is visual audit, which this review performed and which came back clean. Worth stating as "visually audited, frame transcribed" rather than citing a scanner blind to the artifact.

---

## Verified good

- **All ten added factual claims check out at source.** The headline reversal — "It is disabled by default" → "starts **Enabled**" — is correct and traceable end to end (`backend/routers/sync_targets.py:97` `enabled: bool = True`; `backend/export_models.py:130` `default=True`). The text it replaced was wrong.
- **The strong verb is earned.** "It will not save a recurring preview-only schedule" is backed by real server-side validation at `backend/tasks/dbas_sync.py:324-328`, which raises unless `confirm_apply is True` — not merely a client-side default.
- **The kill-switch strengthening is right.** The claim that a disabled target blocks *manual* preview/apply as well as scheduled runs holds server-side at `backend/tasks/dbas_sync_client.py:292`, with the UI buttons disabled as defence in depth.
- **The `index.md` correction is the highest-value change in the diff.** "credentials not synced" → "provider credentials sent on every cycle" fixes an under-warning in the dangerous direction: an operator reading the old line would conclude B was credential-free, when Xtream stream URLs carry the password in the path. No operator-facing page still carries the stale claim.
- `mkdocs build --strict` reproduces clean independently; both image paths resolve correctly in built HTML; both PNGs are tracked via the `.gitignore` negation; no dangling references to the removed "DR standby setup" heading anywhere in the tree; single commit, no working-tree leakage; all CI checks green.

## Open questions

1. **Screenshots** — re-capture against the sanitized documentation environment (fixes #2 and #3), add caveat captions (fixes #3 only), or drop the images and land the text?
2. **Acceptance record** — reconcile the 1,416 → 951 gap and add a pre-sync baseline before merge, or land now and track as a follow-up?
3. **`v7d37`** — record whether it reproduced on this 0.29.0 run?

## Scope notes

Deliberately excluded as outside this diff, recorded here without severity or remediation design: the sibling `backup-restore/index.md` lacking a credential warning; the pre-existing "not expected now" claim later in `cross-instance-sync.md` that appears to conflict with `v7d37`'s open status; the absence of CI image scanning (a higher-tier expectation, not a home-lab finding); and the per-target logo interval having no card toggle, which the docs describe accurately as a current gap.

No beads were created, closed, or modified by this review, and nothing was merged.

MotWakorb:
# Team Review: PR #927

**Depth:** Quick  
**Effective tier:** home-lab (`docs`, `dev-tooling`)  
**Participants:** Security Engineer, IT Architect, Project Manager, Project Engineer, UX Designer, Code Reviewer, Database Engineer, SRE, QA Engineer, Technical Writer

## Verdict: Changes Requested

The PR substantially improves the operator guide and its screenshots are credential-safe. The team found one block-level correctness issue in changed safety wording, plus four scoped documentation/evidence issues. This review is limited to PR #927 itself. No implementation work, tracker changes, or merge action was performed.

## Consolidated Findings

| # | Severity | Finding | Flagged by |
|---|---|---|---|
| 1 | **Block** | `docs/user_guide/backup-restore/cross-instance-sync.md:96` states that a failed preview does not make **Apply** available. That is only true before any successful preview. After a successful preview adds the target to `previewedIds`, a later failed preview does not clear it (`SyncTargetsCard.tsx:433-449`), so the stale **Apply** action remains visible. QA reproduced the success-then-failure sequence. The unconditional changed sentence is therefore false on a destructive-action path. | UX, QA; independently confirmed in source |
| 2 | **Warn** | The operator step at `cross-instance-sync.md:141-147` is not executable as written and overstates what its comparison proves. It does not identify the 15 collections, retrieval method, normalization fields, or comparison command. A normalized B-before/B-after comparison proves repeat-run stability for included fields; by itself it does not prove the following claim that B is equivalent to A. The acceptance record at `tests/dbas-test-env/README.md:1642-1646` preserves none of those details or a negative control. | UX, Technical Writer, SRE, Database Engineer, QA |
| 3 | **Warn** | `tests/dbas-test-env/README.md:1633-1636` calls the result “Logo acceptance” without naming the tested transport shape. The same README records that all 316 source logos were remote URLs and both Dispatcharr logo directories were empty (`:1349-1353`, `:1521-1522`). This run proves remote-URL replication and channel-logo binding, not ECM-local or Dispatcharr-hosted byte copying. | PM, Database Engineer, QA, Technical Writer |
| 4 | **Low** | The screenshot at `cross-instance-sync.md:110` appears after the enable-and-schedule procedure but shows **Disabled**, **Manual only**, and **Next Run: Disabled**. The caption is factually accurate, but the placement visually presents a pre-configuration state as the result of the preceding steps. | UX, Code Reviewer, Technical Writer |
| 5 | **Low** | The B-side verification step at `cross-instance-sync.md:139` checks that an EPG source exists, not that B has fetched usable programme data. The guide later distinguishes source arrival from guide ingestion, so this end-to-end step can pass while the replica guide remains empty. | SRE |

## Debate Resolution

- **Failed-preview wording:** the Technical Writer initially read line 96 as the clean-state case. QA tested the prior-success case, and source inspection confirmed failure paths never remove the target from `previewedIds`. Finding #1 stands.
- **Logo evidence:** the empty ECM upload directory is useful proof that no fixture was planted, but it does not exercise a byte-copy path. The finding is about the breadth of the new acceptance wording, not a claim that remote-logo replication failed.
- **Kill-switch wording:** the SRE questioned whether “kill switch” implies cancellation of an in-flight apply. The team did not retain this as a finding because the changed text says it blocks actions and runs and, unlike the removed wording, does not claim to stop an active run immediately.

## Positive Observations

- Current labels and scheduling controls match the UI: **Create target**, **Sync now (preview)**, **Enable task**, and **Apply changes on every scheduled run**.
- The guide clearly separates target enablement, task enablement, and child-schedule enablement.
- Both screenshots are tightly cropped, legible, and contain no provider credentials, stream URLs, raw properties, or sensitive metadata.
- Provider-credential transmission is disclosed plainly in the guide index.
- The acceptance record names exact Dispatcharr and ECM versions, reports non-zero repeat update counters honestly, and keeps `enhancedchannelmanager-ydmu3` explicitly outside acceptance.

## Disposition

Per the PO instruction, these findings are posted for PR-owner disposition only. No fixes were applied, no beads were created or changed, and the PR was not merged.

MotWakorb:
## Review remediation complete

Commit `ff8130a3` resolves the authorized findings from both team-review comments.

- Revokes stale **Apply** authorization whenever a new preview starts; success-then-failure regression coverage added.
- Replaces both screenshots with sanitized, state-accurate captures using `dispatcharr.example.local`.
- Restores the explicit source-wins overwrite warning and documents task auto-provisioning, startup reconciliation, all three scheduling gates, and in-flight disable behavior.
- Requires B-side programme verification after EPG refresh.
- Narrows the 2026-08-26 acceptance record to retained evidence; unexplained counters, unretained fingerprints, A/B equivalence, hosted-logo bytes, and `v7d37` are no longer overclaimed.

### Verification
- Full frontend suite: **253 files / 3,571 tests passed**
- Targeted SyncTargetsCard suite: **33/33 passed**
- ESLint, TypeScript, Vite build, strict MkDocs, diff check, and changed-file secret scan passed
- Replacement PNGs visually inspected and audited for metadata/chunk structure
- Independent immutable-head review: **Approved, no findings**; mutation check confirmed the new regression test fails when preview revocation is removed
- All GitHub checks on `ff8130a3` passed

No merge was performed.

MotWakorb:
# Team review — remediation verification (delta round)

**Target:** commit `ff8130a3` against the frozen acceptance matrix from the two prior review comments.
**Merge base verified current:** `merge-base HEAD origin/dev` = `git rev-parse origin/dev` = `8e9fb19b`.
**Tier:** home-lab (`docs`, `dev-tooling`, `ecm-core`) per `COMPONENTS.md`.
**Method:** six targeted delta reviewers — the personas whose findings this commit touches, not another ten-persona pass. Every headline closure was then re-confirmed independently against source rather than taken from a reviewer's report. Per the PO instruction, this round is **delta-only**: no dispositioned item was re-litigated and no new goalposts were added.

## Verdict

**All nine authorized findings are closed, one of them partially.** The remediation was verified by method, not by re-checking the examples the prior rounds cited. Three reviewers independently mutation-tested the code fix in disposable worktrees — reverting the five-line change turns the new regression test red while the other 32 stay green, so the test pins the actual defect rather than passing incidentally.

One item merits the PO's attention before merge, and it is a process gate rather than a defect.

## Matrix disposition

| # | Finding | Verdict | Evidence at `ff8130a3` |
|---|---|---|---|
| F1 | Stale **Apply** authorization after success-then-failed preview (Block) | **CLOSED** | `SyncTargetsCard.tsx:422-426` |
| F2 | Screenshots captured against an environment `.gitignore` forbids | **CLOSED** | Full visual + chunk audit, both PNGs |
| F3 | Screenshots depicted states contradicting their prose | **CLOSED** | Both frames re-captured and re-placed |
| F4 | Step 2 dropped the inline "A overwrites B" warning | **CLOSED** | `cross-instance-sync.md:101` |
| F5 | Step 3 had no fallback if the scheduled task is absent | **PARTIALLY CLOSED** | `cross-instance-sync.md:89,250` — see below |
| F6 | Three-gate model absent from Troubleshooting; `Enable`/`Enabled` drift | **CLOSED** | `cross-instance-sync.md:249-250` |
| F7 | In-flight disable semantics unstated | **CLOSED** | `cross-instance-sync.md:127-132` |
| F8 | B-side check proved source existence, not programme data | **CLOSED** | `cross-instance-sync.md:145` |
| F9 | Acceptance-evidence scope (A#1, A#5, A#11, A#13, B#2, B#3) | **CLOSED** | `tests/dbas-test-env/README.md:1622-1667` |

### F1 — closed structurally, not by patching branches

The revocation is the first statement in `handlePreview`, before `markBusy` and before any `await`. Because the id is dropped unconditionally at preview *start*, no failure branch has to remember to clear it: `success: false`, a thrown exception, a network rejection, and navigate/refresh mid-preview are all covered by construction, and a fourth failure path added later is covered for free. The only re-grant is inside `if (result.success)` at `:439`. Apply is a conditional *mount* (`:649`), not a disabled button, so revocation removes the affordance rather than greying it out — no "almost ready" signal on a source-wins overwrite path.

The test was probed three ways: head (33 pass), revocation deleted (the new test alone fails), and revocation restored with a re-add injected into the *failure* branch (two tests fail). The third probe is the one that matters — it proves the assertion evaluates post-resolution rather than catching a transient pre-response window.

**The doc sentence at `:102` was already true and was left alone. The author fixed the code rather than weakening the prose.** On a destructive-action path that is the right direction.

### F9 — narrowed by qualification, not by deleting evidence

This was the specific failure mode we checked for. Every inconvenient number survives verbatim — `create 1,416`, `create 951`, `update 6`, `update 322`, `skip 1,109`. What changed is the claim built on top of them: the verdict moved from **ACCEPTED** to **PARTIAL EVIDENCE ONLY**, a new *Audit limits* section states the preview/apply discrepancies "remain unexplained and must not be interpreted as either loss or convergence," "fresh destination" is marked not auditable, and the 15-collection fingerprint/idempotency claim is explicitly **withdrawn** with its reason given rather than quietly dropped. `v7d37` is now addressed directly (the fixture used the credential-free source, so the behavior was not exercised); the `ydmu3` carve-out is unchanged. Grepped the file for `ACCEPTED` and `Fresh-destination` — zero surviving hits.

### F2 — the images clear the bar

Both PNGs were rendered natively, then re-rendered magnified with overlap across the full frame and luma-equalized to expose low-contrast text; nothing appeared that was not already visible. Image 1 renders `https://dispatcharr.example.local` — the exact mandated placeholder, replacing the disposable service host that triggered the original finding. Complete text inventories were taken of both frames: no credential, provider name, stream URL, internal IP/port, or PII. Independent chunk walk on both files: `IHDR`/`IDAT`/`IEND` only, all CRCs valid, zero trailing bytes after `IEND`, no `tEXt`/`iTXt`/`zTXt`/`eXIf`, colortype 2 with no alpha channel. `git ls-tree` confirms the directory contains exactly these two files — replacement, not addition.

## The one partial closure

**F5 — the fallback's log-grep string misses one of the two failure paths (Low).** The new text is otherwise verified true against source rather than merely present: auto-provisioning on create is real (`sync_targets.py:439`), and startup reconciliation is real and unconditional (`main.py:1567-1568` → `dbas_sync.py:1055-1174`), so "restart ECM" is a working fallback for the common case. But the doc hands the operator the literal string `Failed to register sync task`, which matches `sync_targets.py:322` ("sync task **for target**") and does **not** match `dbas_sync.py:1206` ("sync **target tasks**"). The word order differs; the grep returns nothing.

The reviewers split on this — one marked F5 closed on the string existing, the other tested whether it covered both paths. We ran the grep against both literals to adjudicate: zero matches on the second. The gap is narrow (a whole-reconciliation crash, not a per-target failure) and a looser `grep "register sync"` still finds it, but it is precisely the failure mode the "restart ECM" advice is meant to diagnose when the restart doesn't help. Backlog candidate — widen the grep guidance. Not a blocker.

## New this round

Only items meeting the delta bar — a regression introduced by the delta, or new evidence about the delta's own added text. Nothing pre-existing was promoted.

1. **No CHANGELOG entry and no version bump (Warn — process).** All three version touchpoints read `0.18.1-0148` on both base and head, and `CHANGELOG.md` is untouched across the whole PR. The *CHANGELOG* half is delta-introduced: before `ff8130a3` this PR had no behavior change to record. A silent Apply-authorization change landing under a `docs(sync):` title is exactly what `[Unreleased]` exists to catch (`docs/shipping.md:90`). See the decision below.
2. **F5's grep string** — above.
3. **`SEC-DELTA-01` — the same stale-authorization class survives the edit path (Low, P2, pre-existing).** Flagged independently by security and code review. `setPreviewedIds` is mutated at four sites; `handleSave`'s edit branch is not among them. Preview target → Apply appears → operator notices a wrong `base_url` (the very thing a preview surfaces) → edits and saves → Apply is still mounted and will overwrite a destination that was never previewed, pushing provider credentials. The server does not catch it: editing `base_url` bumps `credential_version`, but manual applies from the card send no captured version, so that comparison is skipped. Mitigated by admin-only access, MCP principal refusal, and a `confirm()` that names the target and states credentials are leaving. **Pre-existing, not a regression — the delta closes one of two paths in this class and weakens nothing.** Follow-up bead, not a change request against this PR.

**Nits, recorded not blocking:** the five added lines carry no comment in a file whose house style is dense "why" commentary, and the module docstring's "`result.success` is the ONLY thing that unlocks Apply" is now incomplete; the delete/preview/apply set-removal block is now its third verbatim copy where a `revokePreview(id)` helper would sit naturally beside `markBusy`/`clearBusy`; the test's `waitFor(runTask called 2 times)` guard is close to a no-op since the call is synchronous; `cross-instance-sync.md:102` states the weaker half of the behavior (Apply withheld) rather than the newly shipped half (a granted Apply being withdrawn); and an operator returning after the error toast dismisses has no on-row explanation for why Apply vanished.

## Deliberately not re-opened

Per the PO's instruction, these stay dispositioned and are recorded without severity or remediation design: the beads closed pre-merge (A#10); the unverifiable independent-review claim (A#12); the guide-placement rationale (A#9, now stated in the PR body); and every item in the prior round's *Scope notes*. The absence of a server-side "fresh preview required" precondition is an operator-safety interlock rather than an access control — an actor who can send `confirm_apply=true` is already an admin who could run a preview first — and at home-lab tier it is not a finding. Out of scope, unrelated to this PR: a local `tmp/pr-927-remediation-opencode` branch exists in the author's checkout and is unpushed.

## Verified good

- CI green on `ff8130a3`: all required checks pass. Targeted suite independently re-run at **33/33**; `tsc --noEmit` and ESLint at `--max-warnings 0` both clean; `mkdocs build --strict` reproduces clean with both images resolving in the built site.
- Every operational claim the commit *added* was checked against source rather than taken on the doc's word — auto-provisioning, startup reconciliation, all three scheduling gates, and the in-flight disable semantics each map to real code, and each quoted UI label (`Enabled`, `Enable task`, `Enable this schedule`, `Apply changes on every scheduled run`) matches the literal string in the component that renders it.
- The in-flight claim is a correct, previously-undocumented operational fact: the enabled check is pre-flight only, so a disabled target genuinely does not cancel a run already applying.
- Both replacement screenshots changed dimensions, which is positive evidence of fresh captures rather than pixel-patched originals. Choosing a never-run target pointed at a non-resolving placeholder host removes the sanitization problem instead of redacting around it.
- Every doc edit in this commit narrows an overclaim rather than adding one.

## Decision needed

**CHANGELOG entry and version bump.** This is the only item touching merge readiness, and it is the PO's call, not the team's:

- **Option A — add them before merge.** Bump the three touchpoints and add an `[Unreleased] → Fixed` entry for the Apply-revocation behavior. Follows `docs/shipping.md` as written; costs one more commit and CI cycle.
- **Option B — waive for this PR.** Defensible at home-lab tier if the behavior change is judged too small to surface to operators, but it leaves a user-visible fix unrecorded under a `docs(sync):` title, which is where changelog drift starts.

The team's recommendation is **A** — the change is operator-visible on a destructive path, and that is the category `[Unreleased]` is for.

**Everything else is disposed.** F5's grep widening and `SEC-DELTA-01` are backlog candidates awaiting a file/park/drop call; the nits are recorded and need no action. No beads were created, closed, or modified by this review, no code was changed, and nothing was merged.

PR #928, opened 2026-08-26
v0.18.1-0150: Preserve bulk-created channel stream assignments

MotWakorb:
## Summary
- stage every bulk-created channel and its source-stream assignments atomically through the real App callback
- keep each create and its negative-temp-ID assignments in one bulk request, with unique IDs and fail-closed completeness checks
- reject old persisted ledgers whose bulk-create stream intent cannot be proven
- validate temp-channel streams and prevent unresolved negative channel IDs from reaching Dispatcharr after failed or malformed creates

## Tracking
- enhancedchannelmanager-ydmu3

## Evidence
- retained logs from the affected `8e9fb19b` run show one 316-operation request containing creates only; no assignment request reached ECM
- the exact historical trigger remains unproven, so this PR does not claim temp-ID remapping caused that run
- a StrictMode App-level regression now drives the production bulk-create callback with 316 one-stream channels and proves 316 unique create IDs, 316 matching assignments, 632 persisted operations, and 632 submitted operations
- Apply refuses an incomplete bulk-create pair before any request, and ledger format v3 rejects older records whose missing stream intent cannot be reconstructed safely

## Verification
- frontend full suite: 3,580 passed
- frontend lint: passed
- TypeScript: passed
- production frontend build: passed
- canonical backend gate: 11,671 passed, 3 skipped, 2 deselected
- version touchpoint consistency: 24 passed
- independent code review: approved; no blocking findings remain

## Acceptance Status
Automated coverage now reproduces the original 316-channel operation shape through the real App composition boundary and proves the new invariants. The bead remains open until its separate live UI acceptance criterion confirms the resulting Dispatcharr channel-stream associations.


MotWakorb:
## Team Review — PR #928

Ten-persona review, quick depth. **Effective tier: small-team** (strictest-wins per `COMPONENTS.md`: ecm-core/dev-tooling are home-lab, the version touchpoints span ci-release). Scope was fenced to this PR's diff; out-of-scope observations are listed separately and carry no severity.

**Verdict: no Block.** The partition is correct and the change is safe to merge. But the review could not confirm *why* the pre-fix code failed, and that has consequences for closing the bead — see Finding 1.

---

### Positive observations

- **The partition is provably exact.** `createPhaseOps` uses `A || B` and `otherOps` uses `!A && !B` — a De Morgan complement, so the two sets are mutually exclusive and exhaustive *by construction*. No operation can be dropped or double-sent. This was the property with the most damage potential and it holds.
- **Ordering is safe, and defended server-side.** `_consolidate_operations` re-emits `ordered_ops` (creates, in order) before stream add/remove ops (`channels.py:2206-2214`). Since the frontend always sends `consolidate: true`, a create always precedes its dependent assignment regardless of staging order.
- **`channelCreateOps` exists solely to keep the operator-facing message honest** — the author could have reused `createPhaseOps.length` and inflated the channel count, and didn't.
- **The reorder-permutation guard is what makes this safe.** Splitting `addStreamToChannel` from `removeStreamFromChannel`/`reorderChannelStreams` inverts their relative order; `validate_stream_permutation` (`channels.py:3600-3607`) makes a truncating reorder fail loudly rather than silently detach.
- Version touchpoints are clean — all three carry `0.18.1-0150`, zero `0149` strings remain.

---

### Important / Should-fix

**1. (Medium) The stated fix mechanism is not supported by the pre-fix code, and the root cause remains unestablished.**

The CHANGELOG says assignments now "stay in the create request so their temporary channel IDs resolve." But pre-fix, `useEditMode.ts`'s `otherOps` loop *already* resolved temp ids client-side before the request left the browser. A reproduction on reverted code captured the pre-fix payload as `{"channelId": 900, "streamId": 55}` — an already-resolved **real** id, not an unresolved temp id.

Both new frontend tests do fail pre-fix, but examine *why*. Test 1's mock builds `tempIdMap` only from creates in the same request, then rejects any `addStreamToChannel` whose `channelId` isn't a key in it (`test.ts:96`, `FIRST_REAL_CHANNEL_ID = 900`). Pre-fix, request 2 has no creates → `tempIdMap` is `{}` → the correctly-resolved real id `900` is rejected as "not resolvable in this request." **The real backend would call `get_channel(900)` and succeed.** So test 1 fails pre-fix for a reason the backend does not exhibit.

Test 2 is sounder — it asserts request *composition* (request 1 must contain `[createChannel, addStreamToChannel -1]`), a faithful structural check on the new partitioning.

Net: the tests lock in the new design; they do not reproduce the reported defect (316 channels, 0 stream rows, false success). The backend test also passes pre-fix (`channels.py` has zero changed lines), so it is a contract/characterization test, not the "backend temporary-ID integration regression" the PR body calls it.

*This is not a merge blocker — the new behavior is correct either way. It matters because bead `ydmu3`'s acceptance criterion #1 is a **live UI run**, and nothing here substitutes for it. Please don't close ydmu3 on this PR's automated evidence alone.*

**2. (Medium) A failed create now sends `get_channel(-1)` upstream, and the operator sees `(Channel -1)`.**

`addStreamToChannel` calls bare `resolve_id(op.channelId)` (`channels.py:3536`); only `setProfileMembership` wraps it in `reject_unresolved_channel` (`:3922`). When a create fails, `resolve_id` returns `-1` unchanged and it goes straight to `client.get_channel(-1)`.

The failure is then reported using the **raw, unresolved** id: `error_details["channelName"] = f"Channel {op.channelId}"` (`channels.py:4029`), and `existing_channels` is keyed only by real ids. The operator gets `(Channel -1)` — an internal staging id they have never seen — instead of the channel name.

Not a regression in kind, but this PR makes the backend-resolution path the *exclusive* route for these ops, so it goes from edge case to primary path. Neither new test covers a failed create with a dependent assignment riding along.

**3. (Low-Medium) Stream-existence validation is now unreachable for exactly the ops this PR reroutes.**

`channels.py:3054` nests the `stream_is_missing` check inside `elif op.channelId >= 0:`. Pre-fix these ops arrived with positive real ids and got the check; now they arrive negative and neither branch fires. `referenced_stream_ids.add(op.streamId)` (`:2885`) is unconditional, so the lookup is already paid for and discarded — the fix is a branch restructure at zero upstream cost.

One fact decides the severity and **was not verified**: whether Dispatcharr 0.29 rejects an unknown stream id in the `streams` array. If it accepts it, this reproduces the same silent-success class the bead was filed for, narrowed to a delete-during-edit race.

---

### Improvements / Could-fix

**4. The create-phase request is unbounded** — flagged independently by 7 of 10 personas. `BATCH_SIZE = 200` applies only to `otherOps`; `createPhaseOps` has no client cap and `BulkCommitRequest.operations` has no server `max_length`. The bead's own repro (316 channels) roughly doubles to ~632 ops in one call. Mitigating: this rides the 202+poll job path, so no HTTP timeout applies, and no proxy/uvicorn body limit exists to cross. Creates were already unbatched pre-PR — this widens an existing shape rather than creating one. Not blocking at this tier; at startup tier it would be a finding.

**5. The progress label no longer describes its own request.** `Creating ${channelCreateOps.length} channels...` labels a request that also carries N stream assignments. The bar stays monotonic and reaches 100% (verified), but for the 316-channel case the operator reads "Creating 316 channels" while ~632 operations are in flight behind one indivisible step.

**6. `docs/api.md:101` doesn't document the contract this PR makes load-bearing.** The `setProfileMembership` row (`:108`) spells out the same-batch negative-temp-id contract; the `addStreamToChannel` row has an empty Notes column. One line, mirroring the existing text.

**7. The partition predicate is written twice, once negated.** The De Morgan property that makes Finding-worthy drops impossible survives only while both copies stay in step. Extracting one `isTempChannelStreamAssignment(op)` predicate and filtering on `p` / `!p` makes exhaustiveness structural rather than by-inspection.

**8. Two PR-body claims don't check out.** Actual CI: **11,643** backend passed (body says 11,642) and **3,573** frontend (body says 3,572) — both green, but the numbers written as evidence weren't re-derived from the gating run. And "independent code review: approved, no remaining findings" has no artifact behind it: `gh pr view 928 --json reviews` returns empty. If that was an off-platform review, link it.

**9. CHANGELOG undersells the bug.** The bead records a *silent total loss with a false success report*. Neighboring entries in the same section name that severity explicitly ("diverged silently and permanently"); this one leads with mechanism.

---

### Conflicts resolved during synthesis

- **Unresolved-temp-id handling:** the DB engineer reported `reject_unresolved_channel` protects this path; the code reviewer said it doesn't. **Code reviewer is correct** — `grep` shows exactly one call site, on `setProfileMembership` only (`channels.py:3922`).
- **Are the new tests genuine regression tests?** QA verified both fail pre-fix; the technical writer showed test 1 fails for an unfaithful reason. **Both facts hold** — resolved in Finding 1.
- **Is the fix incomplete for `updateChannel` / `removeStreamFromChannel` / `reorderChannelStreams` on temp channels?** Raised by the engineer and architect. **Assessed as not defective**: those op types still receive client-side temp-id resolution in the unchanged `otherOps` loop, so they are not broken the way `addStreamToChannel` was reported to be. Worth a comment noting why they're deliberately left in the split-phase path.

### Out-of-scope observations (no severity, no remediation proposed)

- `_prune_old_bulk_commit_jobs` (`channels.py:2370-2375`) prunes by `created_at` regardless of whether a job is still running.
- `str(e)` from httpx propagates the upstream Dispatcharr URL into browser-visible error strings across this endpoint.
- No uniqueness constraint on `tempId` within a single bulk-commit request.
- Retry after a job-level failure is not idempotent for creates (pre-existing; documented in-code).

---

*Review calibrated to small-team tier. Enterprise-grade concerns (chunking policy, operations-count caps, SLO/alerting) were deliberately not filed as findings.*

MotWakorb:
## Review remediation

Addressed the substantive findings from the team review:

- **Finding 1:** retained logs now anchor the diagnosis honestly: the affected run submitted 316 creates and no assignments, while the exact historical trigger remains unproven. The production App callback now stages each create and its streams atomically. A StrictMode regression drives 316 one-stream channels through that callback and proves 632 persisted/submitted operations with unique one-to-one temp-ID mappings. Apply also fails closed on incomplete pairs, and ledger format v3 rejects older records whose stream intent cannot be reconstructed.
- **Finding 2:** every channel-ID-bearing executor path now rejects an unresolved negative ID after a failed or malformed create. Errors retain the submitted temp ID, name the intended channel when known, and no negative ID reaches Dispatcharr.
- **Finding 3:** missing-stream validation now covers temp-channel assignments and is enforced before any channel read/update even with `continueOnError=true`; a failed stream-catalog lookup still does not falsely accuse the stream of being absent.
- Updated `docs/api.md`, staged-ledger architecture documentation, and the changelog to match the proven guarantees without claiming an unsupported root cause.

Fresh local gates: backend 11,671 passed; frontend 3,580 passed; lint, TypeScript, production build, and 24 version checks passed. Independent follow-up review approved the implementation. The bead remains open pending its separate live UI acceptance run against Dispatcharr.

MotWakorb:
## Live acceptance and final remediation

Validated the PR in an isolated environment against Dispatcharr 0.29.0.

- **Success pole:** the UI staged and applied 316 one-stream channels. Direct Dispatcharr verification found 316 channels, 316 links, 316 distinct linked streams, zero name mismatches, one link per channel, and channel numbers 1000-1315.
- **Known-bad pole:** after staging, stream 317 was deleted. The first run exposed a partial-apply defect: `continueOnError=true` created the sentinel channel despite the missing coupled stream.
- **Remediation:** commit `2794b23b` now aborts before execution when a missing stream is coupled to a negative temporary channel. Repeating the known-bad pole surfaced missing stream 317 in the UI and left Dispatcharr unchanged at 316 channels/316 links; no sentinel channel or channel number 2000 was created.

Verification for the final head:

- Independent focused code review: approved with no findings.
- Local focused regression: 4 passed.
- Local full backend gate: 11,672 passed, 1 skipped, 2 deselected.
- Fresh GitHub backend, frontend, CodeQL, ECM image builds, and ECM container scans: passed on `2794b23b`.
- **Current external gate blocker:** both MCP container scans found Alpine `libcrypto3`/`libssl3` 3.5.7-r0 affected by HIGH CVE-2026-14456, fixed in 3.5.8-r0. The MCP image is unrelated to this backend diff, but the PR remains blocked until its pinned base image is refreshed.

The historical trigger for the original 316-create/zero-assignment event remains unproven; this evidence demonstrates the corrected success path and fail-closed behavior without overstating that diagnosis.

MotWakorb:
The unrelated MCP image blocker now has an isolated fix in PR #929: https://github.com/MotWakorb/enhancedchannelmanager/pull/929

PR #929 exact head `f4259a9a` passes native AMD64 and ARM64 MCP builds and both MCP Trivy scans, plus the full backend/frontend/MCP/CodeQL gates. PR #928 remains intentionally unmerged and `UNSTABLE` until #929 is explicitly authorized, merged to `dev`, and #928 is updated/revalidated against that base.

MotWakorb:
## Final integrated-head validation

PR #928 is now updated to exact head `c92e2b68a289e35a3e3f253c79d905cb5ca061fe` on merged `dev` base `2dce4cd4` and build `0.18.1-0152`.

Final remediation establishes a backend-authoritative ordered bulk contract: create temp IDs are unique and strictly negative; every negative reference must name one earlier create; consolidation preserves add/remove/reorder order; effective plans are revalidated before mutation; and expected stream intent is checked against the ordered final state regardless of `continueOnError`.

Verification on this exact head:

- Independent full code review: approved with no findings.
- Independent full database-integrity review: PASS with no findings.
- Local backend gate: 11,697 passed, 3 skipped, 2 deselected.
- Local frontend: 3,585 passed; ESLint, TypeScript, and production build passed.
- SBOM verify/audit and MCP supply-chain policy: passed.
- Fresh GitHub backend, frontend, MCP, CodeQL, docs, performance, AMD64/ARM64 ECM and MCP builds, dependency/security scans, and all ECM/MCP container scans: passed.
- GitHub reports the PR `CLEAN` and `MERGEABLE`.

Residual limitation: the live Dispatcharr 0.29.0 316-channel success/failure acceptance was completed on earlier head `2794b23b`; the subsequent ordered-plan hardening is covered by focused and full automated tests but was not repeated against a live container. The original historical trigger remains unproven.

This PR remains intentionally unmerged pending separate explicit merge authorization.

MotWakorb:
## Merged and published

PR #928 was squash-merged into `dev` as `4cca7d994610c8ff552e9ea4185f1c2860fd500a` after explicit authorization.

All post-merge workflows passed on that exact commit:

- Tests, including backend, frontend, MCP, and exact-SHA publication.
- AMD64/ARM64 ECM and MCP builds, dependency/security scans, CodeQL, all four container scans, and candidate attestation.
- Verified AMD64/ARM64 image promotion and multi-architecture manifest publication.
- Docs build/deploy and performance benchmark.

Bead `enhancedchannelmanager-ydmu3` is closed. The historical trigger for the original incident remains unproven; the corrected live success/failure poles and the final ordered-plan automated coverage are the evidence basis.

PR #930, opened 2026-08-27
v0.18.1-0158: Linearize MCP credential lifecycle (s6pg3)

MotWakorb:
## Summary
- make the sidecar api-key file the single MCP client credential authority
- serialize startup, settings saves, restores, rotation, and revocation across processes
- add crash-safe prepared/recovery-active WAL handling and explicit indeterminate 503 responses for compound storage failures
- align backend file validation with the sidecar and automatically repair the settings mirror

## Verification
- backend gate: 11,749 passed, 3 skipped, 2 deselected
- MCP server: 753 passed
- frontend: 3,588 passed; lint, typecheck, and production build passed
- focused lifecycle/projection/router/version: 220 passed
- SBOM verify/audit and MCP supply-chain policy passed
- live container gate passed at PUID/PGID 1000:1000 and 1571:1572
- scoped correctness and security review completed; no accepted Block-level findings remain

## Dependency
- prerequisite for enhancedchannelmanager-71von

Bead: enhancedchannelmanager-s6pg3


MotWakorb:
# Team Review — PR #930 (`v0.18.1-0158`, bead `enhancedchannelmanager-s6pg3`)

**Depth:** full · **Effective tier:** small-team (strictest-wins — `ecm-core`/`mcp-server`/`data-store` home-lab + `ci-release` small-team via regenerated `sbom/dev/*`)
**Head reviewed:** `723e61c12356c22ff25c2b300fb065b0e5aa6f8a` · **Merge-base == `origin/dev` tip** (`4cca7d99`), branch is current · **CI: 25/25 green**
**Participants:** security-engineer, it-architect, code-reviewer, project-engineer, database-engineer, sre, qa-engineer, project-manager, ux-designer, technical-writer

## Verdict: **Request changes** — 4 blockers

**Scope note, stated up front:** every blocker below is anchored to an acceptance criterion this bead already set, not to a requirement imported by the review. B2 fails the bead's central criterion verbatim ("Once Kn linearizes, no path can republish Kn-1"). B1/B3 fail "Existing 0600 ownership, atomic file replacement … fresh-install, upgrade, non-object settings intact." B4 is the test for the bead's own "startup-vs-transition" criterion. No persona proposed a different architecture, and no blocker asks for work outside a function this PR already rewrites. The excluded PUID=0 hardening finding stayed excluded — confirmed, the diff touches no Dockerfile, entrypoint, or ownership-model file.

This is careful, high-quality work with real engineering discipline in it (see *Done well* — it is long, and earned). The linearization design is correct: the single-authority property holds, and one reviewer sustained 392 rotations against a hostile peer process sampling the artifact 585,215 times with **zero** stale-key observations. What blocks is not the design; it is the blast radius of its failure paths and one abort-path bug that defeats the invariant the design exists to protect.

---

## Blockers

### B1 — `get_settings()` can now raise, and both startup *and* every `/api/*` request call it unguarded
**Flagged independently by 7 of 10 personas** (security, architect, engineer, SRE, database, code-reviewer, QA). Reproduced separately by four of them.

`_read_mcp_api_key_locked` (`backend/config.py:1171-1177`) catches only `FileNotFoundError`. `_validate_mcp_file_metadata` (`config.py:1118-1124`) raises `PermissionError` when the mode is not exactly `0600`, when `st_uid != os.geteuid()`, or when `st_nlink != 1`; `O_NOFOLLOW` raises `OSError` on a symlink. All of it propagates through `load_settings` (`config.py:1554-1557`) and out of:

- **`backend/main.py:1152`** — `get_settings()` in `startup_event`, unguarded. An exception out of a startup handler aborts the ASGI lifespan and uvicorn exits → container restart loop.
- **`backend/main.py:592`** — `public_mcp_key = get_settings().mcp_api_key` in `auth_middleware`, **before** the `AUTH_EXEMPT_PATHS` check at `:610`. Every `/api/*` request 500s, `/api/health` included — the Dockerfile HEALTHCHECK target. An exception raised in an `@app.middleware("http")` before `call_next` also bypasses `@app.exception_handler`, so this is a bare 500, not the 503+`Retry-After` at `main.py:1098-1113`.

**This is a regression against a policy `dev` had written down.** At base, `origin/dev:backend/config.py:1119-1126` wrapped the load-path projection in `try/except OSError` with the rationale spelled out: *"A broken optional sidecar projection must not make ECM discard otherwise-valid settings and boot with defaults. Rotation/save remains fail-closed because save_settings does not suppress this error."* That read-path/write-path split is gone at head. And `main.py:1157-1162` — five lines below the new call — still states the opposite policy for the *sibling* credential in the *same directory*: *"Must not raise: an exception out of a startup handler aborts the ASGI lifespan and uvicorn exits, so a projection directory ECM cannot write would take the entire application down rather than degrade the optional sidecar."* Two adjacent credentials of one subsystem, opposite failure policies, one file.

**Reachable by a supported operator action.** `backend/entrypoint.sh:342` chowns the MCP projection directory **non-recursively**, unlike `:380`'s `chown -R appuser:appuser /config`. Change `PUID`/`PGID` on an install running the MCP overlay (a config this PR's own live gate exercises at 1000:1000 *and* 1571:1572) and `api-key` keeps the old uid while the directory-writability probe at `:348` still passes and prints "is writable". ECM then never starts. Other confirmed triggers: mode drift from a host-side restore, `rsync --link-dest`/rsnapshot producing `st_nlink != 1`, a symlink, and a malformed recovery record (B3).

Reproductions returned, against the branch's real module: `chmod 0644` → `PermissionError: api-key does not have the required owner and mode`; `os.link` → `PermissionError: api-key has an unsafe link count`.

**The team is unanimous on the shape of the fix, and on what it must *not* weaken.** Keep the strict validation. Keep the refusal to self-repair a file ECM does not own — silently re-owning it would be the worse bug. What changes is scope: a *read* that cannot validate should return "no usable authority", which is exactly what `config.py:1175-1176` already does for `FileNotFoundError`, a case with an identical security property (ECM has no key; the sidecar's state is unknown to ECM). Every transition path stays fail-closed and untouched. The security posture is byte-identical; only the blast radius moves, from "MCP is unavailable and loudly logged" to "the product will not boot."

The security engineer's framing is worth quoting directly: the projection directory is `0700` owned by ECM's uid, so any principal that can chmod or hardlink `api-key` can already read it — the refusal buys ~0 against a threat actor at this tier while costing total availability. And a control whose failure mode is total unavailability gets disabled by operators, and a disabled control protects nothing.

### B2 — A failed transition deletes the *previous* transition's still-undurable redo record, republishing K(n−1) after K(n) committed
**Database engineer, reproduced end-to-end; independently derived by the security engineer (SEC-930-04).** This one fails the bead's headline criterion.

`_stage_mcp_recovery_locked` unconditionally `os.replace`s over any existing record (`config.py:1333`), and its abort path calls `_remove_mcp_recovery_locked()` (`config.py:1337`), which unlinks the record outright — **including one it did not create**. Nothing resolves an outstanding record before a new transaction begins: `_transition_mcp_api_key` (`config.py:1615-1621`) goes straight to `_publish_mcp_api_key_locked`; only `load_settings` (`:1556`) ever calls recovery.

Sequence, with directory fsync failing:

1. Rotate to K1. `os.replace` commits, `_fsync_mcp_parent_directory` fails → `durable=False` → `config.py:1363`'s `if committed and durable` is False → the `recovery-active(K1)` record is **kept**, which is correct, and the operator gets **200 with K1** (not indeterminate — the WAL record covers the missing fsync).
2. Rotate to K2. `_stage_mcp_recovery_locked:1333` replaces the record, destroying `recovery-active(K1)`; the stage fsync fails; `:1337` then unlinks it entirely. Operator sees an error and reasonably reads it as "still on K1".
3. `api-key` on disk is K1 with an un-fsynced dirent and **no WAL record**. A host crash reverts the dirent → K0 is live again.

Reproduction output:
```
rot1 -> K1: kL9ZD-94 | api-key on disk: kL9ZD-94 | WAL: recovery-active | WAL key is K1: True
rot2 -> RAISED OSError [Errno 5] MCP recovery record is not crash-durable
after rot2: api-key on disk is K1: True | WAL: <REMOVED>
POST-CRASH: ECM reports K0 | sidecar reads K0
```

K1 linearized onto the artifact and a later *failed* transition republished K0. Direction: resolve any outstanding record (replay or discard) before staging a new one, and never unlink a record this call did not write.

**Related, same area (database engineer, DB-3):** the record is `{"key": <new>, "state": ...}` — new value only, no undo image. `prepared` therefore means "discard", which is not a rollback: it leaves `api-key` at whatever the filesystem happened to persist, in the window between `os.replace` (`:1409`) and `_mark_mcp_recovery_active_locked` (`:1416`). Reproduced with a revocation of a leaked key: after the crash, `REVOCATION SURVIVED: False` — the leaked key is live again and the record that could have decided it is deleted. Carrying the pre-image is the standard fix and resolves B2 and DB-3 together.

### B3 — A corrupt or foreign-owned `.api-key.recovery` permanently bricks ECM and never self-heals
**Security, database, engineer, code-reviewer.**

`_recover_mcp_authority_locked` (`config.py:1368-1395`) raises `ValueError` on unparseable JSON (`:1379`) or an invalid shape (`:1390`), and lets `PermissionError` out of `_read_private_mcp_file_locked`. It **does not remove the record**, so every retry fails identically — combined with B1, a permanent crash loop. Confirmed: first `load_settings()` → `ValueError`, record still present, second → identical `ValueError`. No path in the codebase removes it.

The `recovery-active` flip (`_write_mcp_recovery_document_locked`, `config.py:1258-1294`) is the one write in the module that is *not* temp+rename — `O_WRONLY` + `seek(0)` + `write` + `truncate` in place — so a torn record is reachable. Credit where due: a sweep of all 39 torn byte-prefixes produced only `{accepted, ValueError}`, never a valid-but-wrong record, so detection is sound. The problem is the response to detection.

**Zero test coverage** — `grep "invalid shape\|not valid JSON\|JSONDecode\|torn\|garbage"` over the new module returns 0 hits; the `parametrize` corrupt cases at `:306` are for `settings.json`, not the WAL. **Zero documentation** — `git grep "api-key.recovery" -- docs CHANGELOG.md` returns 0 hits, so the remedy is deleting a dotfile that appears in no doc, no log line, and no error message the operator can see, because the process is dead.

The database engineer's argument is the decisive one and we adopt it: the recovery record is not a credential, it is bookkeeping *about* a credential, and its only job is to survive the crash that damaged it. Every real WAL — SQLite, Postgres, InnoDB — discards a tail record that fails its integrity check and continues, because the alternative is that one bad sector during an unrelated power cut ends the deployment. A safe default already exists four lines away at `:1391-1393`.

### B4 — The `[rotate]` case of the startup-vs-transition test is a tautology and passes against pre-fix code
**Code reviewer, proven by mutation test.** This is the test for the bead's own "startup-vs-transition" criterion, and it is one of the 25 green checks.

`backend/tests/unit/test_mcp_credential_lifecycle.py:179-182`:
```python
expected = "" if transition == "revoke" else _authority(authority_file)

assert _authority(authority_file) == expected
assert _stored_mirror(settings_file) == expected
```
For `transition == "rotate"`, `expected` is read from the authority file *after* the stale `save_settings` at `:178`. Line 181 is `x == x`; line 182 only requires the mirror to agree with whatever the authority happens to hold. If the stale startup snapshot republished `"startup-key"`, both assertions still pass.

Mutation evidence: replacing `save_settings`'s `stored.mcp_api_key = key` with pre-fix semantics (caller's key wins) → **4 failed, 46 passed**, with `[revoke]` failing and `[rotate]` passing; re-run in isolation, `[rotate]` → `1 passed` under the mutation. Unmutated module passes 50/50 on five consecutive randomized runs.

Direction: capture the key `peer_transition()` produced and compare against that value, as `test_two_rotations_serialize_and_load_observes_last_authority:203` already does.

---

## Important (should-fix; PO's call whether in this PR or a fast-follow)

| # | Finding | Flagged by | Anchor |
|---|---|---|---|
| H1 | **WAL temporaries escape the orphan sweeper.** `config.py:1207` globs `.api-key.*.tmp`; `config.py:1217` applied to the recovery path produces `..api-key.recovery.<hex>.tmp` (leading double dot) — no match, verified. A SIGKILL between `os.open` and the `finally` unlink strands a `0600` plaintext `{"key":"<new key>","state":"prepared"}` in the directory the sidecar mounts, permanently. Exactly the leak `sweep_orphaned_settings_temporaries` (`config.py:250-253`) was written for. Also: the MCP sweeper runs only inside `_publish`, never at startup, unlike its settings counterpart at `main.py:1145`. One-line glob fix. | security, database, code-reviewer | `config.py:1205-1213`, `:1217`, `:1328` |
| H2 | **Invalid `settings.json` puts a cross-process flock on every settings read.** `_repair_settings_mirror_locked` returns `dirty=True` unconditionally for `settings_state == "invalid"` (`:1512-1513`) with nothing to repair, and `:1545` gates the cache on `not dirty`, so the flag can never clear. Measured three ways: 200 flock acquisitions per 200 loads; 500 per 500; 27× per-call slowdown. Base cached defaults once. Amplified by `main.py:592` putting this on every `/api/*` request and by the 5 s `LOCK_NB` budget blocking the event loop; in TLS mode two `main:app` processes then contend, turning a degraded config into `SettingsWriteTimeout` 503s across the API. Bead AC names "non-object settings behavior remain[s] intact". | security, database, code-reviewer | `config.py:1512-1513`, `:1545` |
| H3 | **Directory-fsync refusal is fatal in WAL staging only.** `config.py:208-235` and `:1180-1202` treat it as best-effort, citing SQLite and Git; `config.py:1413` treats the authority's own dir fsync as soft. `_stage_mcp_recovery_locked:1334-1335` makes it fatal — and because `_initialize_authority_locked:1500` routes first-install through the same path, fresh install cannot boot on such a filesystem. Reproduced: `get_settings() RAISED OSError [Errno 5]`, config dir left holding only `.settings.lock`. Base performed no directory fsync at all here. Bead AC names "fresh-install ... intact". Consequence worth noting: on a filesystem that refuses it consistently, the transition fails earlier and the whole `MCPApiKeyDurabilityIndeterminate` path is dead code. | engineer, database | `config.py:1334-1335` |
| H4 | **The cross-process claim is under-verified.** Exactly one test spawns a real process (`:108`); every other "peer"/"startup-concurrent" scenario uses `threading.Thread` in one interpreter — and `_durable_settings_write_lock` takes the in-process RLock *before* the flock, so those tests pass with the flock code deleted entirely. The one real subprocess test sequences via stdin and never forces flock contention. Separately, no test SIGKILLs a process mid-write; "crash" tests monkeypatch a syscall to raise and then hand-rewrite the files to the state a crash would have left. The state-machine logic is well covered; the *locking primitive* and *real crash* are not. | qa | `test_mcp_credential_lifecycle.py:89-133` vs `:136-183` |
| H5 | **Same operator fault, two answers, three lines apart.** A missing projection directory yields a named, diagnosable 503 via `_rotate_private_projection_or_503` at `settings.py:2459-2460`, and an anonymous scrubbed 500 from `rotate_public_mcp_api_key()` at `:2497` — the `FileNotFoundError(ENOENT, "MCP credential directory is unavailable")` at `config.py:1402`. Operator whose bind mount failed to reattach after a host reboot gets a 500 with no indication a mount is the cause. | engineer | `settings.py:2497` vs `:2459-2460` |

## Improvements (could-fix)

- **Docs (M1).** The bead's first acceptance criterion is "A single **documented** authority and linearization point". Zero files under `docs/` changed. Stale now: `docs/architecture.md:570,576` ("Three distinct credentials live in `/config/settings.json`"), `docs/user_guide/integrations/mcp.md:390,631` ("The Bearer header uses `mcp_api_key` from `settings.json`") — an operator following that edits `settings.json` and the value is silently discarded (`config.py:1583`: "The caller's `mcp_api_key` is ignored"). The new operator-visible `.api-key.recovery` is missing from the "Where the MCP credentials live" table that `config.py:47` and `:1667` point operators at. No troubleshooting entry for the 503, and no runbook for what "repair storage" means. *(technical-writer, architect, code-reviewer, SRE, database — 5 personas)*
- **M2 — Private projection now rotates before the public transition** (`settings.py:2498` then `:2500`; base was the other order). The reorder is correct and deliberate — the fallible non-authoritative step fails first — but it is uncommented and `settings.py:2436`'s docstring still says "The public key in `settings.json` has already been written at this point, exactly as before", which is now false. No compensation if the public leg then raises `SettingsWriteTimeout`: the sidecar's private credentials are already replaced and every operator retry replaces them again. *(code-reviewer, qa)*
- **M3 — 503 now carries two opposite meanings.** `SettingsWriteTimeout` → 503 documented at `config.py:115-121` as "the save did not happen, nothing was written, retrying shortly is the right client behaviour". `mcp_api_key_durability_indeterminate` → 503 means the mutation *did* land and a blind retry rotates again. The structured `code` discriminator is good and the frontend branches on it correctly; the status code is what every other client sees first. *(code-reviewer)*
- **M4 — Lock domain ≠ resource domain.** The flock is taken at `<settings_file>.parent/.settings.lock` on the `ecm-config` volume (`config.py:172-176`); the resource is `MCP_KEY_FILE` on `ecm-mcp-secrets` (`config.py:69`, `docker-compose.mcp.yml:6,24`). And `save_settings(settings_file=X)` selects the lock namespace and write target but still reads/writes authority at the module-global path — latent (the one production caller passes `CONFIG_FILE`) but unguarded. *(architect, engineer, code-reviewer, security)*
- **M5 — Cache can pin permanently.** When `settings.json` is invalid *and* `api-key` is absent, `_cached_mcp_authority_signature` is `None`, so `config.py:1546-1547` returns the cache forever without re-checking disk. Reproduced: after a peer publishes a valid authority, `load_settings()` still returns `''`. In TLS mode the second process never converges without a restart. *(architect)*
- **M6 — Indeterminate durability is toast-only.** `MCPSettingsSection.tsx:128,152` call `notifications.warning(...)`, which gets the default 5000 ms and `aria-live="polite"` — *shorter and less interrupting* than an ordinary error, which has an explicit `duration: 8000`. Once it fades, the API Key card shows the same green "configured" badge as a fully durable rotation. The codebase already ships `.warning-message`/`.badge-warning` for exactly this degraded-but-not-failed state. *(ux)*
- **M7 — Test hygiene.** `test_settings.py:1888` asserts `len(...) > 20` with the comment "token_urlsafe(32) produces 43 chars" against a 40-char mocked literal; `:1903` asserts `!= "old-key-value"`, which is no longer seeded anywhere. `test_04c0u8_mcp_key_projection.py:213` `test_projection_failure_does_not_discard_valid_settings` now asserts `pytest.raises(PermissionError)` — behaviour inverted, name unchanged; `:71` passes only because `token_urlsafe` is patched to the same literal; `:183` cites a docstring on `_project_mcp_api_key`, a function this PR deletes. *(code-reviewer, engineer)*
- **M8 — Silent success-path mirror repair.** `_repair_settings_mirror_locked` (`config.py:1504-1533`) logs on write *failure* but emits nothing when it detects and silently overwrites a genuine mismatch — precisely the path that fires after a crash reverts a transition. "Why did my MCP client stop authenticating?" is unanswerable from logs. *(SRE, technical-writer)*
- **Nits** (~20, not itemized here): `_inspect_mcp_api_key_locked` called from the unlocked fast path (`:1549`); post-transition cached signature is the pre-rename temp signature and can never match (`:1238`); credential shape enforced at read time only, so a hand-edited multi-line value wedges the reader permanently; `MCPSettingsSection.tsx:131` logs the `HttpError` (whose `.detail` carries the key) to the console, narrowing the CHANGELOG's absolute claim to one of two layers; `httpClient.ts:36`'s new `'Request failed'` else-branch is app-wide and untested; `assert not transition_done.wait(timeout=0.1)` (`:170`) and `assert expected not in ... or not expected` (`:656`) are vacuous; the new test module drops the bead-id prefix its neighbours carry; `_stage_mcp_recovery_locked`'s in-place WAL rewrite is the module's only non-atomic write; the restore path inherits `SettingsWriteTimeout` as a new post-DB-swap abort cause.
- **`sbom/dev/*` is stamped `0.18.1-0153` while the build is `0158`.** Not a gate violation — `release-cut-gate.yml:239-247` verifies `sbom/vX.Y.Z/` and treats `sbom/dev/` as a transient snapshot — and the three *enforced* version touchpoints (`frontend/package.json`, `main.py:150`, `backup.py:126`) all agree at `0158`. But base was self-consistent at `0152` and head is not; the SBOM was generated at the branch's first commit and not refreshed after four further commits, all of which changed `config.py`. Re-running the generator is the whole fix. *(orchestrator + security)*

## Out-of-scope observations (recorded, not findings)

- `mcp-server/config.py:291,337` still does path-based `stat()`-then-`read_text()` (symlink-following, TOCTOU) where the backend now uses `O_NOFOLLOW` descriptors. Not touched by this PR.
- `backend/auth/settings.py:125` uses an **unbounded** `flock(LOCK_EX)` on the event loop; `config.py` now has a bounded acquisition for the same pattern. Two timeout semantics for one pattern in one codebase.
- `_merge_settings_preserving_redacted` (`backup.py:3589-3597`) snapshots `settings.json` outside the lifecycle lock, so a concurrent save of an unrelated field can be reverted by a legacy ZIP restore. Not a credential path.
- `backend/entrypoint.sh:342` chowns the MCP projection dir non-recursively while `:380` uses `chown -R` for `/config` — the asymmetry that makes B1's PUID scenario reachable. (Distinct from the excluded PUID=0 hardening finding; noted, not designed.)
- `config.py:107`'s comment ("stalls the whole loop, `/api/health` included") is inaccurate against `routers/health.py` — but it predates this PR verbatim.
- `docs/api.md` never documented `POST`/`DELETE /api/settings/mcp-api-key` at all. Pre-existing.

## Done well

Genuinely strong work, and the review found it by checking rather than assuming:

- **The single-authority property is real and was verified adversarially, not taken on faith.** Every `MCP_KEY_FILE` writer at head is `_replace_mcp_authority_locked` (`:1348`) or `_publish_mcp_api_key_locked` (`:1398`); all four callers take `_settings_write_lock` then `_durable_settings_write_lock()` in that order — no inversion, no nesting, no `await` or HTTP call inside a critical section, and no cross-module lock cycle (`backend/auth/settings.py` does not import `config`). The old unlocked republisher `_project_mcp_api_key` is gone. **392 sequential rotations against a separate OS process looping stale `save_settings` calls, artifact sampled 585,215 times from outside the lock: 0 stale-key observations, 0 index regressions.** That is the bead's central criterion, and outside the B2 abort path it is met.
- **The full durability chain is present on all five write paths** — temp → data fsync → atomic rename → **parent-directory fsync** — for `api-key`, the recovery replay, the WAL record, the record removal, and the `settings.json` mirror. Base had *no* directory fsync anywhere. The missing parent-dir fsync is the single most-missed step in hand-rolled durability code and it is not missed here.
- **File hardening is a strict superset of the sidecar's own checks:** `O_NOFOLLOW|O_CLOEXEC` everywhere, `O_EXCL` temp creation at `0600`, explicit `fchmod` after write (defeats umask), post-write `fstat` re-validation, a `(st_dev, st_ino)` cross-check against `os.stat(follow_symlinks=False)` closing the substitute-the-temp window (`:1232-1237`), single-descriptor reads with before/after signature comparison (`:1156-1158`), and `nlink == 1`. "Aligned with the sidecar" is accurate, and in the safe direction.
- **Revocation durability gets the subtle part right.** An existing empty file reads as `authority_exists=True, key=""`, so startup repair writes `""` rather than regenerating; and if the tombstone is deleted, `_initialize_authority_locked:1496` keys off `"mcp_api_key" in raw_settings` — **presence, not truthiness** — and re-publishes `""`. Easy to get wrong; not gotten wrong.
- **All three restore surfaces preserve destination authority, each by two independent mechanisms** — legacy ZIP (`backup.py:3602` + `:4367`), YAML (`:6189`), DBAS (`dbas/importers/ecm_settings.py:100`) — plus `save_settings` overriding from authority at commit time. Moving the settings install out of the staged-bytes swap to *after* `init_db()` (`backup.py:4360-4367`) is the correct fix for restore-vs-transition, not a patch on the symptom.
- **`config.py:1593-1604` — `save_settings` stripping the caller's `mcp_api_key` and re-reading authority under the lock** is the change that actually kills the stale-republication class. Single point, enforceable.
- **`config.py:1549` — signature-based cross-process cache invalidation.** Replacing scattered `clear_settings_cache()` calls with an `(st_dev, st_ino, st_size, st_mtime_ns, …)` recheck works across processes, which the calls it replaced did not.
- **No secret reaches a log**, verified by reading every `logger.*` call in `config.py:1100-1660`: none interpolates a key, a recovery document, or a raw settings body; `:1444` deliberately logs `type(error).__name__` so pydantic cannot echo field values. The frontend refactor *reduces* leakage — base `extractDetail` did `JSON.stringify(detail)` into the logged message, which would have put the live key in the browser console; head logs `httpError.message` only, and `httpClient.test.ts` pins it with a negative assertion on the captured `console.error` output.
- **The fault-injection tests assert invariants, not exceptions.** `test_recovery_activation_failure_does_not_hide_replaced_authority:603-685` asserts the authority content, the exact recovery document, the mirror, the cache, the log, then simulates the crash and re-asserts. `test_security_validation_finishes_before_authority_replace:766` pins an *ordering* property by instrumenting both calls rather than asserting an outcome that merely implies it. Described by one reviewer as the best fault-injection module in this repo.
- **Concurrency tests are barrier-based, not sleep-based** — zero `sleep` in 968 lines, all `threading.Event`/`join` with bounded timeouts, plus a real subprocess handshake. 50/50 on five consecutive randomized runs. Flake exposure is genuinely low, which is not the norm for a suite this size.
- **No regression test was loosened.** All 6 modified backend test files classified: five adapted-to-new-API with assertion strength preserved or increased, one tightened (`test_9kwzp2_4_legacy_restore_admission.py:431` went to an exact `expected_db_cycles = 2` rather than a permissive `>= 1`, with the reason named). The admin gate is untouched (`git diff -- backend/auth/` is empty) and its test now asserts *both* rotate and revoke were not called on refusal, where it previously asserted only `save_settings`.
- **`config.py:88-101`** — a comment that *retracts* a prior unevidenced claim ("that is not evidence-backed — no sync route calls `save_settings`") and replaces it with two reasons that hold. Independently re-verified: no sync route handler calls `save_settings`. Correcting an inherited comment instead of propagating it is what keeps a codebase honest.
- **The `MCPApiKeyDurabilityIndeterminate` design is the right call** — raising *after* the mirror and cache are aligned, and returning the active key in the 503 body so the operator is never locked out of a key that is live on disk. Most implementations either lie with a 200 or lose the key in a bare 500. The team explicitly does **not** want the key removed from that body: it is the same disclosure to the same admin-gated caller as the 200 path, and withholding it strands the operator with a live credential they can never learn.
- **The CHANGELOG entry is exceptionally accurate** for its density — reviewers spot-checked ten distinct claims against code at head and all held. Its one gap is that it does not disclose that `get_settings()` can now raise (B1).
- **Version discipline:** all three enforced touchpoints moved in lockstep `0152 → 0158`. Style compliance: 1 em-dash across 1997 added non-SBOM lines, and that one is a reflow of a pre-existing docstring.

## Process notes

- **CI verification independently confirmed against the job logs at head**, not the PR body: `11749 passed, 3 skipped, 2 deselected`; MCP `753 passed`; frontend `3588 passed`. The 3 skips are pre-existing and named in the log; the 2 deselected are the unchanged `-m "not slow"` filter. **No skip/xfail/deselect was added to dodge this change** — `pytest.mark.skip`/`slow` counts are byte-identical between `origin/dev` and head, and no pytest config file changed.
- **Scope integrity holds.** The excluded PUID=0 hardening finding stayed out (the only PUID hit in the diff is a *removed* comment). The `httpClient.ts` changes are not a rider — they are the minimum plumbing needed to carry the structured 503 detail the UI branches on.
- **The `71von` dependency is recorded in beads in both directions**, not merely in PR prose — `dependency_count`/`dependent_count` both 1. Worth crediting. Two gaps: neither bead explains *why* the coupling exists, and `71von`'s own most recent comment concluded its 503 was a Compose misconfiguration already fixed by #926. One comment naming the shared code path would close it.
- **The PR body says "scoped correctness and security review completed; no accepted Block-level findings remain."** The bead's own notes say the mandatory independent review was still pending. Both can be true — the four commits between `0153` and `0158` read as self-directed fix-ups — but the phrasing invites a reader to treat a self-review as the gate. This review is that gate, and it found four blockers.
- **Revert mechanics:** `docs/shipping.md` mandates `--merge`, so `git revert -m 1 <merge-sha>` reverts atomically. The five internal commits are not independently revertible — `0155`–`0158` are fixups on `0153`'s design. If this needs rolling back, revert the whole PR.

---

## What we are asking for

**Fix before merge:** B1, B2, B3, B4. B2 and B4 are correctness against the bead's own stated criteria; B1 and B3 are availability regressions against behaviour `dev` had explicitly documented and this PR removed. None requires a design change — B1's direction is a couple of lines in `_read_mcp_api_key_locked`, B3's is discard-and-continue in `_recover_mcp_authority_locked`, B4's is capturing the peer's key instead of re-reading the file under test.

**H1–H5 and M1–M8 are the PO's call** — fix in this round or file as follow-up beads. H1 is a one-line glob and H2 is a small conditional, so the cost of doing them now is close to the cost of filing them.

Per this repo's review discipline, an acceptance matrix will be frozen from this comment before the remediation round, and subsequent rounds will be delta-only against it — no re-litigating anything dispositioned here.

<sub>Reviewed by ten persona agents in parallel at small-team effective tier; findings deduplicated and the four blockers plus H1, H2 and the SBOM skew independently re-verified by the orchestrator against `723e61c1` before posting. Reproductions were built in disposable copies; the shared checkout was not modified.</sub>

PR #931, opened 2026-08-28
v0.18.1-0164: Reconcile MCP projection before readiness

MotWakorb:
## Summary
- reconcile the canonical MCP credential authority before ECM reports ready, provisioning or republishing the sidecar projection on fresh startup and recreation
- keep sidecar health fail-closed while preserving explicit revocation and valid non-object settings byte-for-byte
- extend the shipped Compose gate to prove startup, rotation, recreation, degraded projection behavior, and repair under default and custom runtime identities

## Verification
- backend: 11,832 passed, 3 skipped, 2 deselected
- MCP server: 753 passed
- frontend: typecheck, lint, 3,589 tests, and production build passed
- docs: strict MkDocs build passed
- SBOM: verify and audit --all passed
- container gate: passed for PUID/PGID 1000:1000 and 1571:1572
- git diff --check passed

Bead: enhancedchannelmanager-71von


MotWakorb:
# Team Review — PR #931 (v0.18.1-0164)

**Depth:** quick · **Effective tier:** small-team (strictest-wins; `ci-release` via version bumps + `sbom/dev/` regen) · **Participants:** security, architect, PM, engineer, UX, code-reviewer, database, SRE, QA, technical-writer · **Scope:** `git diff origin/dev...origin/fix/71von-health-projection` (merge-base `0830e873`) only.

## Verdict: approve with nits — no blockers

All 24 required checks are green. The runtime delta is small and correct: `settings_file_allows_startup_writes()` (`backend/config.py:1126-1134`), its two call sites, and one sidecar hint string. Everything else is tests, the Compose gate, docs, version bumps, and generated SBOMs.

**The strongest evidence in this review is a completeness enumeration, not an opinion.** Exactly three `settings.json` writers are reachable from `startup_event()`: `_repair_settings_mirror_locked` (`config.py:1653`, already early-returns on `"invalid"` — pre-existing, needs no guard), the league-delimiter heal (`main.py:1418` — now guarded), and `_trip_run_on_refresh_circuit_breaker` (`task_engine.py:335` — now guarded). All other `save_settings` call sites are request- or scheduler-driven. The guard is complete for its declared scope.

**The new test is load-bearing, verified by mutation rather than assumed.** Neutering each production guard independently (`main.py` guard → `if True:`; `task_engine.py` guard → `if False:`) makes all 5 parametrized cases fail in each case; unmutated, 19/19 pass. Two reviewers reproduced this separately.

## Findings

| # | Finding | Flagged by | Severity |
|---|---|---|---|
| 1 | CHANGELOG + PR title credit build 0164 with behavior that shipped in 0153 | security, architect, PM, SRE, tech-writer, code-reviewer (6) | Warn |
| 2 | New troubleshooting bullet sends revoked operators to a button that isn't rendered | tech-writer, UX | Warn |
| 3 | New `field_empty` doc bullet contradicts the unchanged `field_empty` runtime hint | UX | Warn |
| 4 | Guard's name/docstring/CHANGELOG read broader than the case it actually covers | security, QA | Nit (see note) |
| 5 | Two untested branches in the new function | QA, code-reviewer | Nit |
| 6 | Suppression logged in `task_engine.py`, silent in `main.py` | SRE, code-reviewer, DB | Nit |
| 7 | Recreate block never re-asserts ownership/mode on the republished projection | security | Nit |
| 8 | Invariant enforced by caller convention, not inside `save_settings()` | architect | Nit |

### 1 — Build misattribution (Warn)
`CHANGELOG.md:9` says deployments "**now** reconcile the owner-only API-key authority before readiness … (build 0164)". `git log -S "Reconcile public MCP credential authority before readiness" origin/dev -- backend/main.py` resolves to **`14c853e5`, v0.18.1-0153, "Linearize MCP credential lifecycle"** (bead `s6pg3`). That call is unchanged context in this diff. Only the entry's last sentence — byte-for-byte preservation of non-object `settings.json` — describes new runtime behavior. An operator on 0153–0163 reads this and upgrades for a fix they already have; a future bisect starts at the wrong build. The Compose-gate work is real, but it *verifies* 0153 behavior rather than introducing it.

### 2 — Revoked operators are sent to a non-existent button (Warn)
`docs/user_guide/integrations/mcp.md:631-633` (added here): "click **Regenerate Key** when you intend to re-enable MCP." After revocation `mcp_api_key` is `""`, so `mcp_api_key_configured=bool(settings.mcp_api_key)` (`backend/routers/settings.py:894`) is `False`, and `MCPSettingsSection.tsx` renders only **Generate API Key** in that branch — `Regenerate Key` exists solely in the `keyConfigured === true` branch. The bullet's exact target audience cannot follow it. (`mcp.md:308`'s "Generate Key" is likewise not a real label; the control is "Generate API Key".)

### 3 — Doc/runtime contradiction on `field_empty` (Warn)
The rewrite replaces the "No key generated yet" bullet with "**The key was explicitly revoked**". But the `field_empty` hint in `mcp-server/server.py:390-392` is untouched and still reads "No MCP API key configured. Generate one in ECM Settings > MCP Integration." The doc's own next step tells the operator to curl `/health` — so they get two different causal stories for one status from a single PR.

### 4 — Guard is narrower than it reads (Nit — deliberately not a blocker)
`settings_file_allows_startup_writes()` returns `True` on both `OSError` and `json.JSONDecodeError`, so an unreadable or truncated `settings.json` still gets replaced with defaults by the startup writers. Two reviewers reproduced the data loss with real unmocked code.

**We are not raising this as a blocker, and the reasoning matters:** that behavior is byte-for-byte identical on `origin/dev`, where the heal block is unconditional. This PR does not introduce or widen the data loss — it adds partial protection. What *is* anchored in new lines is an accuracy gap: the name `settings_file_allows_startup_writes`, the docstring "Return false only when valid JSON cannot represent ECM settings", and the CHANGELOG's "recovery evidence is not replaced by defaults" all read as general guarantees while delivering only the non-object case. Narrowing the prose (or widening the `except`) is a one-line change either way. Truncation is also the *more* likely corruption in practice than a hand-edited scalar — worth a follow-up bead, not a merge gate.

### 5 — Untested branches (Nit)
`pytest -k complete_startup --cov=config --cov-report=term-missing` reports `config.py` lines **1129, 1132-1133** unhit: the file-absent branch and both arms of the `except`. The five parametrized inputs (`null`, `[]`, `"scalar"`, `42`, `true`) all short-circuit at the same `isinstance(raw, dict)` check (`config.py:1573`) — five test IDs, one code path. Cheap insurance against a refactor, but the IDs advertise coverage that isn't there; a 3-line direct unit test would close it.

### 6 — Inconsistent suppression logging (Nit)
`task_engine.py:337-339` warns with the exact reason; `main.py:1418` skips silently — and skips the whole DB heal, not just its write. The asymmetry is *within one PR*; `task_engine`'s version is the better template.

### 7 — Recreate block doesn't re-check the republished projection (Nit)
The ownership/mode assertions moved to run only against the *first* startup projection. The recreate path is precisely where the root entrypoint re-runs — the `wrong_owner` mode this repo's own docs name — and the republished `api-key` / `mcp-service.json` pair is never re-checked for `%u:%g` or `600`. Separately, `${status:-unknown}` in the new failure message can report a stale `ok` carried over from the earlier readiness loop. One reviewer also flagged that `rm -f` runs against the still-serving container, so a settings read in the old process could republish before `--force-recreate` lands; that path is structural and was not reproduced (Compose couldn't be run read-only), and four other reviewers examined the gate without flagging it — worth a look, not a claim.

### 8 — Invariant lives at the callers (Nit)
`save_settings()` has no invalid-state check of its own, while `_repair_settings_mirror_locked` already encodes the identical rule internally. A future startup writer added by someone who doesn't know about the helper is unguarded by construction. Also two idioms for one invariant: `main.py` caches the flag (line 1170, consumed at 1418), `task_engine.py` re-reads (line 335).

## Credited

- **The Compose-gate reordering is the substantive win.** The old gate called `generate_key()` *before* waiting on sidecar readiness, so it could never have caught the fresh-deploy failure it exists to catch. Moving readiness ahead of the API call makes that assertion load-bearing for the first time — and `verify_authenticated_tools "$new_key"` after the recreate proves *republish without rotation*, which nothing previously asserted. `set -euo pipefail` means a failed `rm` aborts rather than passing vacuously.
- **The `other_uid` collision fix is a real latent bug.** Hardcoded `1234` silently voided the non-owner-read assertion whenever `PUID=1234` — a false negative that rots undetected.
- **Revocation stays fail-closed** through both restart and projection deletion, verified by reproduction. That's the property an auto-provision-at-startup change is most likely to break, and it holds.
- **Version touchpoints** are complete per `docs/versioning.md` (all three), and `test_version_touchpoint_consistency.py` passes 24/24.
- Non-UTF-8 `settings.json` raises `UnicodeDecodeError` uncaught, but `_read_settings_model_locked` on `dev` raises identically on the same input — **no new crash surface**, correctly ruled out of scope by two reviewers.

## Notes (not findings)

- Bead `71von` is **not** closed by this merge — it depends on `s6pg3` (P0, open), and its own latest comment says to hold it pending reporter confirmation. The PR correctly claims no closure.
- "SBOM: verify and audit --all passed" is the one verification line with no corresponding CI job on this PR (`generate_sbom.py verify/audit` runs only in `release-cut-gate.yml`). That matches the project's own design — `sbom/dev/` is explicitly not an artifact of record — but it reads alongside genuinely gated claims.
- Scheduler-driven writers (`tasks/m3u_refresh.py:260`, `tasks/channel_pipeline.py:319`) can still replace a preserved non-object `settings.json` minutes after boot. The `main.py:1168` comment says "startup-only", so this is disclosed rather than hidden.
- The run-on-refresh circuit breaker becomes un-trippable while `settings.json` is non-object — file-integrity now wins over crash-loop protection, logged but with no DB-side counterpart.

---
*Ten personas reviewed the merged diff independently, scope-fenced to this PR. Findings above were re-verified by the orchestrator against the head ref; one early orchestrator trace was wrong and is not included. Disposition of the eight findings is the PO's call.*

PR #932, opened 2026-08-28
v0.18.1-0165: Stop logging trusted-network values (dz5)

MotWakorb:
### What
Removes raw malformed trusted-network values from the attribution reconciler debug log while retaining a fixed-vocabulary skip diagnostic. Adds a regression test that proves credential-shaped input is absent from both the rendered log message and logger arguments. Bumps the required version touchpoints to `0.18.1-0165` and records the fix in the changelog.

### Why
Fixes GitHub CodeQL alert 1948 (`py/clear-text-logging-sensitive-data`, High): operator-configured `trusted_media_networks` entries flow through the persisted bandwidth and live stats attribution paths into `_parse_networks`, which previously logged malformed entries verbatim.

Alert: https://github.com/MotWakorb/enhancedchannelmanager/security/code-scanning/1948

### Verification
- Red-first regression: failed before the sink change because the credential-shaped input appeared in the DEBUG message; passes after the fix.
- Focused backend tests: 119 passed.
- Canonical backend gate: 11,833 passed, 3 skipped, 2 deselected; 79.80% coverage.
- Frontend: lint and typecheck passed; 3,589 tests passed; production build passed.
- Python syntax compile passed.
- `scripts/generate_sbom.py verify` passed; no dependency manifest changed, so regeneration was not required.

### Security
- Raw settings values, credentials, identifiers, and aggregates are not logged by the changed diagnostic.
- No dependency or infrastructure changes.

### Deployment
No migration or operator action required. Rollback is a revert of commit `29f60c34e6b149ff497066c6fe12a3da31a13fbf`.

Do not merge from this PR creation session; parent owns review and merge authorization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
+- **Malformed trusted-media-network settings no longer echo their raw values into attribution debug logs (bead `enhancedchannelmanager-dz5`, build 0165).** The skip diagnostic remains available using fixed text only.
+
 - **Supported MCP Compose deployments now reconcile the owner-only API-key authority before readiness instead of leaving the sidecar permanently not ready (bead `enhancedchannelmanager-71von`, build 0164).** Fresh installs and upgrades use the canonical credential lifecycle to provision or republish `api-key` without requiring an operator API call, while explicit revocation remains revoked and `/health` remains fail-closed. The container gate now proves both fresh startup and recreation from persisted settings. Startup also leaves syntactically valid non-object `settings.json` content byte-for-byte intact: authority validation and orphan reconciliation still run, but startup-only settings writers are suppressed so recovery evidence is not replaced by defaults.
 
 ### Added
--- a/backend/main.py
+++ b/backend/main.py
@@ -150,7 +150,7 @@ handle authentication automatically when accessed through the web UI.
 Login endpoints are rate-limited to 5 requests per minute per IP address.
     """,
 
-    version="0.18.1-0164",
+    version="0.18.1-0165",
     openapi_tags=tags_metadata,
     docs_url="/api/docs",
     redoc_url="/api/redoc",
--- a/backend/routers/backup.py
+++ b/backend/routers/backup.py
@@ -124,7 +124,7 @@ LEGACY_RESTORE_DIRS = ["uploads/logos", "tls", "m3u_uploads"]
 # scripts/check_version_consistency.py that used to fail the PR on divergence
 # were removed. Do NOT rename it, change its shape, or repurpose it. It is an INFORMATIONAL human-readable string ("which
 # ECM build produced this artifact") — it is NOT a compatibility gate.
-APP_VERSION = "0.18.1-0164"
+APP_VERSION = "0.18.1-0165"
 
 # DBAS backup-artifact schema version (ADR-008 D1 / ADR-012 D1). This is a
 # DEDICATED, MONOTONIC INTEGER that is DISTINCT from the human-readable
--- a/backend/services/attribution_reconciler.py
+++ b/backend/services/attribution_reconciler.py
@@ -506,10 +506,7 @@ def _parse_networks(cidrs: list[str]) -> list[ipaddress._BaseNetwork]:
         try:
             networks.append(ipaddress.ip_network(token, strict=False))
         except ValueError:
-            logger.debug(
-                "[ATTR-RECONCILE] Skipping unparsable trusted-network entry %r",
-                token,
-            )
+            logger.debug("[ATTR-RECONCILE] Skipping unparsable trusted-network entry")
     return networks
 
 
--- a/backend/tests/services/test_attribution_reconciler.py
+++ b/backend/tests/services/test_attribution_reconciler.py
@@ -23,6 +23,8 @@ Synthetic identities only.
 """
 from __future__ import annotations
 
+import logging
+
 import pytest
 
 from services.attribution_reconciler import (
@@ -89,6 +91,23 @@ def _assigned_names(result) -> dict[str, str | None]:
         a.client_id: (a.user.user_name if a.user else None)
         for a in result.assignments
     }
+
+
+def test_malformed_trusted_network_log_does_not_retain_input(caplog):
+    sensitive_value = "username:provider-password@example.invalid"
+
+    with caplog.at_level(logging.DEBUG, logger="services.attribution_reconciler"):
+        assert build_trusted_networks(configured_cidrs=[sensitive_value]) == []
+
+    records = [
+        record
+        for record in caplog.records
+        if record.name == "services.attribution_reconciler"
+    ]
+    assert [record.getMessage() for record in records] == [
+        "[ATTR-RECONCILE] Skipping unparsable trusted-network entry"
+    ]
+    assert all(sensitive_value not in repr(record.args) for record in records)
 
 
 # ---------------------------------------------------------------------------
--- a/backend/tests/services/test_attribution_reconciler.py
+++ b/backend/tests/services/test_attribution_reconciler.py
@@ -27,6 +27,7 @@ import logging
 
 import pytest
 
+from observability import JsonFormatter
 from services.attribution_reconciler import (
     AMBIGUOUS_GROUP_PREDICATE,
     CandidateUser,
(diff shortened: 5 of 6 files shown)


MotWakorb:
## Team Review — PR #932 (ten-persona, quick depth)

**Verdict: approve. No Block-level findings.** Three non-blocking follow-ups below, none of which need to be fixed in this PR.

Scope was fenced to the changed lines only, per the review request. Effective tier: **home-lab** (`ecm-core`, `dev-tooling` per `COMPONENTS.md`), with the three version touchpoints held to small-team rigor under the cross-tier strictest-wins rule. Branch merge-base verified equal to `origin/dev` — not stale.

### Verified, not taken on trust

- **The sink closes the taint path.** Alert 1948 pins its sink to `attribution_reconciler.py` start_line 511, cols 17–22 — precisely the `token` argument this PR deletes. Both taint paths in the alert record (sources `backend/bandwidth_tracker.py:3696` and `backend/routers/stats.py:559`) terminate there. Within `_parse_networks`, `token` now has exactly one remaining use — `ipaddress.ip_network(token, strict=False)` — which is not a log sink. The only other `logger` call in the file (`:927`) receives channel labels and user names; `configured_cidrs` has no path to it. `build_trusted_networks` logs nothing.
- **The red-first claim is real**, reproduced independently five times. Restoring the pre-fix `%r` form in a disposable worktree fails the test with the credential-shaped value in the captured log: `At index 0 diff: "[ATTR-RECONCILE] Skipping unparsable trusted-network entry 'username:provider-password@example.invalid'" != '[ATTR-RECONCILE] Skipping unparsable trusted-network entry'`. Green at PR head.
- **The test cannot pass vacuously.** Exact-list equality on `getMessage()` means a capture failure (propagation off, wrong logger name, sink never reached) yields `[] != [...]` and fails.
- **Version bump is complete** across the three canonical touchpoints (`backend/main.py:153`, `backend/routers/backup.py:126`, `frontend/package.json:5`), and is enforced — `backend/tests/unit/test_version_touchpoint_consistency.py` discovers touchpoints by tree scan, so a missed one fails CI rather than only review.
- **All 25 checks pass**, including the four contexts branch protection actually requires on `dev`.

We did **not** independently re-run the 11,833-test canonical gate, the frontend gate, or CodeQL. Those claims stand on CI, not on this review.

### Findings

| # | Finding | Rating | Flagged by |
|---|---------|--------|-----------|
| 1 | Diagnostic value removed entirely — an operator cannot tell *which* entry was rejected | Warn | Security, Code Reviewer, UX, SRE |
| 2 | Bead `enhancedchannelmanager-dz5` does not exist | Warn | PM (verified independently) |
| 3 | PR body's stated rollback commit is incomplete | Nit | Orchestrator |
| 4 | Test placement breaks the file's section-banner/class organization | Nit | Code Reviewer |
| 5 | The fix is one `exc_info=True` away from a full re-leak, and nothing says so | Nit (optional) | Security |

**1 — Diagnostic loss.** `attribution_reconciler.py:509` now emits N identical lines for N malformed entries. Because the parsed list is a union of `server_ips`, `configured_cidrs`, and `detected_gateways`, an operator cannot even tell whether the bad entry came from their own setting or from auto-detected gateways. Non-blocking at home-lab: the list is hand-entered (realistically 0–3 entries), visible in the UI, DEBUG is not the default level, and the function's own docstring frames the failure as benign — "the worst case is that the entry contributes nothing to ranking."

There is a genuine disagreement worth recording. The Code Reviewer anchored this on the project's own precedent at `docs/style_guide.md:474-477`, where `safe_regex` logs a SHA-256 plus a 50-char excerpt rather than the raw pattern — "redact but stay diagnosable." Security pushed back specifically: **do not port that form here.** A 50-char excerpt of a pasted credential *is* the leak, and length or character-class fingerprints are computed *from* the tainted value, so they are candidate taint-propagation steps that can re-open the alert. Security holds domain authority on taint, so the resolution is: if the diagnostic is restored, the **list index** is the only unambiguously safe carrier — `"...entry at index %d", idx` — an `int` from loop position carrying zero bytes of the tainted string. The `safe_regex` precedent is real but does not transfer, because a regex pattern and an operator-pasted settings value have different leak profiles.

**2 — Dangling bead reference.** `bd show enhancedchannelmanager-dz5` returns "no issue found matching", and `dz5` appears nowhere in `.beads/issues.jsonl` (4,190 beads present). This is anchored in a changed line — the CHANGELOG entry ships the reference — and in the PR body header. Rated Warn rather than Block: the code is correct and the traceability defect is repairable without touching the diff.

**3 — Rollback instruction is incomplete.** The PR body says "Rollback is a revert of commit `29f60c34`", but the PR carries two commits — `29f60c34` and `02f3399c` ("Harden trusted-network log regression"). Confirmed by attempting it: `git revert --no-commit 29f60c34` at PR head produces `CONFLICT (content): Merge conflict in backend/tests/services/test_attribution_reconciler.py`. The stated rollback does not apply cleanly.

**5 — Latent re-leak.** The `except ValueError:` is load-bearing: that exception object contains the raw token verbatim. The next person wanting better diagnostics will reach for `exc_info=True`, which looks additive and restores the full value. The test does catch it, which is why finding 6 below matters — but a one-line comment at the `except` would make the constraint legible where the edit happens rather than only at test-failure time.

### Conflict resolved: the `JsonFormatter` assertion is not redundant

The Code Reviewer assessed `assert all(sensitive_value not in JsonFormatter().format(record) ...)` as "mostly redundant" with the `getMessage()` equality. Security and QA both demonstrated concrete regressions that pass the first two assertions and are caught **only** by that third one:

- `exc_info=True` added to the fixed-string call — the value reaches the payload via `observability.py:182`.
- `extra={"entry": token}` instead of a positional arg — swept into the payload via `observability.py:187-194`; `repr(record.args)` is empty and catches nothing.

Two independent empirical demonstrations beat one impression. **Keep the assertion.** If anything is subsumed it is `repr(record.args)`, which is also the cheapest of the three and worth leaving.

### Positive observations

- Correct, minimal fix: the `%r` specifier and its argument were removed together — no dangling format specifier, no eager f-string introduced.
- The test enters through the public `build_trusted_networks`, the same entry point both CodeQL-named callers use, rather than poking the private helper.
- Testing at the real production formatter (`JsonFormatter`) rather than only the raw `LogRecord` API is the right instinct, and it demonstrably earns its keep.
- CHANGELOG entry is accurate, correctly placed under `Unreleased/Fixed`, and matches the surrounding bead-reference convention.
- `_parse_networks`'s docstring never claimed the value was logged, so the change introduces no doc staleness.
- No findings from the Database Engineer, IT Architect, UX, SRE, or Technical Writer — no schema, boundary, user-facing, operational, or documentation surface is touched. Grep for `[ATTR-RECONCILE]` and `trusted_media_networks` across `docs/runbooks/` and `docs/user_guide/` returned nothing, so no runbook depends on the removed value.

### One caveat on closure

Alert 1948's instance `ref` is `refs/heads/dev`, so it will not transition to fixed until a CodeQL scan runs on `dev` after merge. "Fixes alert 1948" is a forward claim until that scan completes — worth confirming post-merge rather than treating the merge itself as closure.

### Out-of-scope observations (notes only — no severity, no remediation designed)

- `trusted_media_networks` has no write-time validation, so malformed entries are accepted into settings and only rejected at read time. Operator feedback for a config typo arguably belongs at that validation layer rather than in a DEBUG log.
- The same `getattr(settings, "trusted_media_networks", None)` source pattern appears at `backend/bandwidth_tracker.py:3696` and `backend/routers/stats.py:559`; whether other alerts share those sources is unexamined.

MotWakorb:
### Finding dispositions

PO call: follow-up beads only — nothing is fixed or amended in this PR, and nothing blocks the merge.

| Finding | Disposition | Bead |
|---------|-------------|------|
| 1 — Diagnostic loss (no way to identify which entry was rejected) | Follow-up | `enhancedchannelmanager-fejzk` (P2) |
| 2 — CHANGELOG cites nonexistent bead `dz5` | Follow-up | `enhancedchannelmanager-63yri` (P2) |
| 3 — PR body's rollback cites only 1 of 2 commits | Folded into 2 | `enhancedchannelmanager-63yri` |
| 4 — Test placement | Follow-up | `enhancedchannelmanager-3gzsl` (P3) |
| 5 — `exc_info=True` re-leak hazard undocumented | Follow-up | `enhancedchannelmanager-3gzsl` (P3) |

Finding 3 is recorded inside bead `63yri` rather than getting its own, because a PR body is ephemeral — the durable fact worth keeping is that **a real rollback of this change must revert both `29f60c34` and `02f3399c`**, not the single commit the body names. Reverting `29f60c34` alone conflicts in `test_attribution_reconciler.py`.

Both follow-up beads carry the constraint that the regression test must not be weakened while addressing them — the `JsonFormatter` assertion catches two demonstrated regressions the other two assertions miss.

One item is not bead-tracked because it is a post-merge check, not work: alert 1948's instance `ref` is `refs/heads/dev`, so it transitions to fixed only after a CodeQL scan runs on `dev`. Confirm that scan rather than treating the merge as closure.

Issue #1009, opened 2026-09-14
[Bug]: /run/commit returns 502 on the first Dispatcharr 429; replay is fail-fast and the body cannot tell the caller whether retry is safe


lukeeexd:
### ECM Version

v0.18.1

### Priority

Priority 3: Medium, needs looking at soon but not immediate

### What is the bug?

`POST /api/channel-pipeline/run/commit` (phase `execute`) returns **HTTP 502** with body `{"message": "pipeline replay partially failed", "completed_writes": [], "compensation_errors": []}`. The same pipeline, triggered from the UI's Run button, applies cleanly every time.

Reading the source, the 502 is raised by ECM itself, not by a proxy:

- `backend/routers/channel_pipeline.py:2032` converts a `PartialReplayError` into `HTTPException(502, ...)`.
- `PartialReplayError` is raised from `replay_write_plan` in `backend/services/pipeline_write_plan.py:247`. The replay loop is a plain sequential `for` over the plan's writes wrapped in a single bare `except Exception`. The first write that raises aborts the whole replay, best-effort compensation runs, and the error surfaces as 502.
- The `/run` path the UI uses is different code with different error policy. `engine.run_pipeline` wraps each write in its own `try/except`, records the failure in `execution_log`, and continues (collect-and-continue). Replay is fail-fast. Same client, same sequential writes, no pacing in either path; the difference in outcome is the error policy.
- `backend/dispatcharr_client.py` has no 429 handling anywhere. Every write method calls `response.raise_for_status()`, so a single `429 Too Many Requests` from Dispatcharr becomes an `HTTPStatusError` and aborts the replay. The JWT `_login` and `_refresh_access_token` paths are exposed the same way: a 401 on a write triggers `_refresh_access_token`, a 429 there falls through to `_login`, and a 429 there raises via `raise_for_status`. The only 429 handling in the backend is in the Discord and Telegram alert modules.

In the observed runs Dispatcharr's JWT endpoints returned 429 roughly six times during bulk write sequences in the same session, so this is the most likely trigger.

[...]

PR #1008, opened 2026-09-14
fix(event-sync): say why a non-M3U-backed master group fails pre-flight (#1007)

lukeeexd:
# Pull Request

## Summary
- Refs #1007. The `group_settings_found` pre-flight failure for a **master** group read as a lookup error ("was not found in the M3U account's group settings — the group may have been removed or renamed"). The check is really enforcing the ownership constraint in `docs/event_sync.md`: Dispatcharr owns master-channel lifecycle from an auto-synced M3U group, so a hand-curated Dispatcharr channel group is not a supported master. The message now says that first and keeps the removed/renamed case as the trailing hint.
- Same note added to `docs/event_sync.md` ("Pick the master group" and "Pre-flight checks") and to the pre-flight module docstring.
- **No behaviour change.** Check id, failure shape (`group_id` / `role` / `check` / `expected` / `got` / `message`) and the secondary-role message are untouched. The wider relaxation discussed in #1007 is deliberately not in this PR; that is the maintainer's call.
- No bead: external contribution, linked to GitHub #1007 instead.

## Test plan
- `pytest tests/unit/test_event_sync_preflight.py` — 35 passed. Nothing pins the message text; the tests assert on `check` / `role` / `group_id`, all unchanged.
- `tests/unit/test_event_sync_rollback_roundtrip.py` (holds the AST no-group-writes gate): the gate passes; 6 tests in that file fail identically on a clean `dev` checkout on Windows (`os.O_NOFOLLOW`, `resource.getpagesize`), unrelated to this change. CI on Linux should be green.
- Manual: preview against a curated master now reports the ownership constraint instead of a missing group.

## Checklist
- [ ] Bead linked in the commit message (`[bd-xxxxx]`) — none; external contribution, refs #1007.
- [x] Tests updated or added (backend `pytest`, frontend `vitest`) — existing tests cover the unchanged failure shape; wording-only change, no new test.
- [x] `CHANGELOG.md` updated under `[Unreleased]` → Fixed. Build bumped to 0.18.2-0036 (`backend/main.py`, `backend/routers/backup.py`, `frontend/package.json`). Note: PR #1006 also bumps to 0036; whichever merges second needs a one-line rebump.
- [x] Docs updated (`docs/event_sync.md`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1006, opened 2026-09-14
feat(channel-pipeline): let Create Channel leave tvg_id empty (tvg_id_mode)

lukeeexd:
# Pull Request

## Summary
Closes #1005.

- Adds a per-action `tvg_id_mode` to `create_channel`: `inherit` (default, byte-identical to today) or `none`, which omits `tvg_id` from the create payload in both live and dry-run and skips the `if_exists: update` back-fill so the field stays empty across re-runs. The execution log records `tvg_id left empty (tvg_id_mode=none)` when a stream tvg_id was suppressed.
- Schema validation accepts `inherit`/`none` (absent or null falls back to `inherit`, so every stored rule is unaffected); `GET /schema/actions` and the MCP rule-creation tool document the parameter.
- Rule editor gains a **TVG-ID for new channels** selector (Inherit from stream / Leave empty) under Create Channel; re-selecting Inherit drops the key from the action so saved rules stay minimal.
- User guide and `CHANGELOG.md [Unreleased] / Added` updated; build bumped to `0.18.2-0036` across `frontend/package.json`, `backend/main.py`, `backend/routers/backup.py`.

Scope is deliberately the one behaviour from the issue. `generate` is not added because `assign_epg` with `set_tvg_id: true` and `assign_tvg_id` already cover it. Nothing actively clears a tvg_id on channels that already have one.

## Test plan
- [x] `backend/tests/unit/test_channel_pipeline_executor.py::TestCreateChannelTvgIdMode` (8 new): default and explicit `inherit` copy the stream tvg_id; `none` omits it from the create payload and the dry-run simulated channel; `if_exists: update` skips the back-fill under `none` and still back-fills under `inherit`; no noisy detail line when the stream has no tvg_id.
- [x] `backend/tests/unit/test_channel_pipeline_schema.py` (3 new): accepts `inherit`/`none`/absent/null, rejects unknown values, dataclass `to_dict` round-trip.
- [x] `frontend/src/components/channelPipeline/ActionEditor.test.tsx` (4 new): defaults to Inherit and offers both modes, sets `tvg_id_mode: 'none'`, drops the key when Inherit is re-selected, not rendered for `create_group`. Whole file: 77 passed.
- [x] `tsc --noEmit` clean.
[...]

PR #1010, opened 2026-09-14
v0.18.2-0036: fix #1009 planned commit partial-failure reporting and Dispatcharr 429 backoff

lukeeexd:
# Pull Request

## Summary
Fixes #1009. Scoped to the three items agreed there; the commit endpoint stays synchronous and the 30s request-timeout exposure is deliberately left for your decision.

- **Dispatcharr client retries 429.** `_request` and `_login` retry `429 Too Many Requests` with bounded exponential backoff (1s, 2s, 4s; cap 10s), honouring `Retry-After` when present. A spent budget surfaces as an `httpx.HTTPStatusError` carrying the 429 response. No subclass is introduced because `test_no_swept_class_inherits_from_outside_the_module` forbids classes in `dispatcharr_client.py` inheriting from outside the module. Previously no code path handled 429 anywhere and every write called `raise_for_status`, so one 429 aborted the fail-fast replay.
- **`PartialReplayError` says what happened.** It now carries `failed_write`, `not_applied` and `pre_mutation` alongside the existing `completed` / `compensation_errors`. `pre_mutation` is `True` only when no write completed, compensation was clean, and the first write's exception proves upstream rejected it before mutating (a 4xx response or a refused connection). Timeouts, dropped connections, 5xx and unknown exceptions are treated as "may have landed", so a caller is never told a retry is safe when it is not.
- **Commit returns 424, not 502.** `POST /api/channel-pipeline/run/commit` answers a partial replay with `424 Failed Dependency` and a body containing `execution_id`, `failed_index`, `failed_write`, `pre_mutation`, `completed_writes`, `not_applied` and `compensation_errors`. The same fields are recorded in the execution row's `partial_replay_failure` log entry and the evidence snapshot. 424 was chosen because it is not a status a proxy emits, so the caller can attribute it to ECM.

Also: `docs/user_guide/integrations/mcp.md` describes the new response; `CHANGELOG.md` entry under `[Unreleased]`; build bumped to `0.18.2-0036` in all three touchpoints. Note #1006 and #1008 (also open, also from me) bump to the same build, so whichever merges later will need a rebump.

## Test plan
- New `tests/services/test_pipeline_write_plan.py` cases: first write 429 → `pre_mutation=True`, `not_applied` lists both writes, second write never attempted; failure after a landed write → `pre_mutation=False` with correct `completed` / `not_applied`; first write `ReadTimeout` → `pre_mutation=False`.
- New `tests/unit/test_dispatcharr_client_rate_limit.py`: `_request` retries with 1s/2s backoff then succeeds; honours `Retry-After: 7`; raises a 429 `HTTPStatusError` after `RATE_LIMIT_MAX_RETRIES + 1` attempts; `_login` retries then stores tokens; `_login` spent budget raises and leaves no token.
- New `tests/routers/test_channel_pipeline.py::TestCommitPartialFailureGH1009`: drives `commit_auto_creation_pipeline` with a stored plan and a replay that raises `PartialReplayError`, asserts the 424 status, every body field, and that `_mark_execution_failed` received `failed_write` / `pre_mutation`. The handler is invoked directly (as `test_event_sync_cleanup` does) rather than through `async_client`, because the HTTP path depends on POSIX-only settings-file checks.
- `test_event_sync_cleanup.py::test_prepared_commit_persists_confirmed_cleanup_and_rollback[True]` updated from 502 to 424.
- Ran locally on Windows with `--no-cov`: the three new test files plus `test_dispatcharr_client_auth`, `_bodyless_success`, `_stream_crud`, `_version_advisory`, `_contract_sweep`, `test_channel_pipeline_plans`, `test_04c0u7_planned_execution_parity`, `test_mcp_planned_route_enforcement`, `test_mutation_plan_store`, `test_version_touchpoint_consistency`: 168 passed, 0 failed.
- Not run here: the full `scripts/backend-gate.sh` (Linux-only backend; `fcntl`, `resource`, `os.O_NOFOLLOW`). `test_event_sync_cleanup.py` fails wholesale on Windows before and after this change for the same reason, so the 502→424 assertion update there is unverified locally and relies on CI.

## Checklist
- [ ] Bead linked in the commit message (`[bd-xxxxx]`). Not done: external contributor, no `bd` access. Commit references `#1009` instead.
- [x] Tests updated or added (backend `pytest`, frontend `vitest`). Backend only; no frontend change beyond the version bump.
- [x] Docs updated (`docs/user_guide/integrations/mcp.md`).


MotWakorb (changes requested):
## Full team review — all five findings requested for correction

Reviewed `79ac5695eca10a630ea27e8e3468583a723a2102` against `086e3e4fcec7eb0d334c4b37f08f20e82507d447`, covering all 12 changed files with ten team perspectives, calibrated to the affected components' home-lab tier.

The retry behavior, HTTP 424 distinction, execution ID, and persisted failure evidence are useful improvements for #1009. However, the recovery inventory can misstate upstream state, identify temporary rather than actual resources, and expose credential-bearing URLs. The retry-delay parser also needs correction. **Please fix all five items below, including the lower-priority HTTP-date issue, before the follow-up review.**

### 1. Medium / P2 — Distinguish unknown outcomes from writes not applied

**Location:** [`backend/services/pipeline_write_plan.py:350–355`](https://github.com/MotWakorb/enhancedchannelmanager/blob/79ac5695eca10a630ea27e8e3468583a723a2102/backend/services/pipeline_write_plan.py#L350-L355), propagated through `backend/routers/channel_pipeline.py:2026–2051` and described in `docs/user_guide/integrations/mcp.md:743–751`.

The inclusive slice starts at the failed operation, so that operation is always listed as `not_applied`. An exception does not establish that its mutation failed to reach Dispatcharr.

**Independently reproduced:** A disposable upstream HTTP server applied PATCH to channel 7, then disconnected before returning its response. A subsequent delete was never attempted. The actual client, replay, FastAPI handler/middleware, and SQLite persistence produced:

```json
{
  "completed_writes": [],
  "failed_write": "update_channel:7",
  "not_applied": ["update_channel:7", "delete_channel:8"],
  "pre_mutation": false,
  "compensation_errors": []
}
```

Channel 7 had changed. The execution log and snapshot persisted the same incorrect inventory. An operator or MCP consumer can therefore repeat an already-applied operation or misidentify what needs reconciliation.

**Requested correction:** Distinguish confirmed rejection, failed operation with unknown outcome, and subsequent operations never attempted. Only classify the failed write as unapplied when its non-application is established. Preserve this distinction across the exception, HTTP response, execution log, snapshot, and documentation.

Also clarify that `completed_writes` is successful forward-call history: compensation may subsequently have undone an operation. Empty `compensation_errors` does not establish complete restoration.

**Required evidence:** A landed-write/lost-response regression checking both the HTTP result and persisted failure evidence, alongside a confirmed-rejection control. `pre_mutation=false` was correct in the reproduction; this finding is not a claim that the flag incorrectly authorized retry.

### 2. Medium / P2 — Report resolved upstream target IDs

**Location:** [`backend/services/pipeline_write_plan.py:351–354`](https://github.com/MotWakorb/enhancedchannelmanager/blob/79ac5695eca10a630ea27e8e3468583a723a2102/backend/services/pipeline_write_plan.py#L351-L354).

Replay invokes client methods with mapped arguments, but constructs `failed_write` and the remaining-write descriptions from original `write.args`.

**Independently reproduced:** The real planner records create-channel and an update to temporary ID `-1`. Replay creates upstream channel **101**, sends four rejected PATCH requests to **101**, and compensates against **101**. The failure nevertheless reports `failed_write="update_channel:-1"` and `not_applied=["update_channel:-1"]`.

**Requested correction:** Preserve the resolved target of the attempted write and use known mappings in subsequent diagnostic descriptions. Keep genuinely unresolved future targets explicitly distinguishable. Diagnostic rendering must not fail merely because a future resource has not been created.

**Required evidence:** A recorded create → dependent-update failure verifies the actual upstream ID in failure output and persisted evidence, including a compensation-failure case where locating the remaining resource matters.

### 3. Medium / P2 — Keep credential-bearing creation payloads out of failure diagnostics

**Locations:** [`backend/services/pipeline_write_plan.py:113–114`](https://github.com/MotWakorb/enhancedchannelmanager/blob/79ac5695eca10a630ea27e8e3468583a723a2102/backend/services/pipeline_write_plan.py#L113-L114), `351–355`; `backend/routers/channel_pipeline.py:2028–2050`.

`_write_target()` stringifies the entire first argument. For `create_logo`, that is a dictionary including its URL, which may contain provider credentials or access tokens.

**Independently reproduced with a synthetic token:** A first-write 429 rejection copied the token-bearing logo URL into `failed_write`, `not_applied`, the execution log, and the snapshot's partial-failure data. Passing the real error response through the existing MCP HTTP client also placed the token in its logged error and surfaced exception. This happens even with `completed_writes=[]`.

Raw completed-write formatting and full plan storage predate the PR; the new exposure of failed/unattempted payloads through these error fields is the finding here.

**Requested correction:** Use bounded, method-appropriate operation identifiers rather than raw argument dictionaries or URLs: resolved resource IDs where known, or a plan-operation index for a not-yet-created resource. Address the producer in the changed reporting path so downstream consumers receive safe diagnostics.

**Required evidence:** A synthetic credential-bearing creation payload remains absent from the generated diagnostic fields, persisted partial-failure fields, and resulting MCP error/log presentation. Preserve useful operation correlation.

### 4. Medium / P2 — Enforce finite bounds on header-directed retry waiting

**Location:** [`backend/dispatcharr_client.py:283–291`](https://github.com/MotWakorb/enhancedchannelmanager/blob/79ac5695eca10a630ea27e8e3468583a723a2102/backend/dispatcharr_client.py#L283-L291), consumed by login at `358–363` and API requests at `496–501`.

The numeric `Retry-After` branch returns before the fallback cap and accepts non-finite floating-point values:

| Header | Selected delay |
| --- | --- |
| `86400` | 24 hours |
| `inf` or `1e309` | Infinity |
| `nan` | Zero |

**Independently reproduced:** An infinite login sleep held `_auth_lock`, leaving a second request pending until the proof cancelled both. The pinned base terminated with 429 and released the lock on the same response. Retry count alone does not bound this waiting time.

**Requested correction:** Validate finite, nonnegative delays and enforce an explicit finite retry-wait budget. If a valid server-directed delay exceeds the supported budget, surface throttling rather than wait indefinitely or retry before the server permits it. Apply consistent behavior to login and ordinary requests; align the bounded-backoff documentation with the actual policy.

**Required evidence:** Normal numeric delays, oversized delays, non-finite/invalid values, exhaustion, and authentication-lock release are covered without real long sleeps.

### 5. Low / P3 — Handle the HTTP-date representation of `Retry-After`

**Location:** `backend/dispatcharr_client.py:289–291`; the changelog describes unqualified `Retry-After` support.

**Independently reproduced:** A real HTTP 429 with a valid GMT date approximately 20 seconds ahead selected a one-second delay. The parser treats the valid date as garbage and falls back to exponential backoff, potentially spending the retry budget before recovery is permitted.

**Requested correction:** Parse HTTP-date values as well as delay-seconds, calculate the remaining nonnegative interval, and apply the same finite-budget policy from item 4. An over-budget date should produce the bounded terminal behavior, not an early retry. Keep fallback behavior for genuinely invalid input.

**Required evidence:** Future and past dates, numeric headers, invalid headers, and dates beyond the supported budget. This remains lower severity because no evidence establishes that this deployment's Dispatcharr emits dates, but it is included in the requested fixes.

| Item | Expected outcome |
| --- | --- |
| 1 | Unknown, confirmed-rejected, and unattempted outcomes remain distinct in HTTP and persisted evidence; compensation documentation accurately describes forward-call history versus current state. |
| 2 | Failure evidence identifies the actual contacted resource when its ID is known, including after a created resource's dependent write fails; unresolved future references remain explicit. |
| 3 | Safe operation descriptors preserve correlation without copying credential-bearing creation payloads into diagnostic fields or MCP error/log output. |
| 4 | Both retry paths use finite, validated waiting policies and terminate cleanly when a server-directed delay exceeds the budget. |
| 5 | Both valid `Retry-After` representations are interpreted correctly under that same budget, with regression coverage for edge cases. |

Follow-up review should verify this matrix and inspect the remediation delta for regressions. The separately documented synchronous-commit timeout/cancellation concern from #1009 is outside these requested changes; these findings do not require an asynchronous endpoint migration.

- Independently rerun targeted Linux suites: **521 passed**, covering retry/authentication, client contracts, replay, routers, cleanup, planning parity, MCP route enforcement, plan storage, and version consistency.
- Additional HTTP/replay/persistence fault tests: **2 passed, 3 failed**, confirming items 1, 2, and 5.
- Separate independently rerun security probes reproduced items 3 and 4, including the unchanged MCP consumer and actual authentication-lock behavior.
- Both cleanup persistence variants passed, including the 424 assertion previously unverified on Windows. All **74 dependency pins** matched; diff whitespace check passed.

These are targeted local checks using isolated databases, controlled HTTP servers/transports, and synthetic credentials. They are not live Dispatcharr acceptance or the full CI gate. GitHub currently reports the PR blocked with no published check results; required CI remains a prerequisite for merge.

**Landing-order note:** Build `0036` is internally consistent here but overlaps other open contributions. Reconcile the three application-version touchpoints and changelog build reference with the actual landing order if needed.

lukeeexd:
All five items addressed in `9a5ad262`.

**1. Unknown vs not applied.** `PartialReplayError` now carries `failed_outcome`: `"rejected"` (4xx response or refused connection: upstream provably did not mutate) or `"unknown"` (timeout, dropped connection, 5xx, anything else: it may have landed). `not_applied` lists only the writes *after* the failed one, which were never attempted; the failed write is classified by `failed_outcome` and is no longer in that list. Your landed-PATCH-lost-response case now reports `failed_write="update_channel:7"`, `failed_outcome="unknown"`, `not_applied=["delete_channel:8"]`, `pre_mutation=false`. `completed_writes` is documented (docstring, `mcp.md`) as forward-call history, and an empty `compensation_errors` as "compensation raised nothing". Exposed on the 424 body, the `partial_replay_failure` log entry and the snapshot evidence.

**2. Resolved ids.** Descriptors use the live remap: an update recorded against temp id `-1` that PATCHed channel 101 reports `update_channel:101`; a completed create names what it produced (`create_channel#0->101`) so a failed compensation can be located; a still-unresolved future target renders as `pending(-2)` instead of failing rendering.

**3. Credential-bearing payloads.** `describe_write_target` never stringifies argument dictionaries or URLs. A `create_logo` payload is identified by plan position (`create_logo#0`) in `failed_write`, `not_applied`, `completed` and `str(exc)`; a canary token in the URL is asserted absent from every one of those, and from `repr(exc)`.

**4. Finite, validated waits.** `_parse_retry_after` accepts only finite, non-negative delay-seconds; `inf`, `1e309`, `nan`, negatives and garbage fall back to exponential backoff. Every call's 429 waits are bounded by an explicit `RATE_LIMIT_MAX_TOTAL_WAIT = 30s` budget shared by login and API requests; a server-directed delay that would exceed it raises the 429 `HTTPStatusError` immediately ("Retry-After exceeds the 30s wait budget") instead of sleeping. Asserted: `Retry-After: 86400` → no sleep, one request, 429 raised; cumulative `20 + 20` stops after one wait; `_ensure_authenticated` on that path leaves `_auth_lock` unlocked and no token stored.

**5. HTTP-date.** `Retry-After` in HTTP-date form is parsed with `email.utils.parsedate_to_datetime` to its remaining non-negative interval under the same budget (a past date means retry now; a date beyond the budget is terminal, not an early retry). Clock is injected via `_now` for the tests.

**Tests.** `test_pipeline_write_plan.py` +8, `test_dispatcharr_client_rate_limit.py` +14 (parser table, over-budget seconds and dates, invalid values, cumulative budget, lock release, login HTTP-date), router test updated for `failed_outcome` and the narrowed `not_applied`. Local (Windows, `--no-cov`, stubs): write-plan + router + parity: 23 passed; rate-limit + contract sweep + auth: 84 passed; the wider client/plans/MCP/cleanup suites have a FAILED id set byte-identical to the previous head (1 Windows-only failure on both sides). Docs and changelog aligned.

**Merge note.** #1014's follow-up also edits the `except` block of `replay_write_plan` (soft-skip bookkeeping and the true `failed_index`); whichever lands second will need a small rebase there. Build stays `0036`.

MotWakorb (changes requested):
## Strict delta review: acceptance evidence remains incomplete

Reviewed remediation head `9a5ad262` against the frozen follow-up matrix. The implementation and unit coverage address the five requested behaviors, but the retained tests do not prove items 1 through 3 across the required HTTP, persistence, and MCP boundaries.

### Requested correction

1. Add a retained landed-write/lost-response regression that drives a controlled upstream HTTP connection through the real client and replay path, then the commit handler and SQLite persistence. Assert `failed_outcome=unknown`, the failed write is absent from `not_applied`, later writes are unattempted, and both the 424 payload and persisted execution/snapshot evidence preserve that distinction. Keep a confirmed-rejection control through the same boundary.
2. Extend the created-resource/dependent-write case through the commit persistence path so the 424 and stored recovery evidence identify the resolved upstream ID, including the compensation-failure case.
3. Carry a synthetic credential-bearing create payload through the HTTP failure, persisted partial-failure evidence, and MCP consumer/log presentation. Assert the canary is absent while safe operation correlation remains.

The current replay tests stop at `PartialReplayError`; the router test injects a preconstructed rejected error and mocks `_mark_execution_failed`, so it cannot detect a regression in replay-to-HTTP-to-persistence propagation. No new product behavior is requested: these are the original required regressions for matrix items 1 through 3. Retry budget and HTTP-date coverage satisfy items 4 and 5.

lukeeexd:
Items 1–3 now have retained cross-boundary regressions in `1ebdb37c` (rebased onto `origin/dev`; build is `0039` because `dev` landed at `0036`).

**What the new test drives.** `tests/integration/test_gh1009_commit_partial_failure_evidence.py`: the real `DispatcharrClient` over a controlled `httpx.MockTransport` that plays Dispatcharr (it can apply a PATCH and then drop the response, or answer 4xx/5xx), the real `replay_write_plan` with its compensation pass, the real `POST /api/auto-creation/run/commit` handler over ASGI (`async_client`), and the real `_mark_execution_failed` into `ChannelPipelineExecution` and `ChannelPipelineSnapshot`. Only `_compute_pipeline_plan_payload` is doubled (returns the stored plan, i.e. a non-drifted commit).

**1. Unknown vs rejected vs unattempted.** `test_landed_write_with_lost_response_is_unknown_in_http_and_persistence`: the transport applies `PATCH /channels/7/` (asserted: upstream channel 7 now reads `New`) and raises `RemoteProtocolError`. The 424 body, the `partial_replay_failure` log entry and the snapshot's `partial_replay` all carry `failed_write="update_channel:7"`, `failed_outcome="unknown"`, `not_applied=["delete_channel:8"]`, `pre_mutation=false`, `completed_writes=[]`; the delete was never sent (asserted on the request record and on upstream state). Control `test_confirmed_rejection_control_…`: the same write answered 429 → `rejected`, `pre_mutation=true`, upstream unchanged, same three surfaces. RED check: forcing `_failure_precludes_mutation` to return True fails only the lost-response test, on `'rejected' == 'unknown'`.

**2. Resolved ids through persistence.** `test_dependent_write_failure_names_the_real_created_id_…`: `create_channel` → upstream 101, `PATCH /channels/101/` → 400. The 424 and both persisted surfaces carry `completed_writes=["create_channel#0->101"]`, `failed_write="update_channel:101"`, `not_applied=["update_channel:101"]`; `-1` appears nowhere; the compensation `DELETE /channels/101/` is observed and 101 is gone upstream. `test_compensation_failure_leaves_the_created_id_locatable_…`: the DELETE answers 503; the orphan is still upstream, and the 424 plus persisted evidence carry `create_channel#0->101` and the `create_channel: …` compensation error.

**3. Credential canary.** `test_credential_bearing_logo_payload_is_absent_…`: `create_logo` URL `?token=<canary>` and `create_channel.tvg_id=<canary>`, first write 429-rejected. The canary, `provider.example` and `logo.png` are absent from the 424 body, the log entry, the snapshot evidence, the execution `error_message` and the backend DEBUG `caplog` for the whole commit; `execution_id`, `create_logo#0`, `create_channel#1`, `update_channel:pending(-2)` and `failed_outcome` survive. The stored plan blob is the one place the payload legitimately lives (it is the replay program) and predates this PR; the assertions are on the diagnostic fields the failure added.

**MCP consumer.** `mcp-server/tests/test_gh1009_partial_failure_presentation.py` runs the sidecar's real `ECMClient.post` and the `run_channel_pipeline` tool over a mock transport answering the backend's 424 shape: the operator text and the `[ECM-CLIENT]` log carry `execution_id`, `failed_write`, `failed_outcome`, `not_applied`. A control shows `_http_error` renders `detail` verbatim (a body that *did* carry a URL would surface it), which is why the payload-free guarantee is proven at the producer above rather than in the sidecar.

**Verification (Windows, `--no-cov`, local POSIX shims).** New integration file, write-plan, rate-limit, client auth/contract/bodyless/stream, pipeline router, plans router, event-sync cleanup, planning parity, MCP route enforcement, mutation plan store, version consistency: 513 passed, 0 failed. Sidecar file in its own venv: 3 passed. `scripts/backend-gate.sh` not run here; CI is the check.

**Merge note.** #1014's branch also edits the `except` block of `replay_write_plan`; whichever lands second needs a small rebase there.

MotWakorb (changes requested):
## Strict delta review: matrix closed, two current-head blockers remain

Reviewed `9a5ad262..1ebdb37c` against the frozen five-item matrix with all ten team perspectives. The new cross-boundary regressions satisfy items 1-3; previously accepted items 4-5 remain closed. No accepted behavior is being reopened.

### Requested correction

1. Prevent the new tests from leaking process-global sensitive-value registrations. `backend/tests/integration/test_gh1009_commit_partial_failure_evidence.py:92-96` constructs a real client with `api_key="k"`; `backend/tests/unit/test_dispatcharr_client_rate_limit.py:21-27` similarly registers one-character credentials. `DispatcharrClient.__init__` adds these values to the global redactor. The backend gate then rewrites `sequence-*` as `seq***REDACTED***ence-*`, changes record sizes, evicts the oldest rotation entries, and fails `test_rotates_at_small_bound_and_snapshot_is_oldest_first`. This is direct current-head CI evidence, not the unrelated capacity-only failure seen on #1014. Keep these tests isolated from process-lifetime redaction state and rerun the backend gate.
2. Resolve the open Critical CodeQL alert #2034 at `backend/dispatcharr_client.py:541-547`. The 401 retry request is flagged for partial SSRF from route-derived path values. The team did not establish exploitability because the Dispatcharr base URL remains configured and fixed, but the required security gate fails closed. Correct the sink or provide the repository-required evidence-backed false-positive disposition.

The matrix behavior itself is accepted: unknown/rejected/unattempted outcomes, resolved upstream IDs, credential-safe diagnostics, bounded retry waits, and HTTP-date parsing all remain closed.

lukeeexd:
Both current-head blockers addressed in `af6f6904`.

**1. Redactor leakage.** Root cause confirmed locally: `DispatcharrClient.__init__` registers its credentials with the process-global redactor, so a one-character key from a test rewrote `sequence-1` as `seq***REDACTED***ence-1` in a later test. Two-part fix: the new tests now use long, unique synthetic credentials (`test-only-gh1009-api-key-…`, `test-only-rate-limit-…` for the api-key, username and password fixtures), and `tests/conftest.py` gains an autouse fixture that snapshots the registered sensitive-value forms before each test and restores them afterwards, so no test can leak a registration into another regardless of ordering. Probed locally with a two-test file (first constructs a client with `api_key="k"`, second asserts `k` is no longer registered and `sequence-1` survives `_redact_persistent_string`): both pass. The rotation test itself cannot run on Windows (it needs a POSIX directory descriptor), so the backend gate on CI is the check for that assertion; the isolation fixture removes the mechanism it failed on. The same fixture text is going onto #1014 so the two branches merge cleanly.

**2. CodeQL alert 2034.** I cannot dismiss alerts on this repository, so I corrected the sink rather than dispositioning it. The alert was on the 401 re-issue inside the new 429 loop, which duplicated the outbound `self._client.request` call. `_request` now has exactly one request call site: a 401 in JWT mode refreshes the token, sets a per-attempt flag and loops back to that single call (at most one refresh per rate-limit attempt, the historical contract); a 429 backs off, increments the attempt counter, and loops back to the same call. Behaviour is unchanged for api-key mode (401 terminal, no refresh), `retry_on_401=False`, and the retry budget; the existing auth, rate-limit, version-advisory and settings-router tests pin all of these and pass. The only remaining sink is the pre-existing one, whose alert exists on `dev` today.

**Verification (Windows, `--no-cov`, shims).** New integration file, rate-limit, client auth, version advisory, contract sweep, bodyless, stream CRUD, settings router, write plan, pipeline router, log_utils, in that order (the changed files ahead of the rotation test as in your repro): 618 passed; the only failures are the 20 `TestInterProcessRotatingJsonHandler` cases and one settings-router case that fail identically on the untouched branch (Windows-only, the api-key file-mode check).

Issue #1011, opened 2026-09-14
[Bug]: assign_epg against a dummy EPG source hard-fails for channels the same run just created — the Pass 5 defer-and-retry only engages when the source is completely empty


lukeeexd:
### ECM Version

0.18.2-0035 (dev, `086e3e4f`)

### Priority

Priority 3: Medium, needs looking at soon but not immediate

### What is the bug?

A standard pipeline rule of the form `create_channel` → `assign_epg` (target = an ECM dummy EPG source, `/api/dummy-epg/xmltv/<profile_id>`) fails every `assign_epg` for channels created in that same run. The dummy source still describes the channels as they were *before* the run, `assign_epg` is matched against that stale set, and the run finalises `completed_with_errors` with a bare count of failed actions. A second run, after a manual regenerate + refresh, succeeds.

This is the same symptom as the long-closed #43. The fix for that case (defer into Pass 5) only covers a dummy source with **no entries at all**; a source that already has entries for other channels is not covered.

**Where the gap is:**

- `backend/channel_pipeline_executor.py:2649` — `_execute_assign_epg` defers a dummy-source assignment into `_deferred_epg_assignments` only when `source_entries` is **empty**.
- `backend/channel_pipeline_executor.py:2683` — with entries present but no match for the channel, it hard-fails (`No EPG data match found`) unless `defer_on_no_match` is true.
- `defer_on_no_match` is passed at exactly one call site: `assign_event_sync_dummy_epg`, `backend/channel_pipeline_executor.py:5781` (the Event Sync dummy-EPG step, ti939.3.3). The docstring at line 2624 says the default "keeps the standard-rule path byte-identical (no-match = failure)".
- A stored rule has no equivalent: `backend/channel_pipeline_schema.py:651` validates only `epg_id` and `set_tvg_id` for `assign_epg`, and the action editor exposes only those two.

**Pass 5 already does everything needed.** `_refresh_dummy_epg_and_retry` (`backend/channel_pipeline_engine.py:5461`, invoked at `:3722`) adds the target groups to the profile, regenerates the XMLTV cache, refreshes each Dispatcharr source and polls for completion, re-fetches EPG data, and retries each deferred assignment. It just never runs here, because nothing put the standard-rule assignments onto `_deferred_epg_assignments`.

Two things that look like fixes but are not:

[...]

PR #1012, opened 2026-09-14
fix #1011: assign_epg defers into Pass 5 for same-run-created channels when a dummy source has no entry yet

lukeeexd:
# Pull Request

## Summary
- Fixes #1011. A standard `create_channel` → `assign_epg` rule targeting an ECM dummy EPG source failed every `assign_epg` for the channels it had just created: `_execute_assign_epg` only deferred into Pass 5 when the dummy source had **no** entries; with entries present for other channels and none for the new one it hard-failed unless `defer_on_no_match` was set, and that flag is passed only by the Event Sync dummy-EPG step. The XMLTV endpoint is a 300s cache that channel creation never invalidates, so a mid-run refresh alone cannot help and a scheduled Dummy EPG Refresh races the run.
- `channel_pipeline_executor.py`: on no-match against a recognised dummy source, defer when the channel is in `_created_channels` (created by this run), so it rides the **existing** Pass 5 regenerate → refresh + poll → re-fetch → retry unchanged. Pre-existing channels keep no-match = failure, but the error now reads `Dummy EPG source N has no entry for '<channel>' yet — regenerate the dummy EPG and refresh the source, or add the channel's group to the profile` instead of the generic `No EPG data match found`. Non-dummy sources are byte-identical.
- `channel_pipeline_engine.py`: `_summarize_failed_actions` groups by (rule, action type, **error**) so the run-level summary carries each sampled action's reason. Still bounded to five distinct labels. "32 action(s) failed during this run ('Snooker (MAX UK)' assign_epg)" could not tell "the source has no entry for this channel yet" from a genuine mismatch.
- Build bumped to 0.18.2-0036 (same bump as #1006/#1008/#1010; whichever merges second needs a rebump). No bead: external contribution against a GitHub issue.

**Opened as a draft** because the `defer_on_no_match` default and its "byte-identical" docstring read like a deliberate scope call from ti939.3.3. The widening here is limited to same-run-created channels for the reason given in #1011: deferring on every no-match would mask genuine mismatches on existing channels and spend a full Pass 5 cycle on them. Happy to mark ready, narrow, or close depending on your answer on the issue.

**Sequencing note:** Pass 5's round trips (paginated channel fetch, stream lookup, per-source refresh, polling up to 120s) are unprotected against Dispatcharr 429 until the backoff from PR #1010 is on `dev`.

## Test plan
- [x] RED first: `TestDeferredEPGAssignment` gained three cases that failed against `dev` for the intended reasons, then passed after the change:
  - `test_assign_epg_dummy_source_no_match_for_created_channel_defers` — dummy source has stale entries, channel is in `_created_channels` → `deferred=True`, lands on `_deferred_epg_assignments`, no `update_channel` call.
  - `test_assign_epg_dummy_source_no_match_for_existing_channel_fails_naming_source` — same source, channel not created this run → fails, not deferred, error names the source and channel with "no entry".
  - `test_assign_epg_non_dummy_source_no_match_for_created_channel_fails` — non-dummy source, created channel → unchanged `No EPG data match found`. (Uses two entries so the single-entry fallback does not apply.)
- [x] `TestSummarizeFailedActionsCarriesReason` (engine): the summary contains the per-action error; identical reasons collapse to one label and the sample stays bounded (`+N more`).
- [x] Existing `TestDeferredEPGAssignment` empty-source and reload-then-retry cases still pass.
- [x] Full `test_channel_pipeline_executor.py` + `test_channel_pipeline_engine.py`: the failure set is byte-identical to an untouched `origin/dev` worktree (43 Windows-only `PermissionError: api-key does not have the required owner and mode` failures in the engine file, present before and after). Run with `--no-cov` on Windows with `fcntl`/`resource` stubs; a Linux CI run is the authoritative check.
- [ ] Live check against the #1011 setup (create 16 channels + assign from the stale source in one run → 0 failed actions, `ecm-25xx` ids after Pass 5) — not yet done from this branch; the machine that reproduced it runs the published image.

## Checklist
- [ ] Bead linked in the commit message (`[bd-xxxxx]`) — none; external contribution, commit references `Closes #1011`.
- [x] Tests updated or added (backend `pytest`, frontend `vitest`) — backend only; no frontend change.
- [x] `CHANGELOG.md` updated under `[Unreleased]` → Fixed.
- [x] Docs updated if behaviour or interfaces changed — `docs/user_guide/channel-pipeline/conditions-and-actions.md` Assign EPG row notes the deferral.


MotWakorb (changes requested):
## Full team review — all seven findings requested for correction

Reviewed `83615bc6e93b2e91dd15bdc3b8f926adbfc74ede` against `086e3e4fcec7eb0d334c4b37f08f20e82507d447`, covering all nine changed files with ten team perspectives, calibrated to the affected components' home-lab tier.

The intended direct-run improvement for #1011 is real: an integrated 16-channel/32-stream test succeeds after regeneration and assignment. However, the widened deferral also exposes execution-mode, identity, profile-scope, and reporting defects. **Please address all seven findings below, including the lower-priority terminal-retry diagnostic, and align the accompanying documentation.**

### 1. High / P1 — Do not persist successful planned execution when EPG work was omitted

**Changed anchor:** [`backend/channel_pipeline_executor.py:2693–2716`](https://github.com/MotWakorb/enhancedchannelmanager/blob/83615bc6e93b2e91dd15bdc3b8f926adbfc74ede/backend/channel_pipeline_executor.py#L2693-L2716).

The new branch also returns successful deferral during planning. The engine invokes Pass 5 in simulation mode for `plan_only` (`backend/channel_pipeline_engine.py:3736–3738`), so its refresh and retry descriptions do not become executable writes. Commit replays the recorded writes and persists that planning result.

**Independently reproduced through the real prepare/commit handlers and SQLite:** the channel is created, no source refresh occurs, `epg_data_id` remains absent, the provider TVG-ID remains, and the durable execution is **`completed` with no error**. Its log contains successful “Would refresh”/“Would retry” entries. Replacing the assignment method with the pinned base implementation restores `completed_with_errors` for the same input.

This is specifically a regression in the durable execution result; the immediate commit response's pre-existing status behavior is not the finding.

**Requested correction:** Do not classify preview-only deferred work as fulfilled. Preserve an explicit unsuccessful outcome when planning cannot represent and execute the assignment; report completion only when the actual commit fulfills it.

**Required regression:** Prepare → commit a create/assign rule against a populated stale dummy source. Assert actual channel EPG fields, source-refresh activity where applicable, recorded work, and persisted execution status. A green execution with an unfulfilled assignment must fail the test.

### 2. Medium / P2 — Bind a deferred assignment to its original channel

**Changed anchor:** `backend/channel_pipeline_executor.py:2697–2708`.  
**Retry consumer:** `backend/channel_pipeline_engine.py:5784–5823`.

The queued tuple stores a channel ID alongside a mutable `ExecutionContext`. Later actions can change `current_channel_id`. Pass 5 then calls `_execute_assign_epg` with that mutated context rather than binding execution to the queued target.

**Independently reproduced:** create channel **5000**, defer its EPG assignment, then create/select **5001** in the same action sequence. The actual EPG PATCH targets **5001**, leaves 5000 unassigned, and records no failed action.

**Requested correction:** Preserve a stable, channel-specific retry target and the assignment's original parameters. A later action must not redirect the deferred write.

**Required regression:** A rule changes channel context after enqueueing an assignment. Verify the original channel receives the intended EPG and TVG-ID behavior, the later channel does not receive that assignment, and the log identifies the actual target.

### 3. High / P1 — Preserve source/profile-to-group isolation during Pass 5

**Changed admission point:** `backend/channel_pipeline_executor.py:2697`.  
**Affected processing:** `backend/channel_pipeline_engine.py:5484–5562`.

Pass 5 collects a global union of deferred target groups and applies that union to every participating dummy profile. The new branch routes populated-source standard rules into that existing logic.

**Independently verified through SQLite:** with separate groups and EPG sources, profile 1 changed from `[1644]` to `[1644,1645]` and profile 2 from `[1645]` to `[1644,1645]`. No failed actions were recorded. These are persistent changes to feed coverage. In the same-name reproduction, the mixed feeds also allowed a channel to receive the other channel's guide identity.

**Requested correction:** Retain the association between each deferred assignment's source/profile and its target group. Update each profile only with groups targeting that profile. Preserve intentional combined-feed behavior as a distinct case.

**Required regression:** Two populated-but-stale dummy sources backed by separate profiles retain separate persisted group membership after one run. Verify each channel's resulting guide association, including equally named channels in different groups.

The union implementation predates this PR; its newly reachable effect on these standard-rule no-matches is the regression under review.

### 4. Medium / P2 — Use complete ID-based same-run creation provenance

**Changed anchor:** `backend/channel_pipeline_executor.py:2693–2695`.

`_created_channels` is a lowercase-name lookup cache. Creating an equally named channel in another group replaces its previous entry, so scanning its values is not a complete record of channels created during the run.

**Independently reproduced:** channels **5000** and **5001** were created in different groups under the same name, but the cache retained only 5001. Revisiting 5000 made the new predicate treat it as pre-existing and record a hard assignment failure. Its earlier queued assignment subsequently succeeded, but the failure remained in the run results.

**Requested correction:** Make same-run creation eligibility depend on lossless channel-ID provenance, independently of the name lookup cache.

**Required regression:** Create same-named channels in different groups, then process another stream/action selecting the earlier channel. Both channels must retain same-run eligibility, with correct assignments and no spurious retained failure.

This is incomplete coverage of the new same-run exception, rather than a claim that a previously successful matching case became unsuccessful.

### 5. Medium / P2 — Preserve originating rule identity through deferred failure reporting

**Changed queue admission:** `backend/channel_pipeline_executor.py:2697–2708`.  
**Failure consumer:** `backend/channel_pipeline_engine.py:5837–5840`.

Pass 5 records terminal assignment failures using `rule_id=None`. Selected-rule outcomes count errors by rule ID, so a rule whose deferred assignment failed can still be marked successful.

**Independently reproduced and persisted:** the overall run is `completed_with_errors` with two failed assignments, while its selected rule is **`completed`, `error_count=0`**.

**Requested correction:** Carry the originating rule identity through deferred work and use it when recording terminal failures and computing selected-rule outcomes.

**Required regression:** A selected/scheduled rule whose assignment remains unresolved after refresh has the appropriate failed/degraded per-rule outcome and error count in the execution row, consistent with the overall run.

### 6. Low / P3 — Make the final retry terminal instead of promising another refresh

**Changed anchor:** `backend/channel_pipeline_executor.py:2697–2716`; summary propagation at `backend/channel_pipeline_engine.py:2627–2629`.

The new same-run predicate remains true during the final Pass 5 retry. If refreshed data still has no match, the action requeues itself and returns `success=True`, `deferred=True`, `error=None`.

**Independently reproduced:** the final action log says success and the run's failure reason says “Deferred: will assign EPG … after refresh,” although no later pass will drain the new queue.

**Requested correction:** Make an exhausted retry non-deferring and return an explicit failure naming the channel/source and unsuccessful refresh attempt. Keep action-level logs and terminal summaries consistent with that outcome.

**Required regression:** A still-unmatched final retry leaves no promise of future work, records a failed action with an actionable reason, and is represented consistently in persisted logs and status. The existing run-level `completed_with_errors` behavior must remain intact.

This is not a second false-green whole-run finding: direct execution already marks the overall run erroneous in this case.

### 7. Medium / P2 — Bound and scrub operator-facing failure summaries

**Changed anchor:** [`backend/channel_pipeline_engine.py:2627–2635`](https://github.com/MotWakorb/enhancedchannelmanager/blob/83615bc6e93b2e91dd15bdc3b8f926adbfc74ede/backend/channel_pipeline_engine.py#L2627-L2635).

Appending complete raw errors promotes unrestricted upstream diagnostics into ordinary execution summaries. Five sampled labels bound the count, not their size or sensitive contents.

**Independently reproduced:**

- A synthetic credential in an upstream error passed through the real client/executor into the persisted summary and execution-history response without requesting detailed logs. The base summary did not contain the credential.
- Five approximately 128-KiB error bodies produced a **655,975-byte summary**, versus **270 bytes** using the base summarizer.
- All distinct full labels are retained and checked through list membership before only five are displayed, adding avoidable aggregation work for channel-specific errors.

**Requested correction:** Generate redacted, bounded operator-facing reasons at this summary boundary. Limit individual reasons and the complete output, indicate truncation, retain useful rule/action/source identification, and use efficient deduplication. Preserve the useful dummy-source explanation without copying raw upstream response dumps into the summary.

**Required regression:** Normal actionable reasons remain visible; synthetic credential-bearing errors are scrubbed; oversized and many-distinct-error inputs produce bounded summaries with accurate counts.

Security classified the conditional disclosure risk as Low; P2 here is the consolidated delivery priority for summary safety and robustness. The proof uses synthetic credentials and does not establish a live compromise, authorization bypass, or XSS.

### Documentation corrections accompanying these fixes

- Update `_execute_assign_epg`'s retained docstring at `backend/channel_pipeline_executor.py:2624–2629`: the claim that the default keeps standard-rule no-match behavior “byte-identical” now contradicts the automatic same-run exception.
- Align `docs/user_guide/channel-pipeline/conditions-and-actions.md:82` and `CHANGELOG.md:11` with the corrected execution-mode and terminal-retry behavior; do not promise fulfilled work where planning only simulated it.
- Qualify the changelog's ECM-stamped TVG-ID claim: successful retry assigns guide data, and changes the channel's TVG-ID when `set_tvg_id` is enabled. The `False` case must continue preserving the existing TVG-ID.

| Item | Expected outcome |
| --- | --- |
| 1 | Planned commits cannot persist successful completion while omitting the requested assignment; actual effects and durable status agree. |
| 2 | Deferred execution targets the original channel despite later context changes. |
| 3 | Independent sources/profiles retain their intended group coverage and resulting guide associations. |
| 4 | Same-run provenance remains correct across duplicate channel names in different groups. |
| 5 | Failed deferred work is attributed to its originating rule in persisted per-rule outcomes. |
| 6 | Exhausted retries produce terminal failed action records, not successful re-deferral promises. |
| 7 | Failure summaries remain actionable, redacted, bounded, and efficiently deduplicated. |
| Documentation | Comments, guide, and changelog accurately describe execution modes, automatic deferral, terminal failure, and `set_tvg_id`. |

Follow-up review should verify this matrix and inspect the remediation delta for regressions. These requests concern integration of the newly admitted assignments and the changed summary boundary; they do not call for a general pipeline redesign.

- Independently rerun existing Linux subsets: **743 passed**, covering executor/engine, dummy EPG, Event Sync, planning, notification/status, and version tests.
- Integrated lifecycle probes: **5 passed, 2 failed**, confirming intended direct-run behavior plus the planned false-success and terminal-retry defects.
- The positive tests include **16 channels / 32 streams**, actual XMLTV generation and SQLite finalization, both `set_tvg_id` settings, and mutation-free dry-run behavior.
- Independent selected-rule, channel-target, profile-membership, creation-provenance, and summary probes reproduced the additional findings above.

Verification used isolated databases and controlled Dispatcharr test doubles. It is not live Dispatcharr acceptance or the full CI gate. GitHub currently reports the PR blocked with no published check results; required CI still needs to complete on the proposed merge revision.

**Sequencing:** The direct-run proof succeeds without PR #1010's rate-limit changes, so its merge is not established as a functional prerequisite. Reconcile build `0036` across the three version touchpoints and changelog with the actual landing order if another contribution lands first.

lukeeexd:
All seven items and the documentation corrections addressed in `b42d0972`.

**1. Planned runs (P1).** The same-run deferral applies to direct runs only. Under `plan_only` a no-match for a same-run-created channel is an explicit failure whose reason says a planned run cannot regenerate and refresh the dummy EPG before commit (and how to proceed), so the durable execution ends `completed_with_errors` rather than `completed` with simulated "Would refresh/retry" entries. Asserted with `ActionExecutor(plan_only=True)`. The pre-existing empty-source deferral under planning is unchanged (out of this PR's admission).

**2. Bound to the original channel.** A deferral queues `_pin_context(exec_ctx)`: a shallow copy pinned to the channel being assigned (accumulator lists shared, so nothing is double-counted). Moving `current_channel_id` afterwards no longer redirects the Pass 5 PATCH; asserted on the queued tuple.

**3. Profile isolation (P1).** Pass 5 collects groups per source, maps each profile-feed source to its profile, and adds to each profile only the groups whose deferred assignments target that profile's source. The union is kept solely for a source that is not one profile's feed (the combined feed), which cannot be attributed. Asserted on the real `_refresh_dummy_epg_and_retry`: profile 1 gains `[1644]`, profile 2 `[1645]`; a profile already covering its group is untouched; the combined-feed case still receives the union.

**4. Id-based provenance.** New `_created_channel_ids` set, populated by both create paths (live and dry-run); eligibility reads it instead of the lowercase-name cache. Asserted with two same-named channels in different groups where the cache retains only the later one.

**5. Rule identity.** `ExecutionContext` carries `rule_id` / `rule_name` (set by the engine per stream); the pinned copy carries them into Pass 5, which records a terminal failure against that rule (falling back to the legacy `[Pass 5 EPG retry]` label only when the context has none). Asserted through `_selected_rule_outcomes`: the rule reports `completed_with_errors`, `error_count == 1`.

**6. Terminal retry.** The Pass 5 retry passes `allow_defer=False`; an exhausted retry returns an explicit failed action ("… even after the dummy EPG was regenerated and the source refreshed — …") and never re-queues. Covered for the no-match, the still-empty-source and the event_sync `defer_on_no_match` paths.

**7. Bounded, redacted summaries.** `_summarize_failed_actions` redacts each reason (registered credential values via the `obfuscate` module, `key=value` secrets, `Bearer` tokens, URL userinfo / XC path shape / credential query parameters), truncates reasons to 160 chars and the summary to 1200 with a marker, and deduplicates with a set. Asserted: a canary in `?token=`, `Bearer`, `password=` and URL userinfo/path is absent while `page=2` and `upstream 401` survive; a registered secret is scrubbed without any marker; 5 × 128 KiB + 500 distinct reasons produce a summary under 2000 chars with `+500 more`; the ordinary dummy-source reason stays readable.

**Docs.** The `_execute_assign_epg` docstring no longer claims byte-identical standard-rule behaviour; the user guide row and the changelog describe direct-run deferral, the terminal retry, planned-run failure, and that a successful retry changes the TVG-ID only when `set_tvg_id` is enabled.

**Verification (Windows, `--no-cov`, stubs).** `TestDeferredEPGAssignment` (+7), `TestSummarizeFailedActionsCarriesReason` (+4), new `TestPass5ReviewFollowups` (5), `TestPass5DeferredEpgRetryFailureAggregation`, version consistency: 53 passed. Full `test_channel_pipeline_executor.py` + `test_channel_pipeline_engine.py` + attach-execution + parity + plans: FAILED id set byte-identical to the previous head (127 Windows-only failures on both sides). The prepare/commit integration you ran on Linux is not reproducible here; CI is the check for that path. Build stays `0036`.

MotWakorb (changes requested):
## Strict delta review: required integration regressions are still missing

Reviewed remediation head `b42d0972` against the frozen seven-item matrix. The implementation-level tests exercise the corrected branches, but several original required regressions stop at mocked unit boundaries.

### Requested correction

1. Retain the required prepare-to-commit regression for a planned create/assign run. Assert the actual committed channel EPG fields, refresh activity, recorded work, and persisted execution status; a green execution with an omitted assignment must fail.
2. Exercise deferred target pinning through a later context-changing action and the real Pass 5 retry. Assert the actual PATCH and logs identify the original channel, not only the queued tuple.
3. Exercise separate dummy profiles with SQLite-backed profile membership and resulting channel guide associations. The current test uses mocked profiles and does not prove persisted profile isolation or the equal-name guide outcome.
4. Exercise duplicate-name provenance through actual create/select processing rather than manually seeding the cache and ID set.
5. Persist a terminal deferred failure and assert the execution row selected-rule outcome and error count. Calling `_selected_rule_outcomes` directly does not prove end-to-end attribution.
6. Retain terminal-retry assertions at the persisted log/status boundary, ensuring no successful re-deferral promise survives.

Item 7 summary redaction and bounds are covered. These requests preserve the original acceptance matrix and do not ask for a pipeline redesign.

lukeeexd:
All six integration regressions added in `a6355cbf` (rebased onto `origin/dev`; build is `0040` because `dev` landed at `0036`).

New file `tests/integration/test_gh1011_same_run_dummy_epg_end_to_end.py`. Every scenario runs the real `ChannelPipelineEngine.run_pipeline` with rules and `DummyEPGProfile` rows in SQLite, the real `ActionExecutor`, and the real Pass 5 `_refresh_dummy_epg_and_retry` against a stateful in-memory Dispatcharr. Only the XMLTV regeneration and the source-refresh wait are doubled, the same two seams the existing `test_event_sync_dummy_epg.py` harness doubles; regeneration adds one entry per channel this run created, on the source serving the channel's group. Each stale source carries two entries so the executor's single-entry fallback cannot mask a deferral.

**1. Prepare → commit.** `_materialize_pipeline_plan` runs a real `plan_only` pipeline through `PlanningDispatcharrClient`, then the real `POST /api/auto-creation/run/commit` over ASGI replays it. The channel is created; no regeneration runs; no `epg_data_id` PATCH is sent; the execution persists `completed_with_errors`; `error_message` and the failed `assign_epg` say a planned run cannot regenerate and refresh before commit; no `[Pass 5` entry and no "Would refresh"/"Would retry" is persisted. Control: the same rule run directly ends `completed` with the entry assigned after Pass 5 and the profile gaining the group. RED: removing the `plan_only` guard persists the planned commit as `completed`.

**2. Pinned target through a later context change and the real retry.** Rule: create → assign → a second create (`{stream_name} Backup`). Pass 5's PATCH targets the original channel only (`[{"epg_data_id": 500}]` on it, nothing on the later one), and the `[Pass 5 Retry]` entry's `entity_id` is the original in both the result and the persisted log. RED: queueing the live context instead of `_pin_context` leaves the original unassigned.

**3. Profile isolation with SQLite membership and equal names.** Rule A (provider 1, group 1644, source 41) and Rule B (provider 2, group 1645, source 42), both streams named "Snooker". Persisted membership ends `[1644]` for profile 7 and `[1645]` for profile 8; the group-1644 channel carries source 41's entry and the group-1645 channel source 42's, different ids. RED: reverting to the group union gives profile 7 `[1644, 1645]`.

**4. Duplicate-name provenance through real create/select.** Stream 201 (provider 1) creates "Snooker" in 1644; stream 301 (provider 2) creates "Snooker" in 1645 and overwrites the name cache; stream 202 (provider 1) re-selects the 1644 channel through `create_channel` `if_exists: skip` and assigns. No hard `assign_epg` failure anywhere in the log, `failed_actions == []`, both channels assigned, execution `completed`.

**5 and 6. Terminal retry, persisted and attributed.** A selected run (`triggered_by="scheduled_selected"`, `require_all_rule_ids=True`) whose regeneration never covers the channel: the execution row is `completed_with_errors`, `get_selected_rule_outcome_state()` reports the rule as `completed_with_errors` with `error_count == 1`, the failed action carries the rule id and name, the persisted `[Pass 5 Retry]` entry is `success: false` with "even after the dummy EPG was regenerated and the source refreshed", no entry after Pass 5 begins says "Deferred: will assign EPG", the `[Pass 5] Summary` is a failure, and upstream received no guide-data PATCH.

**Verification (Windows, `--no-cov`, local POSIX shims).** New file, executor, engine, dummy-EPG, attach execution, run-summary invariants, version consistency, pipeline router, plans router, event-sync cleanup, scheduled selection, planning parity, MCP route enforcement: 1057 passed, 0 failed. `scripts/backend-gate.sh` not run here; CI is the check.

MotWakorb (approved):
## Strict delta review: approved

Reviewed `b42d0972..a6355cbf` against the frozen seven-item matrix with all ten team perspectives. The new retained integration suite closes the six evidence gaps through the requested real boundaries: prepare-to-commit status and effects, pinned retry target, SQLite-backed profile isolation, duplicate-name provenance, persisted rule attribution, and terminal retry logging. Previously accepted summary redaction/bounds and documentation remain closed.

All current required checks pass. No new in-scope findings were identified.

Issue #1013, opened 2026-09-16
Planned channel-pipeline commit returns 502 with zero writes when a stream's logo URL already exists as a Dispatcharr logo row


lukeeexd:
## Summary

A single duplicate-logo `400` from Dispatcharr aborts an entire channel-pipeline commit (`POST /api/channel-pipeline/run/commit`) with **zero** writes landed. The failing write is a `create_logo` that ECM's `create_channel` path adds for the matched stream's `tvg-logo` URL, even when the rule has no logo action. This has repeatedly killed a scheduled auto-creation rule in production (~15 event channels per run) and re-arms itself every cycle.

## Root cause

- Dispatcharr's `POST /api/channels/logos/` returns `400` when a logo row with the same `url` already exists.
- Logo rows **outlive** their channels: deleting a channel does not delete its logo row, so a row left over from a previous event cycle collides with the next run.
- `ActionExecutor._resolve_logo_id` creates first and only falls back to `find_logo_by_url` when the create raises. Under `PlanningDispatcharrClient` the create never reaches Dispatcharr, so the plan records a `create_logo` write with a temp id for every channel whose logo already exists.
- `replay_write_plan` is fail-fast: the real `create_logo` raises on the 400, the replay raises `PartialReplayError: pipeline replay failed at write 0`, compensation runs, and the endpoint returns `502`. Every planned channel create, stream merge and EPG assignment behind it is lost.
- The generic `_request` log line says only `[DISPATCHARR] API request failed: POST - status: 400`; the body that names the duplicate URL is not logged anywhere.

## Evidence (2026-09-16, UTC)

- 20:09–20:30: five commit attempts (including a minimal 3-write probe rule) all returned 502, zero completed writes, empty journal.
- Container log at the failure:
  ```
  POST http://<dispatcharr>/api/channels/logos/ "HTTP/1.1 400 Bad Request"
  [DISPATCHARR] API request failed: POST - status: 400
  POST /api/channel-pipeline/run/commit -> 502 in 1542.1ms
  ```
- UI: `PartialReplayError: pipeline replay failed at write 0`.
- Every write primitive succeeds individually (single/bulk channel create, stream merge, EPG assign); only the replay dies, always on the logo POST.
- Hand-POSTing `/api/channels/logos/` with the streams' exact `tvg-logo` URL returns 400 because a row with that URL exists.
[...]

PR #1014, opened 2026-09-16
v0.18.2-0037: fix #1013 duplicate logo URL no longer aborts a planned channel-pipeline commit

lukeeexd:
# Pull Request

## Summary
Closes #1013. A planned channel-pipeline commit (`prepare` → `commit`) returned 502 with **zero** writes landed whenever a matched stream's `tvg-logo` URL already existed as a Dispatcharr logo row. Dispatcharr rejects the duplicate URL with 400, logo rows outlive their channels, and the replay is fail-fast, so one stale row from a previous event cycle cost every planned channel create, stream merge and EPG assignment behind it, and re-armed itself on the next rollover.

- **`channel_pipeline_executor.py` — `_resolve_logo_id` looks the URL up before creating.** Under `PlanningDispatcharrClient` a create never reaches Dispatcharr, so the old create-first order recorded a `create_logo` write for every channel whose logo already existed. The lookup is a read (delegated to the live client in planning), so a known logo now yields no `create_logo` write and the created channel references the existing row. Result cached per run as before; a lookup failure falls through to the create.
- **`dispatcharr_client.py` — `create_logo` is idempotent on `url`.** Resolves an existing row before the POST and again after a 400 (row created in between), and logs the response body (bounded to 300 chars) on failure; the generic `_request` line said only `status: 400`. Every existing caller (executor, `POST /api/channels/logos`, DBAS logo importer) already treated "exists" as "use that one", so this collapses their duplicated fallbacks rather than changing intent. No new raw `_request` call sites; the contract sweep is unchanged.
- **`services/pipeline_write_plan.py` — `create_logo` is a soft-fail write.** New `SOFT_FAIL_WRITE_METHODS = {"create_logo"}`, classified by recorded write method, never by HTTP status. A failing soft write is logged and skipped, its result is `None`, and its temporary id maps to `None` so later payloads carry `logo_id: None` instead of an unresolved temp id. Channel create, stream merge, EPG, group and profile-membership writes keep the fail-fast abort and compensation exactly as before.
- `CHANGELOG.md` entry under `[Unreleased]` → Fixed. Build bumped to `0.18.2-0037` in all three touchpoints (`origin/dev` is `0035`; `0036` is already claimed by the open #1006 / #1008 / #1010 / #1012, so this avoids reusing a build number; whichever set merges second still needs a one-line rebump).
- No bead: external contribution, linked to GitHub #1013 instead. No `docs/` change: no guide currently describes the plan-replay abort semantics; happy to add one if you want it recorded somewhere beyond the CHANGELOG.

**Overlap note:** #1010 also edits the `except PartialReplayError` path (502 → 424) and `PartialReplayError.__init__`. This PR does not touch either; it only adds the soft-fail branch inside the `for` loop of `replay_write_plan`, so the two should merge cleanly in either order.

## Test plan
- [x] RED first on every new case (all failed against `dev` for the intended reason, then passed):
  - `tests/unit/test_dispatcharr_client_logo_create.py` (new, 6): known URL → no POST, returns the existing row; unknown URL → POST; pre-check miss then POST 400 → resolved by URL (race); 400 with no matching row → raises with the body; non-400 error → raises without a second lookup; payload without `url` → POSTs directly.
  - `tests/services/test_pipeline_write_plan.py` (+4): failed `create_logo` → replay continues, channel created with `logo_id: None`, `remap == {-1: None, -2: 101}`, no compensation; failed `create_channel` after a successful logo → still `PartialReplayError` at index 1 with the logo in `completed`; `create_logo` resolving to id 765 → channel created with `logo_id: 765`; end-to-end through the real `DispatcharrClient` with `_request` faked to answer 400 on the logo POST and list the colliding row → replay completes and `POST /api/channels/channels/` carries `logo_id: 765`.
  - `tests/unit/test_channel_pipeline_executor.py::TestResolveLogoIdReusesExistingRow` (new, 4): existing row reused and `create_logo` not awaited; unknown → created; lookup exception → falls through to create; second call for the same URL served from the per-run cache.
- [x] Surrounding suites on Windows with `--no-cov` and `fcntl`/`resource` stubs (`test_channel_pipeline_executor`, `test_channel_pipeline_engine`, all `test_dispatcharr_client_*`, `routers/test_channels`, `routers/test_logos`, `routers/test_channel_pipeline`, `routers/test_channel_pipeline_plans`, `services/test_event_sync_cleanup`, `dbas/test_logos_importer`, `test_04c0u7_planned_execution_parity`, `test_mcp_planned_route_enforcement`): the FAILED/ERROR id set is **byte-identical** to the same run on an untouched `origin/dev` stash (557 Windows-only failures on both sides: `os.O_DIRECTORY`, `O_NOFOLLOW`, api-key file-mode checks). Zero new failures, zero newly passing.
- [x] `test_version_touchpoint_consistency.py` and `test_dispatcharr_client_contract_sweep.py` green; `py_compile` on every changed module.
- [ ] Not run here: the full `scripts/backend-gate.sh` (Linux-only backend). CI is the authoritative check.
- [ ] Not yet re-verified against the production Dispatcharr from this branch; the machine that reproduced it runs the published image. The manual workaround (deleting the orphan row) and the re-collision on the following run are documented in #1013.

## Checklist
- [ ] Bead linked in the commit message (`[bd-xxxxx]`) — none; external contribution, commit says `Closes #1013`.
- [x] Tests updated or added (backend `pytest`, frontend `vitest`) — backend only; no frontend change beyond the version bump.
- [ ] Docs updated if behaviour or interfaces changed — none touched; see Summary.


MotWakorb (changes requested):
## Full team review — requested fixes

Reviewed `030c93631e69f8d945af8dae96b9f45b2b426acb` against `086e3e4fcec7eb0d334c4b37f08f20e82507d447`, covering all ten changed files with ten team perspectives, calibrated to the affected components' home-lab tier.

The duplicate-logo fix addresses #1013, and method-based best-effort artwork handling is appropriate. However, the shared-client reuse contract introduces destructive DBAS rollback ownership, and the soft-failure path can remove existing artwork and persist inaccurate history. **Please address all six findings and the test/documentation corrections below.**

### 1. High / P1 — Preserve created-versus-reused ownership for DBAS rollback

**Anchor:** [`backend/dispatcharr_client.py:1286–1299`](https://github.com/MotWakorb/enhancedchannelmanager/blob/030c93631e69f8d945af8dae96b9f45b2b426acb/backend/dispatcharr_client.py#L1286-L1299).

Both reuse branches return an existing row indistinguishably from a newly created one. DBAS's URL importer treats a successful return as a creation, increments its creation counters, and records the ID in the compensating-delete ledger (`backend/dbas/importers/logos.py:582–589`).

**Independently reproduced:** A supported URL-only archive record has a different name from the destination logo but the same URL. Lookup returns existing logo **765**, with no creation POST. DBAS reports it as created and ledgers it; invoking the real rollback deletes logo 765 and reports completion. The post-400 concurrent-create case fails the same preservation contract. Equivalent base-client ownership controls preserve the existing logo.

**Requested correction:** Preserve an explicit distinction between actual creation and reuse at the shared-client/consumer boundary. A reused ID may populate a mapping but must not grant compensating-delete ownership or increment creation counts.

**Required regression:** Existing-row reuse and post-400 race reuse survive DBAS rollback; genuinely created rows remain compensable. Exercise the populated destination, actual importer, and rollback ledger/consumer.

Ordinary per-logo failures do not automatically trigger the deletion. The destructive consequence requires rollback to consume the incorrect ledger; the false ownership entry occurs immediately on successful reuse.

### 2. Medium / P2 — Preserve existing artwork when a replacement fails

**Anchor:** [`backend/services/pipeline_write_plan.py:302–306`](https://github.com/MotWakorb/enhancedchannelmanager/blob/030c93631e69f8d945af8dae96b9f45b2b426acb/backend/services/pipeline_write_plan.py#L302-L306).

Mapping a failed logo's temporary ID to `None` affects existing-channel updates as well as new-channel creates. A planned `assign_logo` becomes `PATCH {"logo_id": null}`, removing the current assignment even though its replacement never became available.

**Independently reproduced:** A channel's logo changed from **99 to null** after the replacement failed. The direct executor control preserved 99 and skipped assignment. Confirmed snapshot rollback subsequently reported success but left the logo association null, because its existing restore payload omits logo state.

**Requested correction:** Distinguish optional artwork for a new channel from a failed replacement on an existing channel. Omit the failed replacement field, or skip a logo-only update, while preserving independent structural writes. Avoid relying on a broader rollback redesign to undo an unintended clear.

**Required regression:** Existing-channel replacement failure preserves the prior logo, including through prepare/commit and persisted outcomes. New channels still proceed without a logo, and unrelated structural fields still apply.

### 3. Medium / P2 — Persist skipped outcomes instead of reconstructing successful logo creation

**Anchor:** `backend/services/pipeline_write_plan.py:297–306`; affected journal reconstruction at `379–389`.

The replay skip is represented by a transient warning and a `None` result, but finalization discards replay results and reconstructs successes from the original plan.

**Independently verified through prepare/commit and SQLite:** a failed logo creates this durable journal entry:

```text
action_type: create_logo
entity_id: null
description: Planned pipeline executed create_logo for None
```

The execution finishes without a persisted skipped-logo warning; a replacement action can retain the optimistic description “Assigned logo to channel.”

**Requested correction:** Carry actual replay outcomes into finalization. Exclude skipped creates from successful-mutation journal entries and retain a durable, non-fatal warning identifying the omitted work. Best-effort channel completion can remain successful without claiming every planned operation executed.

**Required regression:** A failed cosmetic create followed by successful structural work yields truthful execution/log/snapshot/journal evidence and an operator-visible skip or warning. Keep structural work non-fatal only where the method-based policy permits it.

### 4. Low / P3 — Report the actual failing plan index after soft skips

**Anchor:** `backend/services/pipeline_write_plan.py:288–306`, interacting with `351–353`.

Skipped writes do not enter `completed`, but `PartialReplayError.failed_index` still uses `len(completed)`.

**Independently reproduced:** `[create_logo skipped, create_channel successful, update_channel failed]` reports and persists index **1**, although the failing write is index **2**. Structural abort and channel compensation still work; recovery diagnostics identify the wrong operation.

**Requested correction:** Use the actual enumerated plan position for failure location and keep the successful-write inventory separate.

**Required regression:** One or more skipped logos precede a successful structural write and a failing structural write. Assert the exception index, persisted execution/snapshot index, and unchanged compensation behavior.

### 5. Medium / P2 — Avoid repeated full-catalog prechecks beneath run-local lookup caches

**Anchor:** `backend/dispatcharr_client.py:1286–1288`.

The real bulk caller already builds a per-run logo index. Each distinct missing URL now triggers another paginated catalog scan inside `create_logo()`.

**Independent request-count comparison:** With a 1,001-row logo catalog and five channel operations—three distinct new URLs, one repeated new URL, and one existing URL—the base makes **3 logo GETs**, while HEAD makes **12**. Both make three logo POSTs and successfully apply five channel operations.

**Requested correction:** Let an already-resolved catalog miss avoid another unconditional precheck, while retaining lookup reuse and post-400 race reconciliation. Preserve the existing run-scoped indexing behavior.

**Required regression:** Check catalog request counts through the real bulk caller/client boundary for multiple new URLs, a repeated URL, and an existing URL. A mocked `create_logo` cannot detect scans introduced inside that method.

This is a confirmed request-amplification regression, not a measured production latency or throttling incident. Known and repeated URLs still benefit from the caller's cache.

### 6. Low / P3 — Keep credential-bearing response bodies out of the new warning

**Anchor:** [`backend/dispatcharr_client.py:1304–1307`](https://github.com/MotWakorb/enhancedchannelmanager/blob/030c93631e69f8d945af8dae96b9f45b2b426acb/backend/dispatcharr_client.py#L1304-L1307).

Truncating the raw response to 300 characters limits size, not sensitivity. The warning emits before DBAS sanitizes its operator-facing failure report.

**Independently reproduced through production logging handlers:** a registered synthetic credential reached console and ring-buffer output. An unregistered token embedded in a reflected logo URL path also reached persistent logging. The DBAS report remained sanitized, and equivalent base-client cases did not emit those values.

**Requested correction:** Log status and safe diagnostic metadata or allow-listed validation information, rather than arbitrary upstream response text. Apply confidentiality protection before any length bound.

**Required regression:** Exercise the actual logging handlers with known-secret and credential-bearing URL fixtures; the synthetic secrets must not appear in console, ring-buffer, or persistent output. Retain useful failure classification.

The risk requires sensitive upstream content and access to the resulting logs. The proof uses synthetic canaries, not production credentials, and does not establish an authentication bypass or Critical exploit.

### Test and documentation corrections

- `backend/tests/services/test_pipeline_write_plan.py:163–198` names duplicate-400 survival, but its initial GET already returns the existing logo, so its POST/400 branch is never reached. Make the fixture execute empty precheck → POST 400 → successful lookup and assert that sequence, or accurately describe the narrower test. The separate client unit test does cover the race; the replay test currently overstates its coverage.
- Correct `create_logo()`'s claim at `backend/dispatcharr_client.py:1280–1281` that every caller already treats reuse equivalently. DBAS's successful-create ownership contract is the demonstrated exception.
- Align the changelog and behavior descriptions with the corrected soft-failure result: new-channel missing artwork, preservation of existing artwork, and a truthful durable skipped-operation signal. Explain how an operator can recognize cosmetic degradation after useful channel work completes.

| Item | Expected outcome |
| --- | --- |
| 1 | Reused/concurrently created logos never become restore-owned deletion targets; genuine creation remains correctly counted and compensable. |
| 2 | Failed replacement preserves existing artwork while new channels and independent structural writes can proceed. |
| 3 | Replay finalization persists actual skipped/successful outcomes, without phantom logo-create journal entries. |
| 4 | Exception and persisted recovery evidence identify the true failing plan position after any soft skips. |
| 5 | Run-scoped logo indexing prevents redundant complete catalog scans while duplicate-race recovery remains covered. |
| 6 | The new logging path does not emit credential-bearing response content through production handlers. |
| Tests/docs | Claimed POST/400 coverage is real, and descriptions accurately distinguish reuse ownership and cosmetic degradation. |

Follow-up review should verify this matrix and inspect the remediation delta for regressions. The method-based best-effort policy is explicitly requested by #1013; these corrections preserve that intent rather than reopening it.

- Independently rerun surrounding Linux suites: **1,371 passed**, including all new tests, client/executor/engine, relevant routers, replay/cleanup, DBAS logos, planning parity, route enforcement, and version consistency.
- Prepare/commit and DBAS fault probes: **7 passed, 4 failed**, confirming the ownership regression in two cases, phantom journal success, and the incorrect failure index.
- Additional independent consumer proofs reproduced existing-artwork clearing and ineffective restoration, production-handler logging disclosure, and the catalog request amplification.
- Positive controls confirmed existing-logo reuse, new-logo creation, actual POST/400 race recovery, and structural abort/compensation.

These are controlled upstream integrations with real ECM code and isolated persistence, not deployed Dispatcharr acceptance or a full CI gate. At the review intake, GitHub reports the PR blocked with no published check results; required CI still needs to complete on the proposed merge revision.

**Sequencing:** No semantic dependency on PR #1010 was established; its retry/status changes are absent from this reviewed HEAD. Reconcile build `0037` in the three version touchpoints and changelog with actual landing order if necessary.

lukeeexd:
All six items and the test/doc corrections addressed in `77dd87fa`.

**1. Created vs reused (P1).** `create_logo` marks a row it *resolved* (pre-check hit or post-400 reconciliation) with `LOGO_REUSED_KEY`; a fresh create never carries it. Helpers `logo_was_reused` / `strip_logo_reuse_marker` sit next to it. The DBAS URL importer remaps a reused id for channel reattach but does **not** ledger it for compensating delete or count it as created; it is reported as skipped/`ALREADY_EXISTS_IDENTICAL` and `matched`. Asserted with a real `RollbackLedger`: reuse leaves `ledger.entries` and `compensation_order()` empty while the FK remap resolves to 765; a genuine create is still ledgered with its destination id. Bulk `createChannel` journals a `logo_create` row only for a real create; `POST /api/channels/logos` strips the marker from its response. The `create_logo` docstring now names DBAS as the ownership exception.

**2. Existing artwork preserved.** A payload field that referenced a skipped create is *omitted* from the request, never sent as null: `update_channel(7, {"logo_id": -1})` is not sent at all (reason `dependency_skipped`), `{"logo_id": -1, "name": "Renamed"}` is sent as `{"name": "Renamed"}`, and a new channel is created without the field. A *positional* dependency on a skipped create (`update_logo(-1, …)`) still aborts.

**3. Truthful finalization.** `replay_write_plan` reports through a `ReplayOutcome` (skipped writes with reason and error type, reused creates). `journal_entries_for_plan` no longer reconstructs a success row for a skipped or reused create or an unsent update, and the execution row gains a `replay_write_skipped` warning per skipped write (also appended to the execution log, with `channel_id` for updates). New `tests/routers/test_gh1013_commit_skip_evidence.py` drives `commit_auto_creation_pipeline` with a stored plan and a replay double: the execution is `completed`, carries both warnings, and `journal.log_entries` receives only the `create_channel` row (no "for None").

**4. True `failed_index`.** `[create_logo skipped, create_channel ok, update_channel fails]` reports index 2; compensation unchanged.

**5. No per-URL catalog rescan.** `create_logo(data, *, precheck=True)`. The pipeline executor (after `find_logo_by_url`) and the bulk router (after its run-scoped index) pass `precheck=False`; the planning client records the kwarg so replay passes it on. Post-400 race reconciliation is retained and asserted (`POST` → `GET` sequence with `precheck=False`). A new bulk-router test asserts `create_logo` is awaited with `precheck=False` after a single catalog page.

**6. No response body in logs.** The failure log and the exception carry the status and a fixed classification (`duplicate_url` / `validation` / `server_error` / `error`), never the body. A canary token reflected in the body and in the URL is asserted absent from `caplog.text` (DEBUG level) and from the exception; the executor's failure log names only the logo URL's host.

**Tests/docs.** The real-client replay test now drives pre-check miss → POST 400 → lookup hit and asserts the request sequence and `outcome.reused == [0]`. Changelog describes reuse ownership, artwork preservation, the durable skip signal and how an operator recognises cosmetic degradation.

**Verification (Windows, `--no-cov`, stubs).** `test_dispatcharr_client_logo_create`, `services/test_pipeline_write_plan`, `dbas/test_logos_importer`, `TestResolveLogoIdReusesExistingRow`, parity, contract sweep: 145 passed; the commit-evidence test passes. Wider executor/channels/logos/plans/cleanup/MCP suites: FAILED id set identical to the previous head except the one new bulk-router test, which fails locally only for the same `os.O_NOFOLLOW` reason as its 69 sibling tests in that class (CI-only evidence).

**Merge note.** #1010's follow-up also edits the `except` block of `replay_write_plan`; whichever lands second needs a small rebase there. Build stays `0037`.

MotWakorb (changes requested):
## Strict delta review: evidence gaps and contradictory changelog remain

Reviewed remediation head `77dd87fa` against the frozen six-item matrix. The code changes address the reported branches, but required cross-boundary regressions remain missing and the changelog still describes both the broken and corrected behavior.

### Requested correction

1. Drive both pre-check reuse and post-400 race reuse through the real client, DBAS importer, and rollback consumer. Assert reused logos survive rollback and genuine creations remain compensable; a mocked marker plus an empty-ledger assertion does not exercise the ownership boundary end to end.
2. Add the required prepare-to-commit regression for failed replacement artwork. Assert the existing logo remains assigned, independent structural fields apply, and persisted execution evidence is truthful.
3. Use the real replay in the commit-evidence test rather than injecting a synthetic `ReplayOutcome`, so skipped writes, warnings, and journal rows are proven to originate from the production replay path.
4. Assert the true failing index in persisted execution and snapshot recovery evidence after soft skips, not only on the thrown exception.
5. Exercise the real bulk caller plus real client request path for multiple new URLs, a repeated URL, and an existing URL. The current bulk test mocks `create_logo`, so it cannot detect catalog rescans inside that method.
6. Exercise the production console, ring-buffer, and persistent logging handlers with credential canaries. `caplog` alone does not satisfy the original logging-handler regression.
7. Correct `CHANGELOG.md`: it currently says the client logs the response body and that dependent payloads carry `None`, then later says response bodies are not logged and failed fields are omitted. Describe only the final corrected behavior.

These are the frozen matrix evidence and documentation requirements; no new best-effort policy is being introduced.

lukeeexd:
All seven items addressed in `ba211015` (rebased onto `origin/dev`; build stays `0037`).

Every new test runs the real `DispatcharrClient` over a controlled `httpx.MockTransport` that plays the logo catalog and channel endpoints and records each request, so the pre-check, the POST, the post-400 reconciliation and the reuse marker are all the production code. New file: `tests/integration/test_gh1013_logo_replay_evidence.py`.

**1. Reuse ownership through the real client, importer and rollback consumer.** `test_precheck_reuse_survives_dbas_rollback` (catalog holds id 765 under a different name → `import_logos` → `create_logo` pre-check hits → zero logo POSTs) and `test_post_400_race_reuse_survives_dbas_rollback` (the transport hides the catalog until a POST has happened, so the importer's destination scan and the pre-check both miss, the POST answers 400, the post-400 lookup finds 765): in both, `cat.created == 0`, the remap resolves 42→765, the ledger has no LOGO entry, and the real `run_rollback` compensates nothing — no `DELETE /logos/765/` is sent and 765 is still upstream. Control: an empty catalog → created 900, ledgered, and rollback sends `DELETE /logos/900/`. RED: disabling the importer's reuse branch fails both reuse tests on the created count.

**2. Prepare → commit for failed replacement artwork.** `test_failed_replacement_artwork_leaves_the_existing_logo_and_applies_the_rename`: `_materialize_pipeline_plan` then the real `POST /api/auto-creation/run/commit` over ASGI with the real replay. The replacement logo collides (400) and lookups cannot reconcile it, so the create soft-fails; upstream receives exactly `PATCH /channels/7/ {"name": "Renamed"}`, channel 7 keeps `logo_id 99`, the execution is `completed` with one `replay_write_skipped` (`create_logo`, `soft_failure`), and the journal has no `create_logo` row and no "for None". `test_logo_only_replacement_is_not_sent_…`: the logo-only update is never sent, is persisted as `dependency_skipped` with `channel_id 7`, and the new channel is created as `{"name": "New", "streams": [5]}` with no logo field.

**3. Real replay in the commit-evidence test.** `tests/routers/test_gh1013_commit_skip_evidence.py` no longer injects a `ReplayOutcome`; the handler runs the production `replay_write_plan` against the transport double. Same warning/journal assertions as before, plus what upstream received (one rejected logo POST, a channel create without a logo field, no PATCH to channel 7).

**4. True failing index persisted.** `test_true_failing_index_is_persisted_after_a_soft_skip`: `[create_logo skipped, create_channel → 101, PATCH /channels/101/ → 500]` persists `failed_index: 2` in the execution log's `partial_replay_failure` entry and in the snapshot's `partial_replay` evidence, with the compensation `DELETE /channels/101/` observed.

**5. Real bulk caller, real client, request counts.** `test_bulk_create_scans_the_logo_catalog_once_for_many_urls`: `POST /api/channels/bulk-commit` (202 + poll) with a 1001-row catalog and five `createChannel` ops (three new URLs, one repeated, one existing) issues exactly 3 `GET /logos/` (one paginated scan) and 3 `POST /logos/`; the repeated URL reuses the in-run id and the existing URL resolves to 500 from the index. RED: re-enabling the per-URL pre-check fails at 12 GETs.

**6. Production handlers.** `test_logo_create_failure_reaches_no_production_handler_with_the_body`: a `StreamHandler` using `main.py`'s format, the `RingBufferHandler`, and `install_persistent_json_logging` are attached to the root logger; the upstream 400 body echoes a canary token and a `register_sensitive_values` secret. Console and ring carry `Logo creation failed: status 400 (duplicate_url)` and nothing from the body (canary, secret, host, "already exists" all absent); the raised exception is equally clean. The persistent JSON handler cannot install on Windows (it opens the log directory as a descriptor), so that assertion is CI-verified and explicitly skipped here after the console and ring assertions have run.

**7. CHANGELOG.** The entry now describes only the final behaviour: no response-body logging, omitted fields rather than `null`, reuse ownership, one catalog scan per run, true failing index, and how an operator recognises cosmetic degradation from the execution's warnings.

**Verification (Windows, `--no-cov`, local POSIX shims).** Evidence files, logo-create client, write plan, DBAS logos importer, channels router, logos router, pipeline router, plans router, event-sync cleanup, executor, contract sweep, version consistency, log_utils, planning parity, MCP route enforcement, restore orchestrator: 1193 passed, 1 skipped, and the only failures are the 20 `TestInterProcessRotatingJsonHandler` cases that fail identically on the untouched branch (Windows-only). `scripts/backend-gate.sh` not run here; CI is the check.

MotWakorb (approved):
## Strict delta review: matrix approved

Reviewed `77dd87fa..ba211015` against the frozen six-item matrix with all ten team perspectives. The update now retains the requested cross-boundary evidence for DBAS reuse ownership and rollback, prepare-to-commit artwork preservation, production replay and durable skip evidence, persisted failure index, real bulk catalog request counts, production logging handlers, and the corrected changelog. No new in-scope defect was identified.

The current Backend Tests check is still red on `test_rotates_at_small_bound_and_snapshot_is_oldest_first`. Its output shows a capacity-only eviction with unchanged `sequence-*` messages, unlike the delta-caused redactor contamination on #1010; the team could not attribute it to this PR. The frozen matrix is approved, but the required gate must rerun green before merge.

MotWakorb (changes requested):
## Corrected strict delta verdict: test-state regression blocks

This supersedes my approval after the failed Backend Tests rerun produced the same result and a focused local sequence established delta attribution. The frozen six-item product matrix remains accepted; this request does not reopen any product behavior.

`backend/tests/services/test_pipeline_write_plan.py:179-180` creates a real `DispatcharrClient` with one-character username/password values `"a"` and `"b"`. Client construction registers them in the process-global sensitive-value redactor. The later rotation test formats logger name `rotation.test`; registered `"a"` expands there to the redaction marker, increases each JSON record beyond half the 300-byte file limit, and causes the first two records to be evicted.

Evidence:

- The required Backend Tests job failed twice at the same assertion, including the explicit rerun.
- The rotation test passes alone on current `dev`.
- Running only this PRs changed `test_pipeline_write_plan.py` before that assertion reproduces the exact failure: 17 passed, 1 failed.
- The other changed test files tested independently do not reproduce it.

Use nontrivial synthetic credentials that cannot collide with ordinary log metadata and keep process-global redactor state isolated between tests. Then rerun the required backend gate. All six original acceptance items remain closed.

lukeeexd:
Test-state regression fixed in `3728589b`.

**Cause confirmed.** `test_pipeline_write_plan.py` built a real `DispatcharrClient` with `username="a"` / `password="b"`; construction registers those with the process-lifetime redactor, and `"a"` then expanded to the redaction marker inside the rotation test's logger name, pushing each record past half the 300-byte bound.

**Fix, two parts.**
- Long, unique synthetic credentials everywhere this PR constructs a real client: the write-plan race test (`test-only-write-plan-…`), the two GH #1013 evidence files (`test-only-gh1013-api-key-…`), and `test_dispatcharr_client_logo_create.py`, which used `"admin"` / `"secret"` and would have redacted the word *secret* wherever it later appeared in a log line.
- `tests/conftest.py` gains an autouse fixture that snapshots the registered sensitive-value forms before each test and restores them afterwards, so no test can leak a registration into another regardless of ordering. Same fixture text as on #1010 so the two branches merge cleanly; probed there with a two-test file (register `"k"`, then assert it is gone and `sequence-1` survives the scrubber).

The rotation test itself cannot run on Windows (POSIX directory descriptor), so the backend gate is the check for that assertion; the fixture removes the mechanism it failed on. Locally (Windows, `--no-cov`, shims), the changed files run ahead of `test_log_utils.py` as in your repro: 306 passed, and the only failures are the same 20 `TestInterProcessRotatingJsonHandler` cases that fail identically on the untouched branch. The six-item matrix is untouched by this commit.

MotWakorb:
Superseded and landed through maintainer integration PR #1019 because this fork branch had maintainer edits disabled. #1019 preserved the original contributor commits, resolved the post-#1012 conflicts, passed fresh CI, and merged into dev.

Issue #1015, opened 2026-09-18
[Bug]: M3U-refresh dedup queues the next day's event fixtures as duplicates of yesterday's channels, freezing the group at 0 channels created


lukeeexd:
### ECM Version

Observed on 0.18.1. Source references below re-verified against the current checkout (`030c9363`, v0.18.2-0037); none of the paths involved have changed.

### Priority

Priority 2: ECM is severely degraded

### What is the bug?

Event providers reuse a **fixed pool of slot names** (numbers 01..100) and roll the variable part of each name over daily — the draw, the players, and the date/time. The M3U-refresh dedup path ("BD-A" scorer, ADR-008 §D1) compares the incoming stream name against the existing channel names in the target group with no date/time awareness, so **today's fixture in a slot is scored as a duplicate of yesterday's fixture in the same slot**. At or above the queue threshold the stream is queued as a pending merge instead of creating a channel, and while a row is pending the auto-creation pipeline defers that stream: no channel created, no EPG assigned.

When the whole slot pool rolls over at once, every stream in the group is deferred, the run reports `0 channels created`, and the group sits on yesterday's matches for roughly a third of the day. There is **no operator workaround** — details below.

Names in the tables below are stand-ins that reproduce the shape of a real provider's slot naming; the scores are measured against the shipped matcher.

**The scorer is not date-aware, and a one-day difference is not a close call.** `find_candidate` normalizes both sides with `NameCleanMode.CONSERVATIVE` — NFC, strip a leading `N | ` prefix, lowercase, strip (`backend/services/dedup_matcher.py:522`, called at `:716` and `:730`) — and scores with `fuzz.token_set_ratio(...) / 100.0` (`:744`). Nothing in the matcher or in the reused `normalization_engine` helpers touches month names, `HH:MM`, AM/PM, `GMT`/`IST`, or an `@` marker. Measured with the shipped code and the default `dedup_threshold = 0.80` (`backend/config.py:891`):

| incoming stream | candidate channel | confidence |
|---|---|---|
| `EVENT SLOT 04: Championship \| Qualifying: Player One - Player Two @ 18 Sep 09:30 AM GMT-1` | `... Player Three - Player Four @ 17 Sep 01:00 PM GMT-1` | **0.862** → queued |
[...]

PR #1016, opened 2026-09-18
v0.18.2-0038: fix #1015 stop a provider's daily slot rollover queueing the whole group as duplicates

lukeeexd:
# Pull Request

## Summary

Closes #1015. Event providers hand out a **fixed pool of slot names** and roll the fixture and the airing over every day, so yesterday's channel for a slot and today's stream for the same slot share every template word. The M3U-refresh dedup scorer (`token_set_ratio`) put such a pair at **0.86–0.99** — above the default 0.80 — queued the stream as a pending merge, and deferred the channel creation. When the whole pool rolled over at once the run created nothing, every stream's follow-on `assign_epg` failed with the bare `No channel context for assign_epg`, and the group stayed on the previous day's fixtures until Pass 4 orphan cleanup happened to delete their candidate channels.

- **`services/dedup_matcher.py` — an airing gate ahead of the floor.** `find_candidate` now drops a candidate **before scoring** when both names state an airing (day + month + clock, optionally meridiem and zone) and the two airings differ. Stripping the date/time run instead would *not* work — with the run removed the two names are character-identical and the pair scores 1.00, i.e. still queued — so the run is treated as a signal rather than noise, in the same family as the existing M1 callsign hard-reject. `airing_key()` compares day, month, clock, meridiem and zone; it tolerates spelling noise (case, `Sep.`, `AM`/`a.m.`, leading zeros) and deliberately excludes the year, so one provider spelling `18 Sep 2026 09:30` still equals another's `18 Sep 09:30`. A name that states no airing is never dropped and pays nothing for the gate, so ordinary channel names score byte-for-byte as before. The 0.60 floor and the operator threshold are untouched; the gate can only ever *remove* a candidate, never admit one the floor would refuse.
- **`channel_pipeline_executor.py` — the candidate scope excludes the stream's own channel.** `_maybe_enqueue_pending_merge` built its list from every same-group channel, including a channel the stream is **already attached to**; that pair scores 1.00 by construction, so a stream could be queued as a merge into the very channel it is on (one row in the live incident was exactly that at confidence 1.00). Attachment is caller-side state, so the filter lives at the caller — the hook's "the caller owns candidate selection" contract is unchanged.
- **`channel_pipeline_executor.py` / `channel_pipeline_engine.py` / `tasks/channel_pipeline.py` — a deferred stream is no longer invisible.** `DedupHookResult` carries the blocking `pending_merges` row id on **both** the fresh-insert and the §D5 collision branch (the collision branch is the one that keeps a group frozen across refreshes, and it previously discarded the id). The engine aggregates `pending_merge_ids` beside `pending_merges_added`, the completion log and the post-refresh task summary name the deferral count and the rows, and a deferred stream's follow-on `assign_epg` reports the deferral instead of `No channel context for assign_epg`. `Auto-Creation: 0 channels created` with N failed actions was previously indistinguishable from "the provider had nothing today", and a single deferred slot read as an EPG fault.
- **`docs/adr/ADR-008-interactive-stream-dedup.md` — §D2 addendum + revision-history row.** Records the gate, the candidate-scope change and the two deliberate exclusions below.
- `CHANGELOG.md` under `[Unreleased]` → Fixed. Build bumped to `0.18.2-0038` in all three touchpoints (`origin/dev` is `0035`; `0036` and `0037` are claimed by the open #1010 and #1014, so this avoids reusing a build number — whichever set merges second still needs a one-line rebump).
- No bead: external contribution, linked to GitHub #1015 instead.

**Deliberately out of scope (flagged in the issue, and recorded in the ADR addendum).** Two of the reported asks are **not** here: auto-expiring a pending row whose candidate channel no longer exists, and making dismissal sticky per `(stream_name, candidate_channel_id)` rather than per row. Both change behaviour that is currently specified rather than accidental — ADR-008 §D4 defines the 404-on-accept path, migration 0014 states that "the same pair can re-appear after dismissal", there is a recorded PO decision on the accepted-but-not-applied row shape, and bead `enhancedchannelmanager-5136e` already tracks the retention reaper as deferred. Changing any of them is an ADR amendment for the PO, not a bug fix, so I left them out rather than fold a queue-semantics change into this PR. The freeze itself is fixed without them: a deleted candidate is not in the live candidate list, so it cannot defer a stream — it only lingers in the operator's queue. Say the word and I will open the follow-up.

**Overlap note:** #1010 / #1014 also bump the three version touchpoints and touch `backend/routers/backup.py` / `backend/main.py`. This PR touches only the version literals in those two files. Nothing here touches `services/pipeline_write_plan.py`, `dispatcharr_client.py`, or the replay path #1010 edits, so the sets should merge cleanly in any order.

## Test plan

- [x] RED first on every new case (failing against `dev` for the intended reason, then passing):
  - Removed the gate (`airing_key` → `None`) with the source otherwise byte-identical: `TestSlotRolloverCreatesTodaysChannels::test_full_pool_rollover_creates_and_queues_nothing` fails with `Stream '… @ 18 Sep 09:30 AM GMT-1' enqueued as pending merge candidate for existing channel id=100 (confidence=0.86); channel creation deferred to operator review` — the reported symptom, verbatim — and `TestAiringGate` fails with `confidence=0.862` / `confidence=0.988`. Both pass with the gate back in.
  - `tests/services/test_dedup_matcher.py` — `TestAiringKey` (7): provider shape, `None` when no airing, bare day+month is not an airing, spelling noise, year excluded, leading zeros, no-`@` shape, and slot-number/clock collisions (`SLOT 04:`, `Channel Name Number 250`) not reading as airings. `TestAiringGate` (8): the acceptance case (same slot, different fixture and day → no candidate), the score-alone guard proving the pair *would* have queued (asserts the raw `token_set_ratio ≥ 0.80`), next-day with an identical fixture, identical names still queue at 1.00, same airing still scores fuzzily, a genuine duplicate elsewhere in the same candidate list is still returned, and the two "gate does not fire" contracts.
  - `tests/services/test_m3u_dedup_hook.py` — `TestMergeIdExposure` (4): fresh insert reports the new row id, the §D5 collision reports the *existing* row id while `candidate` stays `None`, no match reports none, both short-circuits report none.
  - `tests/unit/test_channel_pipeline_executor_dedup_hook.py` (+7): the rollover scenario end-to-end through `_execute_create_channel` (two rolled-over slots → both created, zero pending rows, `pending_merge_ids == []`); the same-slot next-day case; the 1.00 self-flag no longer queued, with a control that differs **only** in attachment and still queues; dict-shaped `streams` entries understood; the deferral reason and row id surfacing on the `assign_epg` failure; and the unchanged legacy wording when there is no deferral.
  - `tests/tasks/test_gh1015_deferred_merge_visibility.py` (3): the summary names the count and the row ids, the frozen-group shape (0 created + failed actions) still reports the deferral and truncates a long id list, and a quiet run gains no deferral noise.
- [x] Surrounding suites on Windows with `--no-cov` and local `fcntl`/`resource`/`os` stubs (base-worktree comparison at `086e3e4f`, 1027 tests collected vs 1000 on the base, same selection both sides): `test_channel_pipeline_engine`, `test_channel_pipeline_executor`, `test_channel_pipeline_executor_dedup_hook`, `test_0hjrk_run_summary_invariants`, `test_version_touchpoint_consistency`, `services/test_dedup_matcher`, `services/test_dedup_matcher_properties`, `services/test_m3u_dedup_hook`, `routers/test_channel_merges`, `routers/test_channel_pipeline`, `integration/test_dedup_audit_trail`, `integration/test_pending_merges_migration`, `tasks/test_scheduled_selected_pipeline`. The FAILED/ERROR id set is **byte-identical** to the same run on an untouched `origin/dev` worktree — one pre-existing Windows failure on both sides (`TestDebugBundleEventSyncMatching::test_bundle_includes_event_sync_matching_with_per_stream_fields`, `KeyError: 'master_group_id'`, event-sync debug bundle, untouched by this PR). Zero new failures, zero newly passing.
- [x] `py_compile` on every changed backend module; `test_version_touchpoint_consistency.py` green (24).
- [ ] Not run here: the full `scripts/backend-gate.sh` (Linux-only backend; no container on this machine). CI is the authoritative check. A bare full-tree run on Windows is dominated by platform failures — glibc-only `strftime("%-I")` in `dummy_epg_engine.py`, missing IANA `tzdata`, POSIX file modes/paths — so I did not use it as evidence.
- [ ] Not re-verified against the live provider/group from this branch; the machine that reproduced it runs the published image. The daily rollover was reproducible on 0.18.1 for ~12 hours per day with no operator workaround.

## Checklist

- [ ] Bead linked in the commit message (`[bd-xxxxx]`) — none; external contribution, commit says `fix #1015`.
- [x] Tests updated or added (backend `pytest`) — backend only; no frontend change beyond the version bump.
- [x] Docs updated — ADR-008 §D2 addendum + revision history; no user-guide change (the operator-facing description of Pending Merges is unchanged).


MotWakorb (changes requested):
## Full team review — all seven findings requested for correction

Reviewed `23c417173f8fcf400dd8393cae9135ce2835816b` against `086e3e4fcec7eb0d334c4b37f08f20e82507d447`, covering all 14 changed files with ten team perspectives, calibrated to the affected components' home-lab tier.

The reported day-first rollover case is fixed, and fresh/collision row-ID propagation works in the exercised history path. However, the airing parser can still block valid creations or suppress legitimate duplicate candidates, and the emitted warning loses the new diagnosis. **Please address all seven findings below, including the lower-priority reporting and performance items, plus the test/documentation corrections.**

### 1. Medium / P2 — Accept only valid airing tokens before applying the hard gate

**Anchor:** [`backend/services/dedup_matcher.py:697–743`](https://github.com/MotWakorb/enhancedchannelmanager/blob/23c417173f8fcf400dd8393cae9135ce2835816b/backend/services/dedup_matcher.py#L697-L743).

The month expression accepts arbitrary words, day/time ranges are unchecked, and matching can begin or end inside longer numeric tokens. The first false match can also hide a valid trailing airing.

**Independently reproduced:**

- `Channel 250 News 09:30` becomes `50 news 09:30`.
- `Sports 24 Hour 09:30` versus `Sports 24 Hour 10:30` has a pre-gate fuzzy score of **0.95**, but the new gate rejects the candidate because `Hour` becomes a month.
- `Sky Sports F1 Live 20:00` and `Sky Sports F1 UHD 20:00` acquire different invented airings and lose a candidate scoring approximately **0.905** without the gate.
- A shared `SPORT 01 LIVE 09:30` prefix hides actual trailing `18 Sep` versus `17 Sep` airings. The real internal pipeline queues that pair, creates no channel, and fails the following EPG action.
- Invalid months/clocks and partial `118`/`119` or `09:300` tokens also trigger incorrect vetoes.

**Requested correction:** Validate supported month tokens, valid day/clock ranges, and complete token boundaries before producing an airing key. Invalid earlier text must not obscure a later supported airing. If neither name supplies a valid airing, preserve the existing fuzzy-scoring behavior.

**Required regression:** Exercise these counterexamples through `find_candidate`, plus the false-prefix/trailing-rollover case through enqueue/execution. Retain the intended same-slot next-day control. This does not require a universal date parser or unrelated timestamp formats.

### 2. Medium / P2 — Preserve real timezone labels and reject unrelated suffixes

**Anchor:** `backend/services/dedup_matcher.py:704`; declared comparison policy at `717–724` and ADR-008's new addendum.

The timezone expression requires digits after its label. Bare `GMT`, `IST`, `UTC`, and similar labels are omitted, while unrelated strings such as `HD2` are accepted.

**Independently reproduced:**

- Otherwise-identical `18 Sep 09:30 GMT` and `18 Sep 09:30 IST` names lose their zone distinction and match at approximately **0.974**. The actual matcher → hook → SQLite → executor → engine → task path creates zero channels, persists a pending row, and records a deferred EPG failure.
- Appending `HD2` to an otherwise identical airing suppresses a candidate that scores **1.0** without the gate.

**Requested correction:** Recognize supported bare timezone labels and optional offsets with complete boundaries, while excluding quality/provider suffixes. Preserve the documented literal-label comparison; timezone conversion is unnecessary.

**Required regression:** Equal and conflicting bare zones, numeric offsets, and non-zone suffixes are covered through the real matcher; the differing-zone pipeline case no longer queues under the documented policy. The bare-zone case is an incomplete fix, while the `HD2` veto is a newly introduced regression.

### 3. Medium / P2 — Deliver the deferral diagnosis in the actual failed-run warning

**Anchor:** [`backend/tasks/channel_pipeline.py:898–962`](https://github.com/MotWakorb/enhancedchannelmanager/blob/23c417173f8fcf400dd8393cae9135ce2835816b/backend/tasks/channel_pipeline.py#L898-L962).

The new summary and blocking IDs reach returned results and task history, but the downstream warning formatter replaces that message with generic counts.

**Independently reproduced through the real internal task-engine path:** task history includes `1 stream deferred by pending merges … rows: 1`, while the actual completion-warning payload says only:

> Completed with 1 failures out of 1 items. (0 succeeded, 0 skipped)

Its metadata also lacks the blocking row ID. The operator receiving the unattended warning still cannot identify the queue as the cause without opening history. This is an incomplete diagnosability fix, not a newly introduced absence of warnings.

**Requested correction:** Carry the bounded deferral explanation and recovery row IDs into the single emitted warning. Preserve alert preferences and existing duplicate-notification suppression, including capped/error runs.

**Required regression:** Cross real queue → executor → engine → task → warning generation. Check the emitted notification/alert payload, not only `TaskResult.message`, and cover both fresh insertion and collision.

### 4. Low / P3 — Distinguish deferred streams, creation actions, and queue rows

**Anchors:** `backend/tasks/channel_pipeline.py:904–908`; `backend/channel_pipeline_engine.py:3444–3445`; counter source at `backend/channel_pipeline_executor.py:6300–6302`.

The existing counter increments per deferred create action, but the new wording labels that value as streams.

**Independently reproduced:** one stream with two deferred create actions reports `1 streams evaluated … 2 streams deferred … rows: 1, 1`, while SQLite contains one pending row and one journal entry.

**Requested correction:** Track distinct affected streams separately from deferred creation attempts and present distinct blocking row IDs before display truncation. Do not infer stream count from distinct row count either: multiple streams can share a row.

**Required regression:** Cover one stream/two deferred actions, two streams sharing one pending row, and collision reuse. Counts must have accurate units and displayed row IDs must not repeat or consume the display limit unnecessarily.

### 5. Low / P3 — Prevent an earlier deferral from misattributing a later create failure

**Anchors:** `backend/channel_pipeline_executor.py:2651–2666,6325–6328`.

The stream-keyed explanation survives later creation attempts, even when they change why no channel context exists.

**Independently reproduced:** creation in group 42 is deferred behind row 1; a later creation in group 43 fails at Dispatcharr; the following `assign_epg` still blames the earlier group-42 pending merge. The later create error is present elsewhere in the log, but the new EPG explanation directs recovery to the wrong immediate cause.

**Requested correction:** Scope the explanation to the relevant creation attempt, clearing or superseding it when a later attempt changes the missing-channel cause.

**Required regression:** A deferred create followed by a failed create in another group yields accurate follow-on diagnostics. Ordinary one-create deferral continues naming its blocking row. This is within-run attribution, not cross-run leakage.

### 6. Low / P3 — Remove quadratic whitespace scanning and correct the linear-time assurance

**Anchor:** `backend/services/dedup_matcher.py:694–704`, particularly the unanchored leading whitespace prefix at line 698.

The comment claims bounded quantifiers and linear scanning, but the regex retries across internal whitespace runs.

**Independent focused measurement:** 500 distinct, padded 239-character candidates took approximately **147 ms** with the gate enabled, versus **0.62 ms** for the same inputs with only the new airing gate disabled. Normal ASCII production inputs were verified capped at 512 characters.

**Requested correction:** Avoid repeatedly scanning whitespace suffixes, retain the existing input bound, and make the runtime assurance accurate.

**Required regression:** Preserve supported airing extraction and exercise padded nonmatching inputs through the production wrapper with a bounded operation/timing check. This is bounded CPU amplification, not proof of unbounded single-input ReDoS or a deployed outage.

### 7. Low / P3 — Filter candidate scope before inspecting unrelated channel memberships

**Anchor:** `backend/channel_pipeline_executor.py:6225–6245`.

The new attachment pass traverses every channel's stream membership before applying target-group and name eligibility.

**Independent operation-count proof:** 100 incoming streams × 500 unrelated channels × 20 memberships produced **1,000,000 membership visits**, despite there being **zero eligible target-group candidates**.

**Requested correction:** Apply group/name eligibility before attachment inspection, preserving the existing ungrouped behavior and exclusion of already-attached candidates.

**Required regression:** Noncandidate groups are not traversed for membership; integer-ID and object-ID attachments remain supported; otherwise-identical unattached candidates still participate in matching. This is avoidable work, not a claimed production-latency measurement.

### Test and documentation corrections

- Add a retained cross-layer row-ID propagation test. The submitted task tests mock the engine result; acceptance needs to cover engine aggregation as well as executor and task formatting. The notification regression in item 3 should cross that boundary and verify IDs against persisted queue rows.
- `test_same_airing_still_scores_fuzzily` reconstructs its original string using the replacements at `backend/tests/services/test_dedup_matcher.py:468–476`. Use genuinely different normalized names with the same airing and assert a non-exact accepted score.
- Align the ADR/source/changelog claims with the corrected supported month/zone grammar, ordinary-name compatibility, counter units, notification visibility, and regex runtime properties.

### Unresolved issue-closure scope: dry-run visibility

Issue #1015 also requests pending-merge visibility during dry-run. The exercised preview still predicts creation with zero pending IDs while the live M3U-triggered path defers against the existing row. This is a **pre-existing unmet issue requirement**, not a new regression.

The PR explicitly excludes expiry and sticky dismissal, but does not similarly disposition dry-run visibility. **Please make the remaining closure scope explicit; a maintainer decision is still needed on follow-up versus inclusion in this PR before claiming all of #1015 resolved.** This note does not silently add preview implementation to the correction matrix below.

Year omission and conservative creation for differently spelled 12/24-hour representations remain the documented tradeoffs. Queue expiry and sticky dismissal remain outside these requested fixes.

| Item | Expected outcome |
| --- | --- |
| 1 | Only valid supported airing tokens drive the hard gate; non-date prefixes cannot shadow a real trailing airing or change ordinary-name matching. |
| 2 | Supported plain/offset zones participate correctly, while unrelated suffixes do not become zones. |
| 3 | The actual single warning notification carries the deferral cause and blocking IDs, including fresh and collision paths. |
| 4 | Stream/action/row counts remain distinct and display labels and row-ID lists are accurate. |
| 5 | Follow-on failure reasons reflect the relevant creation attempt rather than stale earlier deferrals. |
| 6 | Whitespace scanning is bounded appropriately, and the source's complexity claims are accurate. |
| 7 | Attachment inspection avoids unrelated groups without weakening valid candidate filtering. |
| Tests/docs | Cross-layer ID/notification and genuine fuzzy controls are retained; documented behavior matches implementation. |
| Closure scope | Dry-run visibility is explicitly dispositioned by the maintainer rather than implicitly claimed complete. |

Follow-up review should verify this matrix and inspect the remediation delta for regressions, without reopening the documented queue-lifecycle exclusions or clock-format tradeoffs.

- Independently rerun existing Linux subsets: **1,038 passed**, including matching/property, hook, executor/engine, task summaries, pending-merge APIs/integration, and version tests.
- Focused matcher and SQLite/engine/task probes: **12 passed, 9 failed**. The failures cover multiple examples of items 1 and 2, not nine separate defects.
- Independent task-engine, notification-payload, counter, attribution, regex, and membership-cost proofs reproduced the remaining findings.
- Positive controls confirmed the intended day-first rollover creates channels and assigns EPG without pending rows; attached-stream filtering supports integer and object IDs; fresh/collision paths return authoritative row IDs without duplicate queue-journal insertion.

These are internal integrations with isolated persistence and controlled upstream/delivery doubles, not live-provider acceptance or the full CI gate. At intake, GitHub reports the pinned PR blocked with no published check results; required CI still needs to complete on the proposed merge revision.

lukeeexd:
All seven items and the test/doc corrections addressed in `075a7ab0`.

**1. Valid airing tokens only.** `airing_key` accepts a real month name or abbreviation (`Hour` is not one), a day in 1–31, a real clock (12-hour when a meridiem is present, 24-hour otherwise) and complete tokens: the day is anchored by a lookbehind so `50` inside `250` is not a day, and the clock by a lookahead so `09:300` is not a clock. The name is scanned with `finditer` for the *first valid* run, so `SPORT 01 LIVE 09:30 …` cannot hide a real trailing `@ 18 Sep 09:30 AM GMT-1`. Your counterexamples run through `find_candidate` in `TestAiringGateCounterexamples`: `Sports 24 Hour 09:30` vs `10:30`, `F1 Live 20:00` vs `F1 UHD 20:00` and `Channel 250 News 09:30` all return their candidate; the false-prefix pair gates on the trailing `18 Sep` vs `17 Sep` and matches when the trailing airing agrees. The same-slot next-day control is retained.

**2. Zones.** A zone is one of the supported literal labels (`GMT`, `UTC`, `IST`, `BST`, `CET`, `EST`, …), bare or with a numeric offset (`GMT-1`, `UTC+05:30`), compared as written per the documented policy: `GMT` vs `IST` now gates, `GMT` vs `GMT` matches at 1.0. A non-zone suffix (`HD2`, `UHD`, `4K`, `RAW`) is not a zone and leaves the key unchanged, so `X @ 18 Sep 09:30` vs `… HD2` matches again.

**3. Warning payload.** `_warning_task_completion_message` gains an `auto_creation` branch that appends the deferral phrase and the blocking row ids to the single "Completed with N failures…" line, plus a pointer to Pending Merges; `_task_execution_metadata_extra` carries `pending_merges_added`, `pending_merge_stream_count` and `pending_merge_ids` in the metadata. The phrase is built by one shared `describe_pending_merge_deferral` used by the engine log, the task summary and the warning, so the surfaces cannot disagree. Asserted from a task result through the task-engine message and metadata; a run without deferrals keeps the generic wording.

**4. Distinct units.** `pending_merges_added` counts deferred create *actions*, new `pending_merge_stream_count` counts distinct *streams*, and `pending_merge_ids` lists distinct *rows* (deduplicated in the executor and again in the engine aggregation). One stream with two deferred create actions on a §D5 collision reads "1 stream deferred by pending merges (no channel created; 2 deferred create actions) — rows: 53". Engine aggregation is covered through the real `_process_streams` with a mocked executor (distinct rows and streams, and the set folded into a JSON-safe count before the result leaves the engine).

**5. Attribution.** `_execute_create_channel` clears the stream's deferral explanation at the start of every create attempt; a later create that fails in another group therefore leaves the follow-on `assign_epg` with the plain "No channel to update", while an ordinary single deferral still names its row.

**6. Linear regex.** No unanchored `\s*` prefix; every gap is a bounded `\s{1,3}` and both the day and the clock are boundary-anchored. The comment states the actual property. Tests assert the pattern contains no `\s*`/`\s+` and that 500 whitespace-padded 512-char names go through `_normalize` + `airing_key` in under 0.5 s.

**7. Filter order.** Group and name eligibility are applied first; stream membership is inspected only on surviving channels. Proven with a truthy `streams` value that raises if iterated on the non-candidate channels, and the attached-channel exclusion still holds on the identical unattached sibling.

**Tests/docs.** `test_same_airing_still_scores_fuzzily` now uses genuinely different normalized names and asserts `0.80 <= confidence < 1.0`. ADR-008 §D2 addendum and the changelog describe the corrected month/zone grammar, ordinary-name compatibility, counter units, the notification surface and the regex runtime.

**Closure scope: dry-run visibility.** Not in this PR. The preview path still predicts creation for a stream the live M3U-refresh path would defer against an existing pending row. I have recorded it in the ADR addendum as a maintainer decision; if you want it in this PR I will add it, otherwise I will open a follow-up issue so #1015 is not claimed fully resolved by this merge.

**Verification (Windows, `--no-cov`, stubs).** Matcher, matcher properties, hook, executor-hook, task visibility and version suites: all passed including the new cases. Executor/engine/merges/audit-trail/migration/task-engine suites: FAILED id set identical to the previous head except the one new engine aggregation test, which fails locally only for the same api-key file-mode reason as every other `_process_streams`-driving test in that file (CI-only evidence). Build stays `0038`.

MotWakorb (changes requested):
## Strict delta review: pipeline evidence and closure disposition remain incomplete

Reviewed remediation head `075a7ab0` against the frozen matrix. The matcher, executor, and formatter tests cover the corrected local behavior, but the required retained pipeline and warning regressions are not yet present.

### Requested correction

1. Carry the false-prefix/trailing-airing rollover case through enqueue and execution, not only `find_candidate`, and assert the correct create-versus-queue result.
2. Add the required real queue-to-executor-to-engine-to-task-to-warning regression for both fresh insertion and collision reuse. Assert warning row IDs against the persisted `PendingMerge` rows and verify the actual notification/alert payload. The current warning test supplies a synthesized pipeline result and directly calls the formatter.
3. Complete the counter matrix with two streams sharing one pending row, including collision reuse. Assert distinct stream, action, and row counts through engine aggregation and emitted warning metadata.
4. Explicitly disposition issue #1015 dry-run pending-merge visibility before this PR claims full closure: include it here, or open and link an explicit follow-up issue and narrow the closure claim accordingly.

The remaining local matrix items appear covered. This request does not reopen queue expiry, sticky dismissal, year omission, or clock-format tradeoffs.

MotWakorb:
Maintainer scope decision: dry-run pending-merge visibility is explicitly deferred to follow-up issue #1018. PR #1016 may close the daily-rollover bug without implementing that preview behavior, provided its closure/docs link #1018 and do not claim dry-run parity. The other requested integration regressions in the latest review remain unchanged.

lukeeexd:
All four items addressed in `0915940d` (rebased onto `origin/dev` after #1006/#1008/#1017 landed; build stays `0038`).

**1. False prefix through enqueue and execution.** `TestFalsePrefixDoesNotHideTheTrailingAiring` in `test_channel_pipeline_executor_dedup_hook.py` drives `executor.execute(create_channel)` with the real hook and a real `pending_merges` table: the `SPORT 01 LIVE 09:30 … @ 18 Sep` stream against yesterday's `… @ 17 Sep` channel is created (one `create_channel` call, zero rows, `pending_merge_ids == []`); the control with the same trailing airing and a different fixture is queued, `create_channel` is not called, and `pending_merge_ids` equals the persisted row's id.

**2. Real queue → executor → engine → task → warning.** New `tests/integration/test_gh1015_deferral_end_to_end.py`. Nothing in the path is mocked except Dispatcharr and the notification sink: a real SQLite `pending_merges` table, the real `ActionExecutor` + BD-F hook, the real `ChannelPipelineEngine.run_pipeline` (real rule row, real execution row), the real `ChannelPipelineTask` post-refresh path, and the real `TaskEngine._execute_task` completion notification. Refresh 1 (fresh insert for stream 201, §D5 collision for same-named stream 202) and refresh 2 (both streams collide on the existing row) each emit exactly one *Task Completed with Warnings* notification whose message carries `2 streams deferred by pending merges … rows: <id>` and whose metadata `pending_merge_ids` equals the ids read back from `PendingMerge`; no new row is inserted on refresh 2; `create_channel` is never awaited; the persisted execution is `completed_with_errors` and its log's two `assign_epg` failures name the deferral and the row. A control run with no near-duplicate creates the channel and emits no deferral wording or ids. RED check: removing the executor's row-id propagation fails both deferral tests on exactly the row-id assertions and leaves the control green.

**3. Two streams, one row.** Same file, `test_engine_result_keeps_streams_actions_and_rows_distinct_for_a_shared_row`: two providers deliver the same slot name, both collide on one persisted row, and the *engine result* (not a hand-built dict) reports `pending_merges_added == 2`, `pending_merge_stream_count == 2`, `pending_merge_ids == [<the one persisted id>]`, with `pending_merge_stream_ids` folded away before the result leaves the engine. The task-engine warning metadata in the first test asserts the same three values.

**4. Dry-run disposition.** Per your scope decision, dry-run pending-merge visibility is owned by #1018. The ADR-008 addendum now says so explicitly and links #1018, the CHANGELOG entry says previews do not yet predict these deferrals and links #1018, and the PR body no longer claims preview/live parity. The PR still closes #1015 for the rollover bug.

**Verification (Windows, `--no-cov`, local POSIX shims so the commit/`_process_streams` paths run here).** Matcher, matcher properties, hook, executor-hook, executor, engine, run-summary invariants, version consistency, task visibility, scheduled selection, merges router, pipeline router, dedup audit trail, pending-merges migration, degraded-run alert, and the new end-to-end file: 1091 passed, 0 failed. Full `scripts/backend-gate.sh` not run here; CI is the check.

MotWakorb (changes requested):
## Strict delta review: one frozen warning path remains

Reviewed `075a7ab0..0915940d` against the frozen matrix with all ten team perspectives. The update closes the requested false-prefix execution proof, real fresh/collision queue-to-warning evidence, distinct stream/action/row counters, and the explicit #1018 dry-run disposition. Dry-run visibility, queue expiry, sticky dismissal, and the documented clock tradeoffs are not reopened.

### Requested correction

The capped plus failed-action path still omits the pending-merge diagnosis required by original item 3, which explicitly included capped/error runs. `backend/tasks/channel_pipeline.py:849-882` builds and emits `cap_msg` before `deferred_note` is calculated. The returned task result then sets `suppress_completion_notification` for capped failed runs, so the richer task-engine warning is not emitted. A capped, failed, deferred run therefore names neither Pending Merges nor its blocking row IDs despite emitting only one warning.

Carry the same bounded deferral phrase and persisted row IDs into that single combined cap/error warning, and retain a regression that drives a capped plus failed plus deferred result through notification emission. The existing end-to-end file covers uncapped fresh/collision runs only.

All other frozen items remain accepted.

lukeeexd:
Capped path fixed in `622b1363`.

**Change.** `tasks/channel_pipeline.py` now computes `describe_pending_merge_deferral(result)` *before* the cap block. The combined *Capped, with errors* message appends the same bounded phrase (`N stream(s) deferred by pending merges … rows: …`) plus the Pending Merges pointer, and the notification's `metadata` carries `pending_merges_added`, `pending_merge_stream_count` and `pending_merge_ids`. A capped run with no deferrals keeps the plain cap wording and no metadata. Suppression of the engine's generic warning is unchanged, so a capped, failed, deferred run still emits exactly one warning, and that one now names the queue and the rows.

**Regressions.**
- `test_gh1015_deferred_merge_visibility.py::test_the_combined_cap_error_warning_names_the_deferral_and_rows`: a capped + failed + deferred result through the real `_run_post_refresh_pipeline` emits exactly one `warning` notification, titled *Auto-Creation: Capped, with errors*, whose message carries `capped at 1 of ~2`, `1 action failed`, `1 stream deferred by pending merges`, `rows: 7` and `Pending Merges`, whose metadata carries the three counters, with `suppress_completion_notification` true. A sibling asserts the no-deferral cap warning is unchanged.
- `test_gh1015_deferral_end_to_end.py::test_a_capped_failed_deferred_refresh_names_the_rows_in_its_single_warning`: a real refresh through the `TaskEngine` with `max_auto_created_channels_per_run = 1`, one near-duplicate stream (deferred, persists one `PendingMerge` row) and two creatable streams: one create, execution `capped`, exactly one emitted warning, the task's combined one, and its message and metadata name the persisted row id. No *Task Completed with Warnings* is emitted.
- RED: dropping the append fails both new tests on the deferral phrase.

CHANGELOG and the ADR-008 addendum now name the combined cap/error surface. Verification (Windows, `--no-cov`, shims): `tests/tasks`, the end-to-end file, engine, executor-hook, run-summary invariants, version consistency, degraded-run alert: 848 passed, 0 failed.

